PR #2, opened 2019-07-30
Setup viewport renderer and shader compilation

yanchith:
Contains the following:

1) Add a build.rs file that watches and compiles shaders in
`src/shaders` and adds the compiled versions to the target
directory. The build script does *not* recurse.

2) Create a very simple (and possibly leaky) abstraction around `wgpu`
for our viewport renderer. This will improve at some later point.

3) Handle window resize events where the renderer is concerned:
    - recreate the swap chain
    - recreate the depth texture
    - recompute the projection matrix


yanchith:
Oh, also as far as rendering backends are concerned:

I tested d3d12, metal and vulkan and they seemed to work on my machines. d3d11 didn't work (just drew garbage triangles) and opengl doesn't compile in our project yet (there is a bit of setup involved on our side to make it compile).

src/main.rs
```diff
@@ -1,21 +1,177 @@
...
     let mut event_loop = winit::EventsLoop::new();
-    let _window = winit::Window::new(&event_loop).expect("Failed to create window.");
-    let mut running = true;
+    let window = winit::Window::new(&event_loop).expect("Failed to create window.");
+    let window_size = window
+        .get_inner_size()
+        .expect("Failed to get window inner size")
+        .to_physical(window.get_hidpi_factor());
+
+    let instance = wgpu::Instance::new();
```
ondrowan: Let's rename this to `wgpu_instance`.

src/viewport_renderer.rs
```diff
@@ -0,0 +1,402 @@
...
+pub type Index = u32;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct GeometryId(u64);
+
+#[derive(Debug)]
+pub enum ViewportRendererAddGeometryError {
+    TooManyVertices(u32, usize),
+    TooManyIndices(u32, usize),
+}
```
ondrowan, line 30: Is the first `u32` needed since it's always its max value? Or is it possible for it to be different value in the future?

src/viewport_renderer.rs
```diff
@@ -0,0 +1,402 @@
...
+
+        // Vulkan (and therefore wgpu) has different NDC and
+        // clip-space semantics than OpenGL: Vulkan is right-handed, Y
+        // grows downwards. The easiest way to keep everything working
+        // as before and use all the libraries that assume OpenGL is
+        // to apply a correction to the projection matrix which
+        // normally changes the right-handed OpenGL world-space to
+        // left-handed OpenGL clip-space (apart from actually
+        // performing the projection).
+        // https://matthewwellings.com/blog/the-new-vulkan-coordinate-system/
```
ondrowan, line 356: Wouldn't this be better as docstring for this function?

src/viewport_renderer.rs
```diff
@@ -0,0 +1,402 @@
...
+            render_pass.draw_grid(...);
+            render_pass.draw_geometry(...);
+            render_pass.draw_geometry_wireframe(...);
+            render_pass.draw_geometry_highlight(...);
+            render_pass_draw_ui(...);
+
+            render_pass.finish()?; // Consumes. Drop panicks unless this is explicitely called.
+        }
+
+        */
```
ondrowan: I'd move this to a task since it's something that probably needs to be done eventually anyway. You could also explain (preferably in some docstring) what's the deal with passing device reference everywhere. Does it have something to do with lifetimes?

src/viewport_renderer.rs
```diff
@@ -0,0 +1,402 @@
...
+        );
+
+        self.geometries_next_id += 1;
+        Ok(id)
+    }
+
+    pub fn remove_geometry(&mut self, id: GeometryId) {
+        log::debug!("Removing geometry with {}", id.0);
+        // Dropping the geometry descriptor here unstreams the buffers from device memory
+        self.geometries.remove(&id.0);
```
ondrowan, line 285: Wouldn't it be a good idea to also implement `Drop` which would remove all geometry?

src/viewport_renderer.rs
```diff
@@ -0,0 +1,402 @@
...
+            format: DEPTH_FORMAT,
+            usage: wgpu::TextureUsage::OUTPUT_ATTACHMENT,
+        })
+    }
+}
+
+struct GeometryDescriptor {
+    vertices: (wgpu::Buffer, u32),
+    indices: Option<(wgpu::Buffer, u32)>,
+}
```
ondrowan, line 407: It'd be probably better if this was on top of this file amongst other structs.

ondrowan (changes requested):

src/viewport_renderer.rs
yanchith: Currently always max value. I'll remove it. There may be another limit, but we can add that after we discover it.

src/viewport_renderer.rs
yanchith: It's private and always will be, though. I can do it if you insist, but I usually try to put most public things towards the top of the file (except maybe consts)

src/viewport_renderer.rs
yanchith: I think that happens automatically when the `HashMap` is dropped. (Which is when this struct is dropped)

src/viewport_renderer.rs
yanchith: I'll keep this comment here, but also add a docstring which explains what the method does... while at it, I'll also docstring some other pub functions.

src/viewport_renderer.rs
yanchith: Ok, I'll create a task for it. Yes, `ViewportRenderer` can't own the device, unfortunately, and neither can it store a permanent shared or mutable reference to it, as there will be one other renderer (`GuiRenderer`) that needs it.

yanchith:
@ondrowan I think I addressed all

ondrowan (approved):

PR #4, opened 2019-08-01
Loading of wavefront obj files into scene

ondrowan:
This is just a draft PR. The following still needs to be done:

* Docstrings.
* Refactoring - I especially don't like confusing names of some variables.
* Finish "libify" of the application. I've created a very basic `lib.rs` just to test `Scene`. Cargo is currently thoroughly confused and thinks no functions are used. I'll need to look into it.
* Absolute file paths are used as file picker provides them. This needs to be changed into relative paths.
* I'm not sure if it's okay to pass path around as `String` or it's gonna be easier as `&str`. There's a lot of cloning going on, but I'm not sure if that can be helped.


ondrowan:
Creating a lib meant just removing `mod`s from `main.rs`, moving them to `lib.rs` and making them `pub`. I'm not sure if I like it, but everything works.

yanchith:
This isn't a review yet, just high level thoughts:

**Overall Organization**

I am not 100% I am saying the most correct things in this section, so let's discuss...

The functionality is split in 3 files, `obj.rs`, `file.rs`, and and `scene.rs`. I wouldn't be against all 3 of these just being in a single file called `loader.rs` or `importer.rs` (the name isn't really _that_ important) containing a single struct `Importer`, which would have all the `HashMap`s `Scene` currently has.

The reason for this is that it may be easier to split into modules later once we know it needs to be, but at the moment the functionality seems closely interconnected.

In integration tests I would `use hurban_selector::importer::Importer` and test that. It might be useful to think what API `Importer` would have:

```rust
let mut importer = Importer::new();

// either returns freshly parsed or cached geometry, depending on whether it changed on FS
let geometry = match importer.import_obj_geometry("foo") {
    Ok(geometry) => geometry,
    Err(ImporterError::FileDoesNotExist) => {
        // notify user? propagate error? Can be tested against
    },
    Err(ImporterError::InvalidGeometry) => {
        // notify user? propagate error? Can be tested against
    }
    // What other error cases are there?
}

// We also might want this in the future so that we don't have an ever-growing cache...
importer.forget("foo");
// ...Or the importer could handle that internally to not expose the complexity, e.g. auto-evict models from the cache
```

**Integration Tests**

Nice! We should do more of these for other systems 😅 

**lib.rs**

To make sure we can cover more things by integration tests, I would prefer to make `main.rs` as thin as possible. Maybe even just something similar to this:

```rust
use hurban_selector::{App, AppParams};

fn main() {
    let params = parse_app_params();
    let app = App::new(params);
    app::run(); // does not return
}

fn parse_app_params() -> AppParams {
    // parse the thing, use clap or something
}
```

**Misc**

- You managed to remove my `Q` keyboard shortcut 😢 
- Use `&str` instead of `&String`. `&str` is more generic and easier to provide than a `&String`, therefore more flexible.
- Shouldn't `calculate_checksum()` receive bytes as the parameter instead of the path? Or if we want to be fancy: `std::io::Read`?
- Currently it seems that the file is opened and read twice. Once for the checksum, once for the obj parser. I would suggest to preload it to a buffer and give that both to `calculate_checksum` and `load_obj` (or `parse_obj`?). `tobj` can eat pretty much anything (it just requires `BufRead`) if you use the more generic method: https://docs.rs/tobj/0.1.8/tobj/fn.load_obj_buf.html
- You use both `test_` and `it_` conventions for test names. I personally prefer `test_add_numbers` to `it_should_add_numbers_correctly`, but feel free to pick either.

yanchith:
Also... It might be good to start thinking about the semantics of import being an operation vs a standalone thing and the ramifications that will have... So far I think it will affect whether the all of the system, or just the importer will want to know about checksums. 

**Import is an operation**
Looks like import as an operation doesn't really care about checksums. We don't store them anywhere in the scene file and only store the operation definitions. The importer will compute and compare checksums, but only as an optimization to not parse the same file twice.

**Import is standalone**
If we present import as something standalone, suddenly the user might expect the scene to look the same between two runs of the editor, and would want to be notified if any imported assets changed. To have that functionality, the checksum needs to be an integral part of the model.



 the integrity of our scene file depends on whether any referenced assets changed. I think , so if any linked assets changed, we should probably notify the user.

ondrowan:
> **Overall Organization**
> 
> I am not 100% I am saying the most correct things in this section, so let's discuss...
> 
> The functionality is split in 3 files, `obj.rs`, `file.rs`, and and `scene.rs`. I wouldn't be against all 3 of these just being in a single file called `loader.rs` or `importer.rs` (the name isn't really _that_ important) containing a single struct `Importer`, which would have all the `HashMap`s `Scene` currently has.
> 
> The reason for this is that it may be easier to split into modules later once we know it needs to be, but at the moment the functionality seems closely interconnected.
> 
> In integration tests I would `use hurban_selector::importer::Importer` and test that. It might be useful to think what API `Importer` would have:
> 
> ```rust
> let mut importer = Importer::new();
> 
> // either returns freshly parsed or cached geometry, depending on whether it changed on FS
> let geometry = match importer.import_obj_geometry("foo") {
>     Ok(geometry) => geometry,
>     Err(ImporterError::FileDoesNotExist) => {
>         // notify user? propagate error? Can be tested against
>     },
>     Err(ImporterError::InvalidGeometry) => {
>         // notify user? propagate error? Can be tested against
>     }
>     // What other error cases are there?
> }
> 
> // We also might want this in the future so that we don't have an ever-growing cache...
> importer.forget("foo");
> // ...Or the importer could handle that internally to not expose the complexity, e.g. auto-evict models from the cache
> ```

This is just how I usualy do things. I split everything and merge it later if it makes sense. E.g. `file.rs` was initially bigger and functionality got later moved elsewhere, so I guess it'd make sense to merge it with `obj.rs` at least. It's basically just wrappers around library functions anyway.

Changing the name and semantics into `Importer` sounds good in case `Scene` will be built and used in a different way. Also good point that `FileDoesNotExist` will be needed in the end since this function is gonna be used to load *.obj files from our project file and those paths could be invalid. I expected the input coming from file pickers only.

> **lib.rs**
> 
> To make sure we can cover more things by integration tests, I would prefer to make `main.rs` as thin as possible. Maybe even just something similar to this:
> 
> ```rust
> use hurban_selector::{App, AppParams};
> 
> fn main() {
>     let params = parse_app_params();
>     let app = App::new(params);
>     app::run(); // does not return
> }
> 
> fn parse_app_params() -> AppParams {
>     // parse the thing, use clap or something
> }
> ```

Agreed. It's the same as we usualy do with web apps. We don't necessarily need to do this PR though.

> **Misc**
> 
>     * You managed to remove my `Q` keyboard shortcut 😢

Ah, the merge conflict was a bit messy and I missed another shortcut being added.

>     * Use `&str` instead of `&String`. `&str` is more generic and easier to provide than a `&String`, therefore more flexible.

The thing is, I tried both scenarios and code-wise it was kinda the same. There's a lot of string cloning going on as hash maps use owned strings as keys and values. And since file picker returns `String`, I chose `String` to be parameter to `Scene` functions. I understand it's more flexible, but I'm not sure if we need any flexibility, so the important question is, is it more idiomatic to use `&str` in this case anyway?

>     * Shouldn't `calculate_checksum()` receive bytes as the parameter instead of the path? Or if we want to be fancy: `std::io::Read`?
> 
>     * Currently it seems that the file is opened and read twice. Once for the checksum, once for the obj parser. I would suggest to preload it to a buffer and give that both to `calculate_checksum` and `load_obj` (or `parse_obj`?). `tobj` can eat pretty much anything (it just requires `BufRead`) if you use the more generic method: https://docs.rs/tobj/0.1.8/tobj/fn.load_obj_buf.html

I knew about `load_obj_buf` but misunderstood its usage. Will change it so we load files just once.

>     * You use both `test_` and `it_` conventions for test names. I personally prefer `test_add_numbers` to `it_should_add_numbers_correctly`, but feel free to pick either.

`test` it is then.

> Also... It might be good to start thinking about the semantics of import being an operation vs a standalone thing and the ramifications that will have... So far I think it will affect whether the all of the system, or just the importer will want to know about checksums.
> 
> **Import is an operation**
> Looks like import as an operation doesn't really care about checksums. We don't store them anywhere in the scene file and only store the operation definitions. The importer will compute and compare checksums, but only as an optimization to not parse the same file twice.
> 
> **Import is standalone**
> If we present import as something standalone, suddenly the user might expect the scene to look the same between two runs of the editor, and would want to be notified if any imported assets changed. To have that functionality, the checksum needs to be an integral part of the model.
> 
> the integrity of our scene file depends on whether any referenced assets changed. I think , so if any linked assets changed, we should probably notify the user.

As discussed IRL, we'll count on import being an operation and for now checksums are gonna be relevant just to importer.

ondrowan:
@yanchith This is ready for re-review. It was changed with a more pragmatic approach: Everything is inside of `importer` module. Once we need some of the functions somewhere else, we'll extract them. More docs and tests were added as well. As I mentioned before, two tests are missing as I don't know how to write them without mocking heavily or being way too clever:

* accessing file for which I don't have permissions
* importing already cached file (path) with changed contents

On a side note, our Travis seems to be dying. At least in this branch it doesn't pass Win or MacOS builds, so please try to run fmt, clippy and tests locally. We'll see if it fixes itself once we merge this to master.

src/main.rs
```diff
@@ -121,6 +110,52 @@ fn main() {
...
+
+                                    tinyfiledialogs::message_box_ok(
+                                        "Error",
+                                        error_message,
+                                        tinyfiledialogs::MessageBoxIcon::Error,
+                                    );
+                                };
+                            }
+                        }
+                        winit::VirtualKeyCode::S => {
```
yanchith, line 137: Do we need this dummy code right now?

src/main.rs
```diff
@@ -121,6 +110,52 @@ fn main() {
...
+                        ..
+                    } => match code {
+                        winit::VirtualKeyCode::O => {
+                            if let Some(path) = tinyfiledialogs::open_file_dialog(
+                                "Open",
+                                "",
+                                Some((&["*.obj"], "Wavefront (.obj)")),
+                            ) {
+                                if let Err(err) = importer.import_obj(&path) {
+                                    let error_message = match err {
```
yanchith: Might be better to `impl fmt::Display for ImporterError` and use it here :)

src/importer.rs
```diff
@@ -0,0 +1,284 @@
...
+/// Converts `tobj::Model` vector into vector of internal `Model` representations.
+/// It expects valid `tobj::Model` representation, eg. number of positions
+/// divisible by 3.
+pub fn tobj_to_internal(tobj_models: Vec<tobj::Model>) -> Vec<Model> {
+    let mut models = Vec::with_capacity(tobj_models.len());
+
+    for model in tobj_models {
+        let mut vertices = Vec::with_capacity(model.mesh.positions.len() / 3);
+
+        for (index, _) in model.mesh.positions.iter().enumerate().step_by(3) {
```
yanchith: Have a look at `chunks` or `chunks_exact`

https://doc.rust-lang.org/std/primitive.slice.html#method.chunks
https://doc.rust-lang.org/std/primitive.slice.html#method.chunks_exact

Your approach is not bad, but `chunks` seems a bit simpler.

I was also gonna say `chunks` avoids runtime bound checks, but that's not actually true, because we are still indexing the slice given to us by `chunks`...

src/importer.rs
```diff
@@ -0,0 +1,284 @@
...
+use std::fs;
+use std::io::ErrorKind;
+
+use crc32fast;
+use tobj;
+
+use crate::viewport_renderer::{Index, Vertex};
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum ImporterError {
```
yanchith, line 15: `fmt::Display` and `error::Error` impls would be nice. Also a `From` impl from `io::Error`.

src/importer.rs
```diff
@@ -0,0 +1,284 @@
...
+    pub vertices: Vec<Vertex>,
+    pub indices: Vec<Index>,
+}
+
+/// `Importer` takes care of importing of obj files and caching of their
+/// internal representations. It holds paths to files, their checksums and
+/// parsed obj files.
+#[derive(Debug, Default)]
+pub struct Importer {
+    pub path_checksums: HashMap<String, u32>,
```
yanchith: These `pub`s make me sad. No tests are worth this 😿 You could make these into borrowing getters, if you really need to assert things against internal state here, e.g.:

`pub fn path_checksums(&self) -> &HashMap<String, u32> { &self.path_checksums }`

But I would prefer not to expose these details altogether. Most of the tests would get by with having a single `Importer::has_cached(&self, &str)` method.

src/importer.rs
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    /// Tries to import obj file from given `path`. If file was already
+    /// imported and its checksum is identical, parsed models are returned
+    /// from cache. If the checksum doesn't match, file is reimported and
+    /// the old one is forgotten. If file wasn't imported, but its checksum
+    /// matches other file, it is cloned. Brand new files are parsed and
+    /// cached.
+    pub fn import_obj(&mut self, path: &str) -> Result<Vec<Model>, ImporterError> {
+        let file_contents = self.load_file_to_string(path)?;
```
yanchith: Hmmm, are OBJ files always valid utf8? We could theoretically operate on `Vec<u8>` if `tobj` supports it.

src/importer.rs
```diff
@@ -0,0 +1,284 @@
...
+            },
+        }
+    }
+
+    fn duplicate_models(&mut self, checksum: u32, path: &str) {
+        let duplicate_paths = self
+            .checksum_paths
+            .get_mut(&checksum)
+            .expect("Checksum expected to be present in cache");
+        let duplicate_path = duplicate_paths[0].clone();
```
yanchith: ~I think you can do without this `clone`.~ See my comment about the `Importer` fields instead.

src/importer.rs
yanchith: Hmmmmm, aren't just 2 fields sufficient?

```rust
checksum_models: HashMap<u32, Vec<Model>>,
path_checksums: HashMap<String, u32>,
```

"Duplicating" would just be adding a entry to `path_checksums`, no need to clone the actual model. This is of course assuming that two checksums for different models aren't the same, but you are already assuming that.

Some logic would change under  this model, e.g. if a file changes, you add the newly parsed model under a new entry to `checksum_models` and rewire the entry in `path_checksums` to it. Once no `path_checksums` point to an entry in `checksum_models`, you can garbage collect it.

I am not sure if this is overall a better approach, but it doesn't duplicate/clone data nearly as much, so it should be more 🍀

src/importer.rs
```diff
@@ -0,0 +1,284 @@
...
+        Default::default()
+    }
+
+    /// Tries to import obj file from given `path`. If file was already
+    /// imported and its checksum is identical, parsed models are returned
+    /// from cache. If the checksum doesn't match, file is reimported and
+    /// the old one is forgotten. If file wasn't imported, but its checksum
+    /// matches other file, it is cloned. Brand new files are parsed and
+    /// cached.
+    pub fn import_obj(&mut self, path: &str) -> Result<Vec<Model>, ImporterError> {
```
yanchith, line 83: This could return a `Cow<Vec<Model>>` in the Ok variant if you want to avoid further clones, but let's not go there just yet 😉

src/importer.rs
```diff
@@ -0,0 +1,284 @@
...
+
+        Ok(models)
+    }
+
+    /// Loads contents of file to string. In case of any error, it is converted
+    /// into `ImporterError`.
+    fn load_file_to_string(&self, path: &str) -> Result<String, ImporterError> {
+        match fs::read_to_string(path) {
+            Ok(contents) => Ok(contents),
+            Err(err) => match err.kind() {
```
yanchith: This can be a `impl From<io::Error> for ImporterError` and this whole function just disappears. `?` operator will do the conversion for you once you have `From`.

tests/importer.rs
```diff
@@ -0,0 +1,252 @@
+use std::collections::{HashMap, HashSet};
+use std::fs;
+
+use hurban_selector::importer;
+use hurban_selector::importer::{Importer, ImporterError};
```
yanchith: Can be written as `use hurban_selector::importer::{self, Importer, ImporterError};`

tests/importer.rs
```diff
@@ -0,0 +1,252 @@
...
+fn import_obj(path: &str) -> Vec<importer::Model> {
+    let file_contents = fs::read_to_string(&path).expect("File should be read to string");
+    let (tobj_models, _) =
+        importer::obj_buf_into_tobj(&file_contents).expect("Obj should be parsed");
+
+    importer::tobj_to_internal(tobj_models)
+}
+
+#[test]
+fn test_caches_valid_obj_file() {
```
yanchith: How about having a `has_cached(&str)` on `Importer`? You wouldn't need to have all those items public and would still be able to test exactly what you want to:

```
create importer;
use importer to load model;
ensure model is cached;
```

tests/importer.rs
```diff
@@ -0,0 +1,252 @@
...
+    let mut expected_loaded_models_paths: HashSet<&String> = HashSet::new();
+    expected_loaded_models_paths.insert(&path);
+
+    assert_eq!(importer.path_checksums, expected_path_checksums);
+    assert_eq!(importer.checksum_paths, expected_checksum_paths);
+    assert_eq!(loaded_models_paths, expected_loaded_models_paths);
+}
+
+#[test]
+fn test_returns_valid_obj_file() {
```
yanchith: This seems to me like it's testing `1 == 1` scenario. Does make sense to have tests like these?

tests/importer.rs
```diff
@@ -0,0 +1,252 @@
...
+    let models = importer
+        .import_obj(&path)
+        .expect("Valid object should be loaded");
+    let expected_models = import_obj(&path);
+
+    assert_eq!(expected_models, models);
+}
+
+#[test]
+fn test_caches_two_different_valid_obj_files() {
```
yanchith: See my earlier comment about `has_cached(&str)`. Should apply here as well.

tests/importer.rs
```diff
@@ -0,0 +1,252 @@
...
+        .expect("Valid object should be loaded");
+    let expected_models_1 = import_obj(&path_1);
+    let expected_models_2 = import_obj(&path_2);
+
+    assert_eq!(expected_models_1, models_1);
+    assert_eq!(expected_models_2, models_2);
+}
+
+#[test]
+fn test_does_not_cache_invalid_obj_file() {
```
yanchith: See my earlier comment about `has_cached(&str)`. Should apply here as well.

tests/importer.rs
```diff
@@ -0,0 +1,252 @@
...
+
+    let error = importer
+        .import_obj(&path)
+        .expect_err("Error should be thrown");
+
+    assert_eq!(error, ImporterError::InvalidStructure);
+}
+
+#[test]
+fn test_does_not_cache_nonexistent_file() {
```
yanchith: See my earlier comment about `has_cached(&str)`. Should apply here as well.

tests/importer.rs
```diff
@@ -0,0 +1,252 @@
...
+        .expect("Valid object should be loaded");
+    let models_2 = importer
+        .import_obj(&path)
+        .expect("Valid object should be loaded");
+
+    assert_eq!(models_1, models_2);
+}
+
+#[test]
+fn test_adds_two_different_files_with_the_same_contents() {
```
yanchith: This test actually tests performance, not correctness. Not sure If I would keep it.

yanchith (changes requested):
Nice job overall! I have some general remarks:

Regarding `importer.rs`, I am not sure it is a good idea to expose that much of a module internal structure just to write tests that test _exactly_ that internal structure. This will be very fragile and we will be rewriting tests for every change of the implementation, not just a change that actually breaks anything. I would instead suggest to spend some time on the module's API, and write tests to that. See inline comments for more on this.

I also have some other, misc inline suggestions. Otherwise the PR seems solid!

src/main.rs
ondrowan: No, but it's not a part of this PR. Will remove it later.

src/main.rs
ondrowan: :+1: Forgot about that. Will check the possibilities.

src/importer.rs
ondrowan: I'm not familiar with 100% of the spec, but at least referenced filenames (materials) can be utf8. I'm not sure about names of models, for example. I'll do some additional research.

src/importer.rs
ondrowan: I'm not any more happy about this than you. I'm used to "we are all adults" approach from Python, so I made all of these properties public. Was thinking about getters, but I found it to be unnecessary. Will try to think of possibilities here.

tests/importer.rs
ondrowan: Oh, nice, I didn't know about this syntax! 👍

tests/importer.rs
ondrowan: Yes, it's a possibility, I'll look into it.

tests/importer.rs
ondrowan: I think so. Returning of parsed models is part of that method and we want to be sure it's returning correct ones and not something from the neighborhood.

tests/importer.rs
ondrowan: Name of the function is actually wrong. `adds` -> `caches`.

And since multiple files can have the same checksum, it makes sense to test whether it was properly cached or not.

src/importer.rs
ondrowan: You got me at 🍀 . Will look into this.

yanchith: Don't they mark private fields with `_` basically meaning: only use/modify this if you really know what you are doing? Our situation looks different.

tests/importer.rs
yanchith: I was basing the comment on the implementation, not the name, but you are right with the name change 👍 

I initially suggested to just remove the test case because I didn't want to over-complicate the API just because a test needs some data, but if you really want to keep the test case, this could work:

We could modify the `has_cached(&str)` I mentioned earlier to also return the number of paths pointing to the model and assert it increments when you load the second model.

src/main.rs
ondrowan: Done.

src/importer.rs
ondrowan: Done.

src/importer.rs
ondrowan: Done.

tests/importer.rs
ondrowan: Done.

src/importer.rs
ondrowan: I've changed it to `fs::read`. It seems to be working well as `tobj` checks for unicode validity anyway. If invalid character is encountered, `ImporterError::InvalidStructure` is returned.

src/importer.rs
ondrowan: Done.

ondrowan:
@yanchith Everything should be fixed now. As we've discussed elsewhere, tests are in temporary state right now. Some of them don't even make sense in current state, however I kept them just to remind myself what to rewrite later on. All of them are passing anyway.

src/importer.rs
```diff
@@ -0,0 +1,268 @@
...
+
+        Ok(models)
+    }
+
+    /// FIXME: This is a poor man's testing method for cache contents. It should
+    /// be removed once cacher is removed from this structure and proper unit
+    /// tests are written for it.
+    pub fn is_cached(&self, path: &str, checksum: u32) -> bool {
+        if self.path_metadata.contains_key(path) {
+            self.path_metadata.get(path).expect("").checksum == checksum
```
yanchith: `.expect("")` doesn't make much sense. You can either use `.unwrap()` (provided this is temporary code) or, even better, not use `.contains_key(path)` at all - `get(path)` returns an Option you can `if let`.

src/importer.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// and its timestamp is identical, parsed models are returned from cache.
+    /// Otherwise, file is read, checksum calculated and cache is checked whether
+    /// given file contents were already saved. If not, obj file is parsed and
+    /// cached.
+    pub fn import_obj(&mut self, path: &str) -> Result<Vec<Model>, ImporterError> {
+        let mut file = fs::File::open(path)?;
+        let file_modified = file
+            .metadata()
+            .and_then(|metadata| metadata.modified())
+            .expect("obj file should return its modified timestamp");
```
yanchith: Any particular reason why you didn't just `?` the error? You already have the `From` impl for it. Unless I am missing something, the code would be both shorter and more resilient.

src/importer.rs
```diff
@@ -0,0 +1,268 @@
...
+            if path_metadata.get().last_modified == file_modified {
+                return Ok(self
+                    .loaded_models
+                    .get(&path_metadata.get().checksum)
+                    .expect("Should get loaded models by obj file's checksum")
+                    .clone());
+            }
+        }
+
+        let file_size = file.metadata().map(|m| m.len() as usize + 1).unwrap_or(0);
```
yanchith, line 98: Why the `+ 1`? EOF?

src/importer.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// cached.
+    pub fn import_obj(&mut self, path: &str) -> Result<Vec<Model>, ImporterError> {
+        let mut file = fs::File::open(path)?;
+        let file_modified = file
+            .metadata()
+            .and_then(|metadata| metadata.modified())
+            .expect("obj file should return its modified timestamp");
+
+        // If paths and timestamps match, we can just return cached models.
+        if let Entry::Occupied(path_metadata) = self.path_metadata.entry(path.to_string()) {
```
yanchith: `if let Some(path_metadata) = self.path_metadata.get(path)`? (`path.to_string()` is not necessary)

yanchith (dismissed):
Let's do this 🎉  (although please fix the minor nits I commented on first)

Would you also please write up a trello issue of what we need to hammer into shape afterwards?

From what I remember, we should at least:

- Factor out the non-io part of the importer
- Convert most of the current integration tests to unit tests
- Stop checksums and caching from leaking into the API (don't have any methods like `is_cached(...)`
- Make `test_does_not_cache_the_same_unchanged_obj_file_twice` assert what it says
- Make `test_does_not_cache_nonexistent_file` assert what it says

src/importer.rs
ondrowan: Most likely. This is stolen from `std::fs::read`.

PR #5, opened 2019-08-11
Add orbit camera and input system

yanchith:
Adds a simple orbit camera based on spherical coordinates and a proto input system.

The camera supports:

- panning in the screen plane _(rmb + movement)_
- panning in the XY plane _(shift + rmb + movement)_
- rotation around origin _(lmb + movement)_
- step zoom _(wheel)_
- smooth zoom _(lmb + rmb + movement)_
- origin reset _(A)_

The keybindings are not set in stone and can be changed later.

@ondrowan @janper I did some testing on my two devices, but please do test the camera behavior on various mouse devices and touchpads, and various OSes and give feedback 

The input system is factored out to a separate struct: `InputManager`. Each frame, the manager needs to be reset via `start_frame()` and filled with events it needs to process via  `process_event()`. Each processed event is accumulated into the `InputState`. A shared pointer to `InputState` can be queried at any time, but the state only contains useful information after all the events have been processed.


src/main.rs
```diff
@@ -91,52 +103,50 @@ fn main() {
...
-                        );
-
-                        swap_chain = create_swap_chain(&device, &surface, physical_size);
-                        viewport_renderer.set_screen_size(&mut device, physical_size);
-                    }
-                    _ => (),
-                }
+        input_manager.start_frame();
+        event_loop.poll_events(|event| input_manager.process_event(event));
+        {
```
ondrowan: Does this scope need to be here? From what I've tried, it doesn't change anything and it looks kinda confusing being right after `poll_events` (looks like it's part of some callback or loop).

src/camera.rs
```diff
@@ -0,0 +1,162 @@
...
+    fn compute_eye(&self) -> Point3<f32> {
+        let x = self.radius * self.azimuthal_angle.cos() * self.polar_angle.sin();
+        let y = self.radius * self.azimuthal_angle.sin() * self.polar_angle.sin();
+        let z = self.radius * self.polar_angle.cos();
+
+        self.origin + Vector3::new(x, y, z)
+    }
+}
+
+fn clamp(x: f32, min: f32, max: f32) -> f32 {
```
ondrowan, line 156: Just FYI that `f32` already has [clamp implementation](https://doc.rust-lang.org/stable/src/std/f32.rs.html#986-992) in nightly. Might be a space for `FIXME`!

src/camera.rs
```diff
@@ -0,0 +1,162 @@
...
+    f32::max(min, f32::min(max, x))
+}
+
+fn fmod(x: f32, y: f32) -> f32 {
```
ondrowan: How is this different from just using modulo operator? From what I've tried, the results are mostly the same. Either way, it took me a while to figure out what this does and why it does it, so if it's useful, it deserves a docstring.

src/camera.rs
```diff
@@ -0,0 +1,162 @@
...
+        let camera_rotation =
+            Rotation3::new(Vector3::z() * (self.azimuthal_angle - f32::consts::FRAC_PI_2));
+        self.origin += camera_rotation * ground_translation;
+    }
+
+    pub fn pan_screen(&mut self, dx: f32, dy: f32) {
+        let pan_factor = self.options.speed_pan * self.radius / self.options.radius_max;
+        let ground_translation = Vector3::new(dx, dy, 0.0) * pan_factor;
+        let camera_rotation =
+            Rotation3::new(Vector3::z() * (self.azimuthal_angle - f32::consts::FRAC_PI_2));
```
ondrowan, line 77: These 3 lines are exactly the same as in `pan_ground`. Could it perhaps have some common name?

src/camera.rs
```diff
@@ -0,0 +1,162 @@
...
+        };
+
+        self.radius = clamp(new_radius, self.options.radius_min, self.options.radius_max);
+    }
+
+    pub fn zoom_step(&mut self, zoom_steps: i32) {
+        let zoom_speed = ZOOM_SPEED_BASE.powf(self.options.speed_zoom_step);
+
+        let mut new_radius = self.radius;
+        match zoom_steps.cmp(&0) {
```
ondrowan, line 117: Since `Ordering::Equal` isn't needed, is this more readable than simple `if / else`?

src/input.rs
```diff
@@ -0,0 +1,191 @@
...
+
+#[derive(Debug, Default)]
+pub struct InputManager {
+    lmb_down: bool,
+    rmb_down: bool,
+    shift_down: bool,
+    input_state: InputState,
+}
+
+impl InputManager {
```
ondrowan, line 26: We should talk briefly about design of these systems. `InputManager` is currently too monolithic and I think it's time to discuss some form of component system we might later use. As far as I can tell, there are currently 3 entities we should separate: Application, UI and Camera (perhaps even selected objects with contextual input handler in the future). All three of them need to handle rendering and inputs. E.g. it's already visible that some inputs are contextual if it comes to UI vs Camera (you don't wanna do anything with camera if you're filling out some input). So they should be abstracted into their own entities that handle processing of these components. It doesn't need to be any library, just some primitive implementation for our needs. And it certainly doesn't need to be ready for current milestone, but we should create a task in Trello for future refactoring.

src/input.rs
```diff
@@ -0,0 +1,191 @@
...
+                        // Cmd+Q for macOS
+                        #[cfg(target_os = "macos")]
+                        (
+                            Some(winit::VirtualKeyCode::Q),
+                            winit::ElementState::Pressed,
+                            winit::ModifiersState {
+                                logo: true,
+                                shift: false,
+                                ctrl: false,
+                                alt: false,
```
ondrowan: This is slightly incorrect. Even if you add `alt` to the mix, it should close the application. `ctrl` and `shift` are used for system specific actions (logout and close all opened apps respectively). I think this should be rewritten as:

```rust
winit::ModifiersState {
    logo: true,
    shift: false,
    ctrl: false,
    ..
}
```

There's also other (primary) way to quit MacOS app and that's `cmd + w`. It means "Close current window" (or tab). Since we'll most likely have just 1 window it's gonna be identical to `cmd + q`. If we'd have multiple windows or tabs, we'll need to implement `cmd + shift + w` to close all windows / tabs as well and behaviour or `cmd + w` would change to close just active tab / window. `alt` behaviour of these shortcuts is the same as described above.

There's also an important difference between `cmd + w` and `cmd + q`. `cmd + w` keeps application loaded in memory. This can be observed by the app staying in dock and loading much faster. `cmd + q` kills the app. I don't know how to implement the former, but we can look into it (much) later. Either way a Trello task should be created with everything that's not gonna be currently implemented.

You can test all of this behaviour in some web browser or Finder (although it cannot be closed with `cmd + q`).

src/input.rs
```diff
@@ -0,0 +1,191 @@
...
+                                _ => (),
+                            }
+                        }
+                    }
+                    winit::MouseScrollDelta::LineDelta(_, y) => {
+                        if !gui_captured_mouse {
+                            match y.partial_cmp(&0.0) {
+                                Some(Ordering::Greater) => self.input_state.camera_zoom_steps += 1,
+                                Some(Ordering::Less) => self.input_state.camera_zoom_steps -= 1,
+                                _ => (),
```
ondrowan, line 181: As asked above, is this and `partial_cmp` in `winit::MouseScrollDelta::PixelDelta` prefered to simple `if / else`?

src/main.rs
```diff
@@ -1,14 +1,11 @@
+use std::f32;
```
ondrowan: Is this needed for something?

src/viewport_renderer.rs
```diff
@@ -1,13 +1,13 @@
 use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::error;
+use std::f32;
```
ondrowan: The same as in one of the previous comments. Is this needed for anything?

ondrowan: And I found one more occurence in the very first file, but I'm now lazy to rearrange the comments.

ondrowan (changes requested):
It seems to be working smoothly for me, good job! However, I haven't had a chance to try it in a proper environment. I'm currently stuck with MacOS and have only trackpad available. Some of the shortcuts were too awkward to try with this setup, so I'll retest it once I get home. This however doesn't block this PR since we can fix possible issues later on.

src/main.rs
yanchith: Just wanted the reference to go out of scope. Seemed more readable to me, but I can remove if you insist.

src/camera.rs
yanchith: 😂 

Yeah, I know about it, but it seems unlikely to stabilize soon. I'll add the FIXME though 😺

src/main.rs
yanchith: Nope, not in this file. Weird that there is no unused lint. I'll remove.

src/viewport_renderer.rs
yanchith: Nope, not in this file. Weird that there is no unused lint. I'll remove.

src/camera.rs
yanchith: Thought about it, but didn't do it since these two occurrences are the only ones, and will likely remain that way. I would definitely extract to a function if there was a third.

src/camera.rs
yanchith: Good catch! I was confused from writing shaders, where there is not `%` on floats, only a `mod()` function. I'll remove.

src/camera.rs
yanchith: Strangely enough, I find `match` more readable than `if/else` (except maybe for a single `if` or `if let`), but I can change it if we agree otherwise.

src/input.rs
yanchith: We should definitely discuss, but I would leave it until after the first milestone. The thing is we don't currently know how it should look like yet. We'll have a better picture once we have a more complete application.

As can be seen from the `gui_captured_*` flags, the systems should know about each other "somehow". My current plan was to process gui first, get those two flags out of it, and pass them into our own input processing (the flags are present on imgui).

As for `InputManager` being monolithic, that is on purpose. We need to have all the controls defined somewhere, with clear precedence and priority, and possibly the ability to swap to different control layouts to simulate other 3d software - the input manager would be the one consuming raw events and producing application events dependent on its keymap. I was also envisioning the `InputManager` to be the insulation layer between raw windowing API and our application, so that various other components need know nothing about window events, possibly even allowing us to swap the windowing abstraction without rewriting a lot of other code. The intention was to map from `winit` to our own model (currently the `InputState` struct). The inspiration for that came from this tutorial: https://lokathor.github.io/learn-gfx-hal/01_introduction.html. I'd like to hear your thoughts on this in person.

If we are going to clean things up some way, I would first suggest to create an `Application` struct holding most of the current `main()` function's logic. Usage could look like:

```rust
// pseudocode

app.start_frame(); // reset state, counters, and whatnot
app.process_inputs(); // pipe inputs to gui and our own components
app.process_computations(); // dispatch computations to jobs, collect job results
app.draw();
```

src/input.rs
yanchith: I'll fix the `alt` here and create an issue for the rest.

src/camera.rs
yanchith: Done

src/camera.rs
yanchith: Done

src/input.rs
yanchith: Done

src/main.rs
yanchith: Done

src/main.rs
yanchith: Removed.

src/viewport_renderer.rs
yanchith: Done. I left it in `camera.rs` though, it is actually needed there.

ondrowan (approved):

PR #6, opened 2019-08-16
Add matcap shading

yanchith:
This modifies the viewport renderer to shade the geomeotry with the material
capture (matcap) technique. Matcap entails using the .xy components of
a viewspace projected normal to sample a texture containing
pre-renderered lighting information. Note that the texture included in
this PR is not very good and we should create a better one later.

**Some changes:**

There is a new `uv_cube()` primitive geometry. All primitives
live in RH coordinate space with XY being the ground and Z growing upwards.

Matcap means we now always need the geometry normals present (and will
have to compute them if they are not). `Geometry` struct is created to
encapsulate what the renderer needs. Its `indices` field is optional.

The renderer currently decodes the matcap texture during
initialization, and the texture is baked to the program binary (might
want to change that to runtime loading later).

Since we need to also use the view matrix on the shaders to transform
object normals to view space, camera now needs to export its matrices
separately.

@ondrowan sorry for the messy diff, I had to rewrite significant portions of renderer code.


run-with-vulkan-validation.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
ondrowan: I'm not a fan of having additional shell scripts if it can be avoided. From what I've searched, it's possible to set environment variables for Rust's compile environment by outputting `cargo:rustc-env=...` from within build script, e.g. `cargo:rustc-env=VK_INSTANCE_LAYERS=VK_LAYER_KHRONOS_VALIDATION`. These can be then retrieved using `env!` (or `option_env!`) macro. What I had in mind is setting this `rustc-env` variable in case the app is compiled in debug mode and if it's set and, for example, `--feature vulkan-validation` is also present, checking for `VK_LAYER_PATH` env variable as well and warning them if it's not set. This way (if it works as I expect) we get rid of the shell file and it'll be up to user to set `VK_LAYER_PATH` themselves. This would have to be documented.

If you decide to keep this file, at least change its permissions to allow execution and document it in readme file.

But as everything these days, it's not crucial for the milestone, so it can be deferred to Trello for the time being.

src/primitives.rs
```diff
@@ -1,47 +1,183 @@
...
-        v(-1.0, -1.0, 0.0, position, scale),
-    ]
+//! The `primitives` module contains parametric primitives for testing
+//! renderer development. Each primitive contains both vertex
+//! positions and normals.
+//!
+//! The primitives live in a right-handed coordinate space with the XY
+//! plane being the ground and and Z axis growing up.
+
+use std::f32;
```
ondrowan: This can be safely removed.

src/main.rs
```diff
@@ -167,10 +165,16 @@ fn main() {
...
 
-        viewport_renderer.set_camera_matrix(&mut device, camera.matrix());
+        // Camera matrices have to be uploaded when either window
+        // resizes, or anything the camera moves. Seems easier to just
+        // upload them always.
+        viewport_renderer.set_camera_matrices(
+            &mut device,
+            &camera.projection_matrix(),
+            &camera.view_matrix(),
+        );
```
ondrowan, line 177: Isn't this one of the cases which melts my PC? I think we should start making a list of fps / performance related optimizations.

src/viewport_renderer.rs
```diff
@@ -2,23 +2,132 @@ use std::collections::HashMap;
...
+    // These are defined as [f32; 4] for 2 reasons:
+    //
+    // - point vs vector clarity
+    // - padding: vec3 has the same size as vec4 in std140 layout.
+    //   https://www.khronos.org/opengl/wiki/Interface_Block_(GLSL)
+    //
+    // We might decide to pack more useful data than just 0/1 in the
+    // last component later.
+    pub position: [f32; 4],
+    pub normal: [f32; 4],
```
ondrowan, line 42: Is the 4th byte always ignored when dealing with GPU functions that don't expect it then?

src/viewport_renderer.rs
```diff
@@ -2,23 +2,132 @@ use std::collections::HashMap;
...
 
+/// The geometry containing index and vertex data in the format as
+/// will be uploaded on the GPU.
+#[derive(Debug, Clone, PartialEq)]
+pub struct Geometry {
+    indices: Option<Vec<Index>>,
+    vertex_data: Vec<Vertex>,
+}
+
+impl Geometry {
```
ondrowan, line 58: This asks for a test or two.

ondrowan (changes requested):
It works on my computer (tm). Good job!

What's the expected state of normals in the future then? Are we gonna compute them if they're not present? I assumed that much from our conversation on Slack, but I guess it's something we're gonna finish after current milestone, right?

src/viewport_renderer.rs
yanchith: Not sure I understand the question. The 4th byte is expected in the shaders, they ask for `vec4` and it has meaning (more on that later).

I think (but am not 100% sure) that if you changed the shaders to only accept `vec3`, creating `matcap_render_pipeline` would or should fail (I don't thing wgpu does any shader related validation yet), because it specifies it's gonna push down a `Float4`. If we passed down data where `position` and `normal` were just a `[f32; 3]`, the first 4 bytes of the next value would be read when reading the first `vec4` - not good. So we have to match what the render pipeline and shaders say with our `#[repr(C)` struct.

The 4th component of a `vec4` for positional data is "a flag" whether the `vec4` is a point or a vector. Multiplication with a matrix in homogeneous coordinates behaves differently: vectors are not translated by a transformation matrix, only rotated and scaled. (I think)

src/primitives.rs
yanchith: Damn. Yeah, I was using it earlier, only to realize they don't have the constant I needed 😄

src/main.rs
yanchith: I believe this shouldn't be significant. Per-frame data update usually happens on a much grander scale than 2 matrices. I do agree overall that we should profile for expensive things... maybe RenderDoc will help... once I learn how to use it 😂 

A related performance problem may be that we multiply those 2 matrices for every vertex in the VS, and we work with unusually large geometry. We could improve that by sending a pre-multiplied version of the matrices also.

run-with-vulkan-validation.sh
yanchith: I'll remove it then and add an entry to the readme on how to run with the vulkan validation layers on window/linux. Don't think complicating the build script is worth it.

yanchith:
@ondrowan yeah, We'll need to compute the normals if they are missing, the ~geometry is~ OBJ files are perfectly valid without them. We don't have to do it for the milestone though.

src/main.rs
yanchith: Also... maybe it's time to try flamegraph 🔥

src/viewport_renderer.rs
yanchith: Added some more asserts and tests for them. I don't think this is the place to do in-depth validation, so I stopped there.

run-with-vulkan-validation.sh
yanchith: So... no `build.bat`? 😸

src/viewport_renderer.rs
```diff
@@ -63,10 +63,18 @@ impl Geometry {
         vertex_positions: Vec<[f32; 3]>,
         vertex_normals: Vec<[f32; 3]>,
     ) -> Self {
+        assert!(
+            !vertex_positions.is_empty(),
+            "Vertex positions must not be empty",
+        );
```
ondrowan, line 69: I don't want to nitpick too much, but this actually crashes app even with valid, although empty, obj files. If this is expected behaviour, there should be also some guard before calling this method.

src/viewport_renderer.rs
```diff
@@ -665,3 +682,99 @@ fn wgpu_size_of<T>() -> wgpu::BufferAddress {
...
+
+        // Didn't panic!
+    }
+
+    #[test]
+    fn test_create_valid_indexed_geometry_does_not_panic() {
+        let (indices, positions, normals) = triangle_indexed();
+        let _geometry = Geometry::from_positions_and_normals_indexed(indices, positions, normals);
+
+        // Didn't panic!
```
ondrowan: Just assert indices and vertex data in both of these tests. There's logic transforming incoming data into other data, so that's what should be tested here anyway. If it doesn't fail, it didn't panic, so that's tested as well.

I know you find it to be a noise, but you'll write it once and most likely never read it ever again unless there's a bug. If it was up to me, I'd create 4 tests with 1 assert each:

* `test_geometry_from_positions_and_normals_produces_valid_vertex_data`
* `test_geometry_from_positions_and_normals_produces_valid_indices`
* `test_geometry_from_positions_and_normals_indexed_produces_valid_vertex_data`
* `test_geometry_from_positions_and_normals_indexed_produces_valid_indices`

But I'm ok with 2 tests with 2 asserts each (asserting both vertex data and indices) if you don't want that much noise.

We most likely already spent more time discussing tests than it'd take to actually write or read them.

ondrowan (changes requested):
FYI I've found couple of random bugs while doing review, both of them replicable also on current master branch, so I'll add them to Trello.

* Minimizing the window crashes the application on Windows (both vulkan and d3d12).
* There's a massive memory leak on Mac when the window is inactive. Talking about gigabytes per second. If I'm not mistaken, it happens when the application is inactive in background. E.g. I run the application, put my browser to the foreground and it suddenly starts to leak memory.

There's also small memory leak on both `metal` and `d3d12` backends, most likely caused by some bug in their implementations. I'll write it down and we should be probably good internet citizens and give it at least few minutes of investigation and file bug reports once we have time for it.

src/viewport_renderer.rs
yanchith: Exactly. As I mentioned in my earlier comments, This needs to be validated at some stage before creating. Empty geometries don't make much sense though so I made it part of the contract.

src/viewport_renderer.rs
yanchith: Yeah, seems like the `assert!(1 == 1)` scenario to me, but I'll add those asserts. I'd keep the rest of the tests. Geometry has to be validated at some point, but the renderer seems like too late a place.

yanchith:
Thanks for the bug reports! Didn't notice the leaks for sure... The https://github.com/sub-digital/HURBAN-Selector/tree/msaa branch has gfx-hal updated to 0.3 - we should check whether that still happens there, and possibly report and fix upstream.

src/viewport_renderer.rs
yanchith: Hmmm, perhaps we can allow empty geometries... They don't really harm much, except they don't do anything and are probably an error somewhere... not sure. I would keep the other asserts here. We will have to do proper validation of the geometry before it reaches the renderer, but that is a much larger topic and the conversion to renderer-style, flat buffers geometry doesn't seem the right place for all those expensive checks.

src/viewport_renderer.rs
ondrowan: Why `1 == 1`? It's input -> transformation -> output and you want to test if the transformation part does what it's supposed to do. It's not like the output is the same as input.

ondrowan (approved):

PR #7, opened 2019-08-19
Implementation of UI wrapper

ondrowan:
This is a convenience wrapper around imgui and all of its orchestration. Its practicality can be questioned, but it's the first iteration and I find it to be friendlier than the barebones alternative. All options are also hardcoded, which I (partly) don't expect to stay that way in the future.

### DIFF
--- a/src/imgui_renderer.rs
+++ b/src/imgui_renderer.rs
@@ -168,7 +168,7 @@ impl ImguiRenderer {
             sample_count: 1,
             dimension: wgpu::TextureDimension::D2,
             format: wgpu::TextureFormat::Rgba8Unorm,
-            usage: wgpu::TextureUsage::SAMPLED,
+            usage: wgpu::TextureUsage::SAMPLED | wgpu::TextureUsage::TRANSFER_DST,
         });
 
         upload_texture_immediate(
@@ -213,7 +213,7 @@ impl ImguiRenderer {
             sample_count: 1,
             dimension: wgpu::TextureDimension::D2,
             format: wgpu::TextureFormat::Rgba8Unorm,
-            usage: wgpu::TextureUsage::SAMPLED,
+            usage: wgpu::TextureUsage::SAMPLED | wgpu::TextureUsage::TRANSFER_DST,
         });
 
         upload_texture_immediate(device, &texture, width, height, data);
--- a/src/imgui_renderer.rs
+++ b/src/imgui_renderer.rs
@@ -389,8 +389,6 @@ impl ImguiRenderer {
 }
 
 struct Texture {
-    texture: wgpu::Texture,
-    sampler: wgpu::Sampler,
     bind_group: wgpu::BindGroup,
 }
 
@@ -417,11 +415,7 @@ impl Texture {
             ],
         });
 
-        Texture {
-            texture,
-            sampler,
-            bind_group,
-        }
+        Texture { bind_group }
     }
 
     pub fn bind_group(&self) -> &wgpu::BindGroup {
--- a/src/imgui_renderer.rs
+++ b/src/imgui_renderer.rs
@@ -1,0 +1,501 @@
+use std::convert::TryFrom;
+use std::mem;
+
+use imgui;
+use imgui::internal::RawWrapper;
+
+#[derive(Debug, Clone)]
+pub enum ImguiRendererError {
+    BadTexture(imgui::TextureId),
+}
+
+pub struct ImguiRenderer {
+    render_pipeline: wgpu::RenderPipeline,
+    uniform_buffer: wgpu::Buffer,
+    uniform_bind_group: wgpu::BindGroup,
+    texture_layout: wgpu::BindGroupLayout,
+    textures: imgui::Textures<Texture>,
+    clear_color: Option<wgpu::Color>,
+}
+
+impl ImguiRenderer {
+    const DRAW_VERT_SIZE: usize = mem::size_of::<imgui::DrawVert>();
+
+    pub fn new(
+        imgui: &mut imgui::Context,
+        device: &mut wgpu::Device,
+        format: wgpu::TextureFormat,
+        clear_color: Option<wgpu::Color>,
+    ) -> Result<ImguiRenderer, ImguiRendererError> {
+        // Link shaders
+
+        let vs_spv = include_bytes!(concat!(env!("OUT_DIR"), "/shaders/imgui.vert.spv"));
+        let fs_spv = include_bytes!(concat!(env!("OUT_DIR"), "/shaders/imgui.frag.spv"));
+        let vs_module = device.create_shader_module(vs_spv);
+        let fs_module = device.create_shader_module(fs_spv);
+
+        // Create ortho projection matrix uniform buffer, layout and bind group
+
+        let uniform_buffer = device.create_buffer(&wgpu::BufferDescriptor {
+            size: UNIFORM_BUFFER_SIZE,
+            // FIXME(yanchith): `TRANSFER_DST` is required because the
+            // only way to upload the buffer currently is by issueing
+            // the transfer command in `upload_buffer_immediate`. We
+            // can remove the flag once we get rid of the hack and
+            // learn to write to mapped buffers correctly.
+            usage: wgpu::BufferUsage::UNIFORM | wgpu::BufferUsage::TRANSFER_DST,
+        });
+
+        let uniform_layout = device.create_bind_group_layout(&wgpu::BindGroupLayoutDescriptor {
+            bindings: &[wgpu::BindGroupLayoutBinding {
+                binding: 0,
+                visibility: wgpu::ShaderStage::VERTEX,
+                ty: wgpu::BindingType::UniformBuffer,
+            }],
+        });
+
+        let uniform_bind_group = device.create_bind_group(&wgpu::BindGroupDescriptor {
+            layout: &uniform_layout,
+            bindings: &[wgpu::Binding {
+                binding: 0,
+                resource: wgpu::BindingResource::Buffer {
+                    buffer: &uniform_buffer,
+                    range: 0..UNIFORM_BUFFER_SIZE,
+                },
+            }],
+        });
+
+        // Create texture uniforms layout
+
+        let texture_layout = device.create_bind_group_layout(&wgpu::BindGroupLayoutDescriptor {
+            bindings: &[
+                wgpu::BindGroupLayoutBinding {
+                    binding: 0,
+                    visibility: wgpu::ShaderStage::FRAGMENT,
+                    ty: wgpu::BindingType::SampledTexture,
+                },
+                wgpu::BindGroupLayoutBinding {
+                    binding: 1,
+                    visibility: wgpu::ShaderStage::FRAGMENT,
+                    ty: wgpu::BindingType::Sampler,
+                },
+            ],
+        });
+
+        // Create render pipeline
+
+        let pipeline_layout = device.create_pipeline_layout(&wgpu::PipelineLayoutDescriptor {
+            bind_group_layouts: &[&uniform_layout, &texture_layout],
+        });
+
+        // Setup render state: alpha-blending enabled, no face
+        // culling, no depth testing
+
+        let render_pipeline = device.create_render_pipeline(&wgpu::RenderPipelineDescriptor {
+            layout: &pipeline_layout,
+            vertex_stage: wgpu::PipelineStageDescriptor {
+                module: &vs_module,
+                entry_point: "main",
+            },
+            fragment_stage: Some(wgpu::PipelineStageDescriptor {
+                module: &fs_module,
+                entry_point: "main",
+            }),
+            rasterization_state: wgpu::RasterizationStateDescriptor {
+                front_face: wgpu::FrontFace::Ccw,
+                cull_mode: wgpu::CullMode::None,
+                // Depth test is disabled, no need for these to mean anything
+                depth_bias: 0,
+                depth_bias_clamp: 0.0,
+                depth_bias_slope_scale: 0.0,
+            },
+            primitive_topology: wgpu::PrimitiveTopology::TriangleList,
+            color_states: &[wgpu::ColorStateDescriptor {
+                format,
+                // Enable alpha blending
+                color_blend: wgpu::BlendDescriptor {
+                    src_factor: wgpu::BlendFactor::SrcAlpha,
+                    dst_factor: wgpu::BlendFactor::OneMinusSrcAlpha,
+                    operation: wgpu::BlendOperation::Add,
+                },
+                alpha_blend: wgpu::BlendDescriptor {
+                    src_factor: wgpu::BlendFactor::SrcAlpha,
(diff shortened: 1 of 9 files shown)


src/main.rs
```diff
@@ -98,7 +109,10 @@ fn main() {
         };
 
         input_manager.start_frame();
-        event_loop.poll_events(|event| input_manager.process_event(event));
+        event_loop.poll_events(|event| {
+            let (ui_captured_keyboard, ui_captured_mouse) = ui.process_event(&event);
```
yanchith: I think the use of `ui_captured_{keyboard,mouse}` here is incorrect. One can know their true values only after imgui has seen all the input events for the current frame.... What needs to happen is:

1) feed all the events to imgui
2) get `ui_captured_{keyboard,mouse}` from imgui
3) feed all the events to our input manager, together with the flags

This is a bit unfortunate, because it forces us to buffer events. Pseudocode:

```
input_buffer = []
for ev in poll_events():
    input_buffer.push(ev);
    ui.process_event(ev);

(ui_captured_kbd, ui_captured_mouse) = ui.captured_keyboard(), ui.captured_mouse();
for ev in input_buffer:
    input_manager.process_event(ev, ui_captured_kbd, ui_captured_mouse);
```

Definitely not ideal, but will do as a workaround for now. We can clean this up later, but I am not sure. For more context, see: https://github.com/Gekkio/imgui-rs/issues/128

src/ui.rs
```diff
@@ -0,0 +1,92 @@
...
+            window,
+            context: imgui_context,
+        }
+    }
+
+    pub fn context(&mut self) -> &mut imgui::Context {
+        &mut self.context
+    }
+
+    pub fn update_delta_time(&mut self, delta: Instant) -> Instant {
```
yanchith: It seems very weird that imgui should be updating our time delta when we do it already... This should be just `set_delta_time()`.

The imgui implementation is this:

```rust
pub fn update_delta_time(&mut self, previous: Instant) -> Instant {
        let now = Instant::now();
        let delta = now - previous;
        let delta_s = delta.as_secs() as f32 + delta.subsec_nanos() as f32 / 1_000_000_000.0;
        if delta_s > 0.0 {
            self.delta_time = delta_s;
        } else {
            self.delta_time = f32::MIN_POSITIVE;
        }
        self.delta_time = delta_s;
        now
    }
```

We can steal the "duration to float" conversion from its implementation and use it here to set time delta on IO.

Float duration (https://github.com/rust-lang/rust/issues/54361) would be helpful but they will only be stabilized in 1.38... we can add a FIXME to use them 😂

src/shaders/imgui.vert
```diff
@@ -0,0 +1,22 @@
+#version 450
+
+layout(set = 0, binding = 0) uniform Transform {
+    vec2 u_Translate;
```
yanchith: I meanwhile started naming the variables and functions in shaders in `snake_case` only, e.g: `u_translate`. We can pick either, but if you want this one, please change the existing shaders to the same convention.

src/shaders/imgui.frag
```diff
@@ -0,0 +1,13 @@
+#version 450
+
+layout(set = 1, binding = 0) uniform texture2D u_Texture;
+layout(set = 1, binding = 1) uniform sampler u_Sampler;
+
+layout(location = 0) in vec2 v_UV;
```
yanchith: Would name this `tex_coords`  instead of `uv` to be consistent with the other shader

src/shaders/imgui.frag
```diff
@@ -0,0 +1,13 @@
...
+layout(location = 1) in vec4 v_Color;
+
+layout(location = 0) out vec4 o_Target;
+
+void main() {
+    o_Target = v_Color * texture(sampler2D(u_Texture, u_Sampler), v_UV);
```
yanchith: Would name this `o_color` instead of `o_target` to be consistent with the other shader... `target` is not a bad name too tho, so If you like it better, just rename `o_color` in the other fragment shader to `o_target`

src/imgui_renderer.rs
```diff
@@ -0,0 +1,495 @@
...
+
+    pub fn new(
+        imgui: &mut imgui::Context,
+        device: &mut wgpu::Device,
+        format: wgpu::TextureFormat,
+        clear_color: Option<wgpu::Color>,
+    ) -> Result<ImguiRenderer, ImguiRendererError> {
+        // Link shaders
+
+        let vs_spv = include_bytes!(concat!(env!("OUT_DIR"), "/shaders/imgui.vert.spv"));
```
yanchith: Could use the `include_shader!` macro defined in the viewport renderer. You will need to move it to a separate module though because of how `#[macro_use]` works.

src/imgui_renderer.rs
```diff
@@ -0,0 +1,495 @@
...
+
+        Texture { bind_group }
+    }
+
+    pub fn bind_group(&self) -> &wgpu::BindGroup {
+        &self.bind_group
+    }
+}
+
+fn upload_buffer_immediate(device: &mut wgpu::Device, buffer: &wgpu::Buffer, data: &[f32]) {
```
yanchith: Instead of `data: &[f32]` I would pass a `repr(C)` struct named `TransformUniforms` or similar, symmetric to what viewport renderer does with `GlobalMatrixUniforms` struct. This upload then becomes much less error-prone. Have a look at `upload_global_matrix_buffer`. I learned this nice trick reading the `wgpu-rs` examples.

This function won't really be general-purpose anyway, so it may as well expect a concrete struct with correct memory layout.

(I know, I know... I learned a lot about things since writing the base code you were starting with, sorry for giving you bad starting code)

src/imgui_renderer.rs
```diff
@@ -0,0 +1,495 @@
...
+        mag_filter: wgpu::FilterMode::Linear,
+        min_filter: wgpu::FilterMode::Linear,
+        mipmap_filter: wgpu::FilterMode::Linear,
+        lod_min_clamp: -100.0,
+        lod_max_clamp: 100.0,
+        compare_function: wgpu::CompareFunction::Always,
+    })
+}
+
+const UNIFORM_BUFFER_SIZE: u64 = 16;
```
yanchith: I would replace this constant at the use sites with much more meaningful: `wgpu_size_of::<T>()`. You can find the function in the viewport renderer and extract it (or just copy it for now - we will have to refactor later anyway).

The `T` would be the `TransformationUniforms` `repr(C)` struct.

src/imgui_renderer.rs
```diff
@@ -0,0 +1,495 @@
...
+            clear_color,
+        })
+    }
+
+    pub fn add_rgba32_texture(
+        &mut self,
+        device: &mut wgpu::Device,
+        width: u32,
+        height: u32,
+        data: &[u8],
```
yanchith, line 208: Let's assert `data.len() % 4 == 0` here. We can later create a wrapper (e.g.: `Rgba32Texture` for the buffer) ensuring this holds.

src/imgui_renderer.rs
```diff
@@ -0,0 +1,495 @@
...
+                alpha_blend: wgpu::BlendDescriptor {
+                    src_factor: wgpu::BlendFactor::SrcAlpha,
+                    dst_factor: wgpu::BlendFactor::OneMinusSrcAlpha,
+                    operation: wgpu::BlendOperation::Add,
+                },
+                write_mask: wgpu::ColorWrite::ALL,
+            }],
+            // Disabled depth test
+            depth_stencil_state: None,
+            index_format: wgpu::IndexFormat::Uint16, // FIXME(yanchith): may need 32bit indices!
```
yanchith, line 136: Ugh... Let's let this slide for now, but we shouldn't forget about it. If UI suddenly breaks when it gets big, remember this :))

src/imgui_renderer.rs
```diff
@@ -0,0 +1,495 @@
...
+            vertex_buffers: &[wgpu::VertexBufferDescriptor {
+                stride: u64::try_from(Self::DRAW_VERT_SIZE).unwrap(),
```
yanchith: Let's change this to `wgpu_size_of::<imgui::DrawVert>()` and remove the associated const.

src/main.rs
```diff
@@ -176,6 +190,12 @@ fn main() {
             &camera.view_matrix(),
         );
 
+        time = ui.update_delta_time(time);
```
yanchith: See my other comment about `ui.update_delta_time()`... basically we probably don't want to be using this imgui helper as we are already computing the same thing ourselves (see top of the application loop).

I think this should  be: `ui.set_delta_time(duration_last_frame);` with the duration being internally converted to a float inside the method.

src/ui.rs
```diff
@@ -0,0 +1,92 @@
+use std::time::Instant;
+
+use imgui_winit_support::{HiDpiMode, WinitPlatform};
+use wgpu::winit;
+
+/// Convenience wrapper around imgui and its winit support.
+pub struct Ui<'w> {
+    platform: WinitPlatform,
+    window: &'w wgpu::winit::Window,
```
yanchith: `winit::Window`?

src/ui.rs
```diff
@@ -0,0 +1,92 @@
...
+        (
+            self.context.io().want_capture_keyboard,
+            self.context.io().want_capture_mouse,
+        )
+    }
+
+    /// Prepares imgui frame, creates windows defined inside of `callback`,
+    /// prepares rendering and finally returns imgui's `DrawData` that can be
+    /// passed into renderer.
+    pub fn create<F>(&mut self, callback: F) -> &imgui::DrawData
```
yanchith: `create()`? Isn't `draw()` or something more accurate?

src/imgui_renderer.rs
```diff
@@ -0,0 +1,495 @@
...
+                width,
+                height,
+                depth: 1,
+            },
+            array_layer_count: 1,
+            mip_level_count: 1,
+            sample_count: 1,
+            dimension: wgpu::TextureDimension::D2,
+            format: wgpu::TextureFormat::Rgba8Unorm,
+            usage: wgpu::TextureUsage::SAMPLED,
```
yanchith: You are missing the `TextureUsage::TRANSFER_DST` bit, VK validation layers complain:

```
[2019-08-19T14:12:27Z ERROR gfx_backend_vulkan]
VALIDATION [VUID-vkCmdCopyBufferToImage-dstImage-00177 (0)] : Invalid usage flag for VkImage 0x41[] used by vkCmdCopyBufferToImage(). In this case, VkImage should have VK_IMAGE_USAGE_TRANSFER_DST_BIT set during creation. The Vulkan spec states: dstImage must have been created with VK_IMAGE_USAGE_TRANSFER_DST_BIT usage flag (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-vkCmdCopyBufferToImage-dstImage-00177)
object info: (type: IMAGE, hndl: 65)

[2019-08-19T14:12:27Z ERROR gfx_backend_vulkan]
VALIDATION [VUID-VkImageMemoryBarrier-oldLayout-01213 (0)] : vkCmdPipelineBarrier(): Image barrier 0x000000832B8FC5C8 new Layout=VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL is not compatible with VkImage 0x41[] usage flags 0x4. The Vulkan spec states: If either oldLayout or newLayout is VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL then image must have been created with VK_IMAGE_USAGE_TRANSFER_DST_BIT set (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkImageMemoryBarrier-oldLayout-01213)
object info: (type: IMAGE, hndl: 65)

[2019-08-19T14:12:27Z ERROR gfx_backend_vulkan]
VALIDATION [VUID-VkImageMemoryBarrier-oldLayout-01213 (0)] : vkCmdPipelineBarrier(): Image barrier 0x000000832B8FCD48 old Layout=VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL is not compatible with VkImage 0x41[] usage flags 0x4. The Vulkan spec states: If either oldLayout or newLayout is VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL then image must have been created with VK_IMAGE_USAGE_TRANSFER_DST_BIT set (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkImageMemoryBarrier-oldLayout-01213)
object info: (type: IMAGE, hndl: 65)
```

src/imgui_renderer.rs
```diff
@@ -0,0 +1,495 @@
...
+                width: font_texture.width,
+                height: font_texture.height,
+                depth: 1,
+            },
+            array_layer_count: 1,
+            mip_level_count: 1,
+            sample_count: 1,
+            dimension: wgpu::TextureDimension::D2,
+            format: wgpu::TextureFormat::Rgba8Unorm,
+            usage: wgpu::TextureUsage::SAMPLED,
```
yanchith: This is missing the `TextureUsage::TRANSFER_DST` bit

src/imgui_renderer.rs
```diff
@@ -0,0 +1,495 @@
+use std::convert::TryFrom;
+use std::mem;
+
+use imgui;
+use imgui::internal::RawWrapper;
```
yanchith, line 6: Is this unused?

yanchith (changes requested):
Nice job!

Except for `ui_capture_{keyboard,mouse}` and the vulkan validation layer errors, all my comments are minor and easily fixable.

Note: either the FPS computation is broken, or vsync doesn't work, got 2000fps on debug builds and over 9000 on release builds 😄
Would explain the GPU usage.

src/imgui_renderer.rs
ondrowan: There's `.raw()` call that needs it.

src/imgui_renderer.rs
```diff
@@ -129,7 +130,8 @@ impl ImguiRenderer {
             depth_stencil_state: None,
             index_format: wgpu::IndexFormat::Uint16, // FIXME(yanchith): may need 32bit indices!
             vertex_buffers: &[wgpu::VertexBufferDescriptor {
-                stride: u64::try_from(Self::DRAW_VERT_SIZE).unwrap(),
+                stride: u64::try_from(wgpu_size_of::<imgui::DrawVert>())
```
yanchith: Redundant `try_from`, `wgpu_size_of` already returns `u64` (type aliased as `wgpu::BufferAddress`)

src/main.rs
```diff
@@ -105,14 +105,41 @@ fn main() {
             let duration_running = now.duration_since(time_start);
             time = now;
 
+            let duration_last_frame_s = duration_last_frame.as_secs() as f32
```
yanchith, line 118: Can leave a FIXME here to change to https://doc.rust-lang.org/std/time/struct.Duration.html#method.as_secs_f32 when it stabilizes in 1.38.0

yanchith (approved):
Beautiful 👍

PR #8, opened 2019-08-19
Implement MSAA antialiasing for viewport renderer

yanchith:
This commit updates wgpu-rs and gfx-hal to 0.3 (needed because MSAA
didn't work on metal with 0.2) and implements MSAA for viewport renderer.

With gfx-hal 0.3 MSAA now works, but 3 different stencil-state vulkan
validation messages popped up:

- stencil reference is not set (this commit "fixes" that by setting
  it)
- stencil read mask is not set
- stencil write mask is not set

The last two don't look like they can be set with the current API. No
real problems that look related to these validation messages manifest
while the program is running on any of our backends, though.

Multi-sampling itself is implemented by the renderer allocating a MSAA
framebuffer texture ans using it as the render target while using the
original render target as the MSAA resolve target. This is only done
if multi-sampling is enabled, which is an option in viewport renderer.

This commit leaves the application in a temporary state - MSAA should
happen only once and for both the GUI and the window renderers, but
the GUI renderer is not present at the time of writing. We'll have to
refactor later.

EDIT: the vulkan validation messages:

```
[2019-08-17T10:14:16Z ERROR gfx_backend_vulkan]
VALIDATION [VUID-vkCmdDraw-commandBuffer-02701 (0)] : VkCommandBuffer 0x186f47cd378[]: Dynamic stencil read mask state not set for this command buffer.. The Vulkan spec states: If the VkPipeline object bound to the pipeline bind point used by this command requires any dynamic state, that state must have been set for commandBuffer (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-vkCmdDraw-commandBuffer-02701)
object info: (type: COMMAND_BUFFER, hndl: 1679139066744)

[2019-08-17T10:14:16Z ERROR gfx_backend_vulkan]
VALIDATION [VUID-vkCmdDraw-commandBuffer-02701 (0)] : VkCommandBuffer 0x186f47cd378[]: Dynamic stencil write mask state not set for this command buffer.. The Vulkan spec states: If the VkPipeline object bound to the pipeline bind point used by this command requires any dynamic state, that state must have been set for commandBuffer (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-vkCmdDraw-commandBuffer-02701)
object info: (type: COMMAND_BUFFER, hndl: 1679139066744)

[2019-08-17T10:14:16Z ERROR gfx_backend_vulkan]
VALIDATION [VUID-vkCmdDraw-commandBuffer-02701 (0)] : VkCommandBuffer 0x186f47cd378[]: Dynamic stencil reference state not set for this command buffer.. The Vulkan spec states: If the VkPipeline object bound to the pipeline bind point used by this command requires any dynamic state, that state must have been set for commandBuffer (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-vkCmdDraw-commandBuffer-02701)
object info: (type: COMMAND_BUFFER, hndl: 1679139066744)
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -6,7 +6,9 @@ use hurban_selector::camera::{Camera, CameraOptions};
 use hurban_selector::importer::Importer;
 use hurban_selector::input::InputManager;
 use hurban_selector::primitives;
-use hurban_selector::viewport_renderer::{Geometry, ViewportRenderer};
+use hurban_selector::viewport_renderer::{
+    Geometry, Msaa, ViewportRenderer, ViewportRendererOptions,
+};
 
 const SWAP_CHAIN_FORMAT: wgpu::TextureFormat = wgpu::TextureFormat::Bgra8Unorm;
 
@@ -55,10 +57,17 @@ fn main() {
     );
     let mut viewport_renderer = ViewportRenderer::new(
         &mut device,
-        SWAP_CHAIN_FORMAT,
         window_size,
         &camera.projection_matrix(),
         &camera.view_matrix(),
+        ViewportRendererOptions {
+            // FIXME: Msaa X4 is the only value currently working on
+            // all devices we tried. We should query the device
+            // capabilities (but how?!) to select the proper MSAA
+            // value.
+            msaa: Msaa::X4,
+            output_format: SWAP_CHAIN_FORMAT,
+        },
     );
 
     // FIXME: This is just temporary code so that we can see something
--- a/src/viewport_renderer.rs
+++ b/src/viewport_renderer.rs
@@ -176,6 +176,38 @@ impl fmt::Display for ViewportRendererAddGeometryError {
 
 impl error::Error for ViewportRendererAddGeometryError {}
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum Msaa {
+    Disabled,
+    X4,
+    X8,
+    X16,
+}
+
+impl Msaa {
+    pub fn enabled(self) -> bool {
+        match self {
+            Msaa::Disabled => false,
+            _ => true,
+        }
+    }
+
+    pub fn sample_count(self) -> u32 {
+        match self {
+            Msaa::Disabled => 1,
+            Msaa::X4 => 4,
+            Msaa::X8 => 8,
+            Msaa::X16 => 16,
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct ViewportRendererOptions {
+    pub msaa: Msaa,
+    pub output_format: wgpu::TextureFormat,
+}
+
 /// 3D renderer of the editor viewport.
 ///
 /// Can be used to upload `Geometry` on the GPU and draw it in the
@@ -183,11 +215,13 @@ impl error::Error for ViewportRendererAddGeometryError {}
 pub struct ViewportRenderer {
     geometries: HashMap<u64, GeometryDescriptor>,
     geometries_next_id: u64,
+    msaa_framebuffer_texture_view: Option<wgpu::TextureView>,
+    depth_texture_view: wgpu::TextureView,
     global_matrix_buffer: wgpu::Buffer,
     global_matrix_bind_group: wgpu::BindGroup,
-    depth_texture: wgpu::TextureView,
     matcap_texture_bind_group: wgpu::BindGroup,
     matcap_render_pipeline: wgpu::RenderPipeline,
+    options: ViewportRendererOptions,
 }
 
 impl ViewportRenderer {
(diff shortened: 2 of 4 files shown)


yanchith:
I tried updating wgpu again locally to see if the VK validation layer issues persist (and potentially report them). They seem to be gone with `69a894e91aa983021c6266fe4290023464e17113` but at the same time many breaking changes were introduced to the API with that commit. @ondrowan do we want to update to not have thousands of errors spammed in the console, or would you prefer to keep API changes to a minimum for now?

EDIT: This is the diff for the update they did in wgpu https://github.com/gfx-rs/wgpu-rs/commit/69a894e91aa983021c6266fe4290023464e17113

src/viewport_renderer.rs
```diff
@@ -176,18 +176,52 @@ impl fmt::Display for ViewportRendererAddGeometryError {
 
 impl error::Error for ViewportRendererAddGeometryError {}
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
```
ondrowan, line 179: Is `Hash` needed here?

src/main.rs
```diff
@@ -55,10 +57,17 @@ fn main() {
...
         &mut device,
-        SWAP_CHAIN_FORMAT,
         window_size,
         &camera.projection_matrix(),
         &camera.view_matrix(),
+        ViewportRendererOptions {
+            // FIXME: Msaa X4 is the only value currently working on
+            // all devices we tried. We should query the device
+            // capabilities (but how?!) to select the proper MSAA
+            // value.
```
ondrowan, line 67: I found out this bit in Apple OpenGL docs that might help:

> The best way to find out which sample modes are supported is to call the CGL function CGLDescribeRenderer with the renderer property kCGLRPSampleModes or kCGLRPSampleAlpha. You can also determine how many samples the renderer supports by calling CGLDescribeRenderer with the renderer property kCGLRPMaxSamples.

It might give you some hints while searching for it further.

ondrowan (commented):
I managed to panic the application couple of times while loading the reduced Pluto model, but got only these 2 stack traces from the events: https://gist.github.com/ondrowan/c9dea8f6ba7390f6d292edc5a1408a5d I wasn't able to reproduce these crashes again after these initial events.

Also, if you're gonna update `wgpu-rs` and refactor the app to match their API changes, do it in a separate PR.

src/main.rs
yanchith: Thanks! Usually this kind of querying is available in the graphics API, but we may be out of luck for wgpu... It may not be too terrible to have a platform specific device capability detectors we can later remove once wgpu starts supporting capability querying.

src/viewport_renderer.rs
yanchith: It is not. I think I read somewhere you want to by default implement everything you can guarantee will stay implemented, and this seems like it could since `Msaa` is a very simple enum. But because we are an application, we don't need to think too heavily about backwards compatibility and removing trait implementations...

Not really sure what to do. Adding it later if we need it should be ok as well. Should we remove `PartialEq` and `Eq` too then?

yanchith:
> Also, if you're gonna update wgpu-rs and refactor the app to match their API changes, do it in a separate PR.

You can see the changes here: https://github.com/sub-digital/HURBAN-Selector/compare/msaa...wip-udpate-wgpu

I was asking if we want to do it now because without the update this branch produces vulkan validation warnings, so it's correctness vs git-convenience.

yanchith:
> I managed to panic the application couple of times while loading the reduced Pluto model

I think these should not be related to the validation warnings, but I am not sure. If you manage to reproduce, can you try reproduce on the https://github.com/sub-digital/HURBAN-Selector/tree/wip-udpate-wgpu branch too?

src/main.rs
ondrowan: Might be worthwhile to at least create an issue in their repo.

src/viewport_renderer.rs
ondrowan: I don't know what performance impact it has and couldn't quickly duckduck anything relevant. I'd remove it either way for clarity reasons. `Eq` is actually used so it cannot be removed.

ondrowan:
Sure, then update wgpu in this branch.

I managed to reproduce the "add with overflow" panic on the other branch as well. It seems super random and it'll have something to do with minimizing bug. Sometimes when I open a file to import, the application jumps into background after the open dialog is closed and that seems to be crashing it. I'll consider it to be a duplicate until proven otherwise.

I cannot duplicate the other bug anymore.

ondrowan (approved):

PR #9, opened 2019-08-24
Add zoom-to-fit (bounding sphere)

yanchith:
To add the functionality, some inspect-able geometry representation
needed to be created (previous renderer geometry has all fields module
private). I took this as an opportunity to start sketching what the
"editor geometry" structure would look like.

Former `primitives` module has been moved to the `geometry` module,
and the primitive constructors now create the editor geometry. A
conversion exists from editor geometry to renderer geometry.

Zoom-to-fit itself is implemented by finding a bounding sphere of the
scene and using it to change zoom and origin (but not the angle) on
the camera.

The zoom seems to be very conservative and there is a good chance I've got an error in the math somewhere.


src/geometry.rs
```diff
@@ -0,0 +1,498 @@
...
+/// the data in those lists.
+///
+/// One such index type (`Index::TriangleFace`) binds vertices and
+/// normals in triangular faces. `Index::TriangularFace` is always
+/// ensured to have counter-clockwise winding.
+///
+/// The geometry data lives in right-handed coordinate space with the
+/// XY plance being the ground and Z axis growing upwards.
+#[derive(Debug, Clone, PartialEq)]
+pub struct Geometry {
```
janper, line 20: Is this the type I should review?

src/geometry.rs
```diff
@@ -0,0 +1,498 @@
...
+    indices: Vec<Index>,
```
janper: You are being very specific with naming `vertices` and `normals`, so why not calling these `faces`? You might in the future need to implement other indexing data, such as UVW attribution etc.

also you quite nicely suggested to treat the faces as an enum:
```
enum Index {
    Triangle(u32, u32, u32),
    Quad(u32, u32, u32, u32),
}
```
This looks really nice (in a rusty way) and explicit, however I would call it `Face` (which it is) rather than `Index`.

Also let's discuss how should we treat other vertex properties, such as normals, texturing coords etc. What about doing it the OBJ way?

```
# The first number is the point,
# then the slash,
# and the second is the texture point
f 1/1 2/2 3/3 4/4
```
so eventually maybe something like this:
```
enum Index {
    Triangle(u32, u32, u32),
    Quad(u32, u32, u32, u32),
}

struct Face {
    vertices: Index,
    normals: Index,
    textures: Option<Index>,
    vertex_color: Option<Index>,
    some_unknown_properties: Option<Index>,
}

struct Geometry {
    faces: Vec<Face>,
    vertices: Vec<Point3<f32>>,
    normals: Option<Vec<Vector3<f32>>>,
    uvws: Option<Vec<Vector3<f32>>>,
    colors: Option<Vec<RGB>>,
}
```

src/geometry.rs
yanchith: yes :)

src/viewport_renderer.rs
```diff
@@ -40,23 +44,110 @@ pub struct Vertex {
...
+            let normals_len = normals.len();
+            let len = vertices_len.max(normals_len);
+
+            let indices_len = geometry.triangle_face_indices_len();
+            let mut indices: Vec<RendererIndex> = Vec::with_capacity(indices_len);
+            let mut vertex_positions: Vec<Point3<f32>> = vec![Point3::origin(); len];
+            let mut vertex_normals: Vec<Vector3<f32>> = vec![Vector3::zeros(); len];
+
+            match vertices_len.cmp(&normals_len) {
+                Ordering::Less => {
```
yanchith: I should test this with a compressed version of `geometry::uv_cube()` which has more normals than vertices. The conversion should decompress the vertices correctly.

src/viewport_renderer.rs
```diff
@@ -40,23 +44,110 @@ pub struct Vertex {
...
+                        vertex_positions[n2] = vertices[v2];
+                        vertex_positions[n3] = vertices[v3];
+
+                        // Copy normal data as is
+                        vertex_normals[n1] = normals[n1];
+                        vertex_normals[n2] = normals[n2];
+                        vertex_normals[n3] = normals[n3];
+                    }
+                }
+                Ordering::Greater | Ordering::Equal => {
```
yanchith: I should test this with a simple triangle/quad with only one normal. The conversion should decompress the normal and replicate it across vertices.

src/geometry.rs
yanchith: I believe what is in the code essentially conforms to what you wrote. The lists of data (vertices, normals) are variable-length. They are brought together by indices, which could potentially be many types, but are just of one type at the moment: `TriangleFaceIndex`.

> You are being very specific with naming vertices and normals, so why not calling these faces? You might in the future need to implement other indexing data, such as UVW attribution etc.

So the other kinds of indices (those that are not faces) should not be in the list of indices, but rather in a separate list? I can't really imagine what other kinds of indices we will need in the future, so you know better if they should be in a single list or multiple.

In that case, yeah, we can have just `faces: Vec<Face>` where:

```rust
enum Face {
    Triangle(TriangleFace),
    Quad(QuadFace),
}

struct TriangleFace {
    pub vertices: (u32, u32, u32),
    pub normals: Option<(u32, u32, u32)>,
}

struct QuadFace {
    pub vertices: (u32, u32, u32, u32),
    pub normals: Option<(u32, u32, u32, u32)>,
}
```

Since we don't have any other kind of indices now anyway, I'll make this change to the code now.

yanchith: Done

yanchith:
So after playing around with `RendererGeometry::from_geometry()` I found out this is still not correct. Everything else can be reviewed, but the contents of that function will likely still change.

src/convert.rs
```diff
@@ -0,0 +1,26 @@
...
+///
+/// # Panics
+/// Panics if the conversion returns an error.
+pub fn cast_usize<T>(n: T) -> usize
+where
+    T: TryInto<usize>,
+    <T as TryInto<usize>>::Error: Debug,
+{
+    n.try_into().expect("Expected N to fit in usize")
+}
```
ondrowan, line 26: While trying out generating of docs, I found out these functions weren't present since they're in private module. There's a `--document-private-items` flag you most likely knew about, but it'd be better if its usage was explicitly described in readme. I'm gonna add a Trello task for documenting this.

src/geometry.rs
```diff
@@ -0,0 +1,586 @@
...
+        Self {
+            faces: faces.into_iter().map(Face::Triangle).collect(),
+            vertices,
+            normals: Some(normals),
+        }
+    }
+
+    /// Return a view of all triangle faces in this geometry. Skip all
+    /// other types of faces.
+    pub fn triangle_faces<'a>(&'a self) -> impl Iterator<Item = TriangleFace> + 'a {
```
ondrowan: I'd call this `triangle_faces_iter` to match naming of other functions that return iterators.

src/geometry.rs
```diff
@@ -0,0 +1,586 @@
...
+}
+
+fn tf_vn(a: u32, b: u32, c: u32) -> TriangleFace {
+    TriangleFace {
+        vertices: (a, b, c),
+        normals: Some((a, b, c)),
+    }
+}
+
+fn tf_vn_separate(va: u32, vb: u32, vc: u32, na: u32, nb: u32, nc: u32) -> TriangleFace {
```
ondrowan: Wouldn't it be better to stick to x, y, z in these last 2 functions?

src/main.rs
```diff
@@ -76,7 +77,26 @@ fn main() {
...
+        geometry::plane_same_len([0.0, 0.0, 20.0], 10.0),
+        geometry::plane_var_len([0.0, 0.0, -20.0], 10.0),
+    ];
+
+    let mut scene_renderer_geometry_ids = Vec::with_capacity(scene_geometries.len());
+    for geometry in &scene_geometries {
+        let renderer_geometry = RendererGeometry::from_geometry(geometry);
+        let renderer_geometry_id = viewport_renderer
+            .add_geometry(&device, &renderer_geometry)
+            .unwrap();
```
ondrowan: `expect` instead.

src/viewport_renderer.rs
```diff
@@ -40,23 +43,109 @@ pub struct Vertex {
 // FIXME: @Optimization Determine u16/u32 dynamically per geometry to
 // save memory
 /// The geometry indices as uploaded on the GPU.
-pub type Index = u32;
+pub type RendererIndex = u32;
 
-/// The geometry containing index and vertex data in the format as
-/// will be uploaded on the GPU.
+/// The geometry containing index and vertex data in same-lenght
```
ondrowan: lenght -> length

src/viewport_renderer.rs
```diff
@@ -40,23 +43,109 @@ pub struct Vertex {
...
+            // of the index that constructed the vertex.
+            for triangle_face in geometry.triangle_faces() {
+                let v = triangle_face.vertices;
+                let n = triangle_face
+                    .normals
+                    .expect("Normal indices must be present if normals are");
+
+                for (vi, ni) in &[(v.0, n.0), (v.1, n.1), (v.2, n.2)] {
+                    let vi = *vi;
+                    let ni = *ni;
```
ondrowan: Isn't `for &(vi, ni) ...` more idiomatic than this?

ondrowan (changes requested):
It works well apart from some minor things that need to be fixed.

src/geometry.rs
yanchith: Those aren't aren't coordinates though, but indices to vertices and normals. We can call them v1, v2, v3, etc... Or anything you suggest.

src/convert.rs
yanchith: I'll add docs for these here

ondrowan (approved):

PR #10, opened 2019-08-26
Model loading UI

ondrowan:
This is by no means a masterpiece of a code, but it's temporary and it's gonna be deleted in 2 weeks, so let's review this accordingly.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,3 +1,6 @@
+use std::collections::HashMap;
+use std::env;
+use std::fs;
 use std::time::Instant;
 
 use wgpu::winit;
@@ -6,7 +9,6 @@ use hurban_selector::camera::{Camera, CameraOptions};
 use hurban_selector::imgui_renderer::ImguiRenderer;
 use hurban_selector::importer::Importer;
 use hurban_selector::input::InputManager;
-use hurban_selector::primitives;
 use hurban_selector::ui;
 use hurban_selector::viewport_renderer::{
     Geometry, Msaa, ViewportRenderer, ViewportRendererOptions,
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -104,6 +88,31 @@ fn main() {
     let time_start = Instant::now();
     let mut time = time_start;
 
+    // Temporary model list
+
+    let models_dir =
+        env::var_os("MODELS_DIR").unwrap_or_else(|| env::current_dir().expect("").into());
```
yanchith: Might as well `unwrap()` if you `expect("")`

src/main.rs
```diff
@@ -104,6 +88,31 @@ fn main() {
...
+    for obj_path_result in obj_path_results {
+        let obj_path = obj_path_result.expect("Should read directory entry");
+        let path = obj_path.path();
+
+        if let Some(ext) = path.extension() {
+            if ext == "obj" {
+                let filename = obj_path
+                    .file_name()
+                    .into_string()
+                    .expect("Filename should be converted into String");
```
yanchith: Maybe write a more descriptive message here... "Filename UTF-8 conversion failed" or something

src/main.rs
```diff
@@ -227,21 +239,52 @@ fn main() {
...
+                            model.vertex_normals,
+                        );
+                        let model_handle =
+                            viewport_renderer.add_geometry(&device, &geometry).unwrap();
+
+                        for model in dynamic_models.iter() {
+                            viewport_renderer.remove_geometry(*model);
+                        }
+
+                        dynamic_models.clear();
```
yanchith, line 269: This is a leak. You need to `viewport_renderer.remove_geometry(id)` for each one you cleared. You can use the handy `Vec::drain()` for that (https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain)

Later we might think about adding a `Drop` impl to `RendererGeometryId` (or is still called `GeometryId` in master? 😂 ) that would remove itself from the renderer automatically, but may not be worth the trouble.

yanchith (changes requested):
All around nice! Can be merged promptly once we fix the memory leak.

src/main.rs
yanchith: My bad. Didn't notice it 🙏

yanchith (approved):

PR #12, opened 2019-08-28
Refactor rendering system

yanchith:
This commit presents 2 major changes:

1) Viewport (now renamed to Scene) and Imgui renderers are no longer
visible. They are instead hidden in the renderer module, which only
exports a single `Renderer` facade.

2) Drawing is done via the `Renderer.begin_render()` method and the
`Render` struct, which delegate work to Viewport/Scene and Imgui
renderers.

Feature-wise, the refactor allows us to have MSAA for imgui. Also, the
renderers can now more easily share resources internally where
applicable:

- The renderer facade now manages the swapchain, the depth texture,
the MSAA buffer, etc.

- Command encoders are still re-used across both renderers, having in
a single submission for the entire gui and scene, but now the usage
code doesn't need to know anything about it.

- Although I originally wanted to factor out GPU resource management
from the subrenderers to the renderer facade, due to a limitation in
`imgui::TextureId` (there is no way not to use it when writing a
renderer for imgui-rs), the texture resource management needs to be
internal to `ImguiRenderer` and cannot be extracted out. For symmetry
reasons, I left geometry management internal to `SceneRenderer`.

- I also created a swamp file for common renderer functionality:
`rendrer/common.rs` (commoners!)

### DIFF
--- a/src/renderer/mod.rs
+++ b/src/renderer/mod.rs
@@ -22,6 +22,8 @@ pub struct RendererOptions {
     pub msaa: Msaa,
 }
 
+/// Multi-sampling setting. Can be either disabled (1 sample per
+/// pixel), or 4/8/16 samples per pixel.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum Msaa {
     Disabled,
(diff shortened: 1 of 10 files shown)


src/renderer/mod.rs
```diff
@@ -0,0 +1,357 @@
...
+            encoder: Some(encoder),
+            msaa_attachment: self.msaa_texture_view.as_ref(),
+            depth_attachment: &self.depth_texture_view,
+            scene_renderer: &self.scene_renderer,
+            imgui_renderer: &self.imgui_renderer,
+        }
+    }
+}
+
+pub struct Render<'a> {
```
ondrowan: I'd prefer if this was a noun. Could it be `Frame` or something along that way?

src/renderer/mod.rs
```diff
@@ -0,0 +1,357 @@
+use nalgebra::base::Matrix4;
```
ondrowan, line 1: This file is missing docstrings on some structs / methods / functions.

ondrowan (changes requested):

ondrowan (approved):

PR #13, opened 2019-08-29
Importer worker thread

ondrowan:
This patch extracts `Importer` into a separate thread, so it doesn't block main thread while importing. The only downside is that it processes just one message at a time, so if there's a long running import, a terminate message still wait for it before it cleans up the thread. As discussed IRL, we'll keep it as is for now since all imports are very fast in release binary. This will be fixed later.

### DIFF
--- a/src/importer.rs
+++ b/src/importer.rs
@@ -150,7 +150,7 @@ impl Importer {
     }
 }
 
-enum ImporterWorkerMessage {
+enum ImporterWorkerRequest {
     NewImport(String),
     Terminate,
 }
@@ -163,12 +163,12 @@ enum ImporterWorkerMessage {
 pub struct ImporterWorker {
     thread: Option<thread::JoinHandle<()>>,
     response_rx: crossbeam_channel::Receiver<ImporterResult>,
-    request_tx: crossbeam_channel::Sender<ImporterWorkerMessage>,
+    request_tx: crossbeam_channel::Sender<ImporterWorkerRequest>,
 }
 
 impl ImporterWorker {
     /// Prepares communication channels and spawns a thread listening for
-    /// `ImporterWorkerMessage`.
+    /// `ImporterWorkerRequest`.
     pub fn new() -> Self {
         let (request_tx, request_rx) = unbounded();
         let (response_tx, response_rx) = unbounded();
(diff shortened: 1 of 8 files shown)


src/math.rs
```diff
@@ -0,0 +1,28 @@
+/// Linear interpolation between values `a` and `b` for parameter `f`.
+fn lerp(a: f32, b: f32, f: f32) -> f32 {
```
yanchith: Make `pub` or move down below.

src/importer.rs
```diff
@@ -61,6 +64,8 @@ pub struct FileMetadata {
     last_modified: std::time::SystemTime,
 }
 
+type ImporterResult = Result<Vec<Model>, ImporterError>;
```
yanchith: Shouldn't this be `pub`?

src/importer.rs
```diff
@@ -145,6 +150,101 @@ impl Importer {
...
+}
+
+/// Single threaded worker encapsulating obj importer.
+///
+/// Only one message (import or termination request) is processed at a time, so
+/// if there's a long running import, this thread can still block termination
+/// of the application.
+pub struct ImporterWorker {
+    thread: Option<thread::JoinHandle<()>>,
+    response_rx: crossbeam_channel::Receiver<ImporterResult>,
```
yanchith: Let's call these `response_receiver` and `request_sender`

src/importer.rs
```diff
@@ -145,6 +150,101 @@ impl Importer {
     }
 }
 
+enum ImporterWorkerRequest {
```
yanchith, line 153: Since this isn't `pub`, let's move it down.

src/importer.rs
```diff
@@ -145,6 +150,101 @@ impl Importer {
...
+            .send(ImporterWorkerRequest::NewImport(filename.to_string()))
+            .expect("Failed to send new import message to improter worker");
+    }
+
+    /// Reads response queue and returns last parsed obj in case there's any.
+    pub fn parsed_obj(&self) -> Option<ImporterResult> {
+        if self.response_rx.is_empty() {
+            None
+        } else {
+            let obj_importer_result = self
```
yanchith, line 217: Could possibly be a little shorter if `try_recv` was used.

yanchith (dismissed):
All good! Got some minor stylistic requests!

yanchith (approved):

PR #15, opened 2019-09-01
Use cubic bezier curve to interp camera-visible sphere over time

yanchith:
This adds the following:

1) A `Camera::visible_sphere()` method to compute the sphere currently
visible. This is the same sphere that would be passed to
`Camera::zoom_to_fit_visible_sphere()`

2) `math::CubicBezierEasing` that can smoothly interpolate a floating
point number between 0 and 1.

3) Glue code in `main.rs` that, when triggered, computes the currently
camera-visible sphere and the scene's bounding sphere and sets up and
interpolation that will slowly change one into the other over the
course of `CAMERA_INTERPOLATION_DURATION`. Bezier easing is used to
smoothly change the `weight` parameter of the linear interpolation.


yanchith:
@ondrowan you might want to turn off whitespace changes when reviewing, a lot of CRLF -> LF shows up in the diff.

ondrowan (approved):

PR #16, opened 2019-09-02
UI refactoring and styling

ondrowan:
First part is refactoring of UI representation. Since we need to hold also fonts, I've created `Ui` structure to hold all UI/imgui related settings. `Ui` produces `UiFrame` for each frame. This structure is used to draw actual UI elements for each frame.

Second part is styling of model list. I'm sure many of these things could be done in a more idiomatic way, but since it's temporary and we're learning, I'd refer fixing them for some future PR.


src/main.rs
```diff
@@ -146,7 +148,7 @@ fn main() {
         let duration_last_frame_s = duration_last_frame.as_secs() as f32
             + duration_last_frame.subsec_nanos() as f32 / 1_000_000_000.0;
 
-        imgui_context.io_mut().delta_time = duration_last_frame_s;
+        ui.imgui_context().io_mut().delta_time = duration_last_frame_s;
```
yanchith: This could use a setter later.

src/main.rs
```diff
@@ -160,22 +162,18 @@ fn main() {
...
-        let imgui_io = imgui_ui.io();
+        let ui_frame = ui.prepare_frame();
 
         input_manager.start_frame();
 
         // Imgui's IO is updated after current frame starts, else it'd contain
         // outdated values.
-        let ui_captured_keyboard = imgui_io.want_capture_keyboard;
-        let ui_captured_mouse = imgui_io.want_capture_mouse;
+        let ui_captured_keyboard = ui_frame.io().want_capture_keyboard;
```
yanchith: These two could use a getter.

src/main.rs
```diff
@@ -272,22 +270,25 @@ fn main() {
...
+            selected_model = clicked_model;
             is_importing = true;
             import_progress = 0.0;
         }
 
-        imgui_winit_platform.prepare_render(&imgui_ui, &window);
-        let imgui_draw_data = imgui_ui.render();
+        let imgui_draw_data = ui_frame.render();
+
+        // imgui_winit_platform.prepare_render(&imgui_ui, &window);
```
yanchith: Leftover comment?

yanchith (dismissed):
Good enough. Might need a bit of cleanup later, but I don't want to block it now.

yanchith (approved):
Looks good!

PR #17, opened 2019-09-02
Center models when imported

ondrowan:
Since I was extracting initialization code into `new`, I have also moved other bits of functionality into `CameraInterpolation` struct.

Also don't get fooled like me. You have to test this in release mode since in debug mode there's significant lag at the end of import and the interpolation takes just couple of frames to finish (at least on my desktop machine).


src/main.rs
```diff
@@ -350,6 +339,37 @@ struct CameraInterpolation {
...
+        CameraInterpolation {
+            source_origin,
+            source_radius,
+            target_origin,
+            target_radius,
+            target_time: time + CAMERA_INTERPOLATION_DURATION,
+        }
+    }
+
+    fn duration_left(&self, time: Instant) -> f32 {
```
yanchith: Maybe have just a single function instead of these 3. Usage:

```rust
let (sphere_origin, sphere_radius) = interp.update(time);
camera.zoom_to_fit_visible_sphere(sphere_origin, sphere_radius);
```

yanchith (dismissed):
Good enough for now.

src/main.rs
ondrowan: Done. It also takes easing as parameter now. Later on we could make this function generic and pass some `Easing` trait if we ever need more different types / configurations.

yanchith (approved):

PR #19, opened 2019-09-03
A draft PR adding touch very basic touch handling

yanchith:
@ondrowan I will fill in a better description later and *will* force push here, but please have a preliminary look.

This removes `imgui_winit_support` and replaces it with our own`imgui_input.rs`. A very hacky and pretty terrible touch handling is added on top.

I added 2 things:

- Mouse is disabled for N milliseconds after last touch event is received (not sure this is still needed. I will test)
- Touch events are added. They emulate clicks by setting imgui's mouse position and mouse downs.


ondrowan:
@yanchith I think this should be merged into `hbd` branch only. If it's the case, let's just get it to "it's working" state and throw it away later.

I checked the code, but since I have no way to test it, I have no idea what half of it means and if it really works.

PR #20, opened 2019-09-10
Update wgpu to newest master and winit to 0.20.0-alpha3

yanchith:
... and handle the aftermath.

Most interesting changes are:

- We can now select a rendering backend in runtime. Theoretically
Vulkan on macOS should work via MoltenVK, ~but it doesn't (see the
FIXME)~

- The structure of the main loop changed to accomodate
`EventLoop::run()` craziness

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7,7 +7,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.4"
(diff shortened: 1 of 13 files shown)


yanchith:
@ondrowan from the nature of the change, the diff of the main loop in `main.rs` is basically unreadable. You have to trust me - I changed nothing, except I moved most of the loop to handle `Event::EventsCleared`.

ondrowan (approved):

PR #23, opened 2019-09-13
Librarification2

yanchith:
This one seems less overthunk. Just moved things from `main.rs` to `lib.rs`, slapped some `allow(dead_code)`s in there, pulled a couple of settings out to env vars and documented things.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,11 +22,13 @@ We use standard `cargo` workflows:
 - `cargo build` to build,
 - `cargo run` to run.
 
-A gpu backend is automatically selected, but optionally a non-default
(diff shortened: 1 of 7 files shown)


yanchith:
Can squash the commits once we are done with it.

ondrowan (dismissed):
I don't really want to be a buzzkill, but I wouldn't use `rty` as module alias or env variable prefix. It's our internal project name and it's confusing for readers of the code that aren't familiar with it. Let's use `hs` instead.

I'm preapproving it though.

Also, please rebase some of the commits before merging.

PR #25, opened 2019-09-21
Add UV Sphere geometry primitive and ability to compute normals

yanchith:
This PR adds the UV Sphere geometry primitive and makes normals mandatory in geometry. The UV Sphere will be required later, most notably for the "covering sphere" operation I am implementing, but we might also use it in the renderer to produce skydomes/skyspheres.

Making normals mandatory gets rid of a few corner-cases we had to otherwise assert here and there.

If no normals are not present in the source data, there is `Geometry::from_triangle_faces_with_vertices_and_computed_normals` which can compute them.

### DIFF
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -2,7 +2,13 @@ use nalgebra as na;
 use nalgebra::base::Vector3;
 use nalgebra::geometry::Point3;
 
-use crate::convert::cast_u32;
+use crate::convert::{cast_u32, cast_usize};
+
+#[derive(Debug, Clone, Copy)]
+pub enum NormalStrategy {
+    Sharp,
+    // FIXME: add `Smooth`
+}
 
 /// Geometric data containing multiple possibly _variable-length_
 /// lists of geometric data, such as vertices and normals, and faces -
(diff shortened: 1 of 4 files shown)


src/geometry.rs
```diff
@@ -20,41 +26,67 @@ use crate::convert::cast_u32;
...
         Self {
-            faces: faces.into_iter().map(Face::Triangle).collect(),
+            faces: faces
+                .into_iter()
+                .enumerate()
+                .map(|(i, (i1, i2, i3))| {
+                    let normal_index = cast_u32(i);
+                    TriangleFace::new_separate(i1, i2, i3, normal_index, normal_index, normal_index)
+                })
+                .map(Face::from)
```
ondrowan, line 86: Doesn't this iterate over the vector again? Couldn't this be included in previous `map` instead?

src/geometry.rs
```diff
@@ -392,28 +458,117 @@ pub fn uv_cube_var_len(position: [f32; 3], scale: f32) -> Geometry {
...
+pub fn uv_sphere(position: [f32; 3], scale: f32, n_parallels: u32, n_meridians: u32) -> Geometry {
+    assert!(n_parallels >= 2, "Need at least 2 paralells");
+    assert!(n_meridians >= 3, "Need at least 3 meridians");
+
+    // Add the poles
+    let lat_line_max = n_parallels + 2;
+    // Add the last, wrapping meridian
+    let lng_line_max = n_meridians + 1;
+
+    use std::f32::consts::PI;
```
ondrowan, line 498: I'd move this `use` to the top since this function is going to be compiled every time anyway.

src/geometry.rs
```diff
@@ -460,25 +615,18 @@ fn n(x: f32, y: f32, z: f32) -> Vector3<f32> {
...
-    }
+    Vector3::cross(&u, &v)
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
 
-    fn quad() -> (Vec<TriangleFace>, Vec<Point3<f32>>) {
+    fn quad() -> (Vec<(u32, u32, u32)>, Vec<Point3<f32>>) {
```
ondrowan, line 629: Clippy actually complains about complexity of this type, but since it's in tests it probably doesn't matter that much.

ondrowan (changes requested):
Good job overall. There are just couple of notes you could go through before merging this.

src/geometry.rs
yanchith: Yeah, it's purely anonymous test code

src/geometry.rs
yanchith: The reason it's local is that there is also a `std::f64::consts::PI`... I wanted to use the short name `PI`, but not force the entire module to use `f32::consts::PI`. I can hoist the declaration of `std::f32`, but locally, I would still declare a `const PI: f32 = f32::consts::PI` to have a shorter alias...

src/geometry.rs
yanchith: It doesn't. Iteration only happens once, when a terminal operation (like `collect()`) is chained on the iterator. An iterator performs a single loop, and applies all operations in sequence for each element.

If you crank up the opt level, even the function calls disappear: https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(j:1,lang:rust,source:'pub+fn+use_iter(num:+i32)+-%3E+i32+%7B%0A++++let+data+%3D+%5Bnum%3B+1000%5D%3B%0A++++data.iter()%0A++++++++.map(%7Cx%7C+x+%2B+1)%0A++++++++.map(%7Cx%7C+x+%2B+1)%0A++++++++.sum()%0A%7D%0A'),l:'5',n:'0',o:'Rust+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:nightly,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'1',trim:'1'),lang:rust,libs:!(),options:'-C+opt-level%3D1',source:1),l:'5',n:'0',o:'rustc+nightly+(Editor+%231,+Compiler+%231)+Rust',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4

src/geometry.rs
yanchith: Let me know what you think

ondrowan (approved):

PR #26, opened 2019-09-24
Revamp Azure Pipelines configs to support nightly build

ondrowan:
There are now 2 entry points with different triggers. One for "PR and
master" builds, other for nightly release build. Since there's no
azure-pipelines.yml in root of the project both of those pipelines will
have to be manually set in UI.

Since `cargo fmt` checks all code regardless of debug/release mode, it's
omitted from nightly build.


ondrowan:
Just a note, there's one dead code warning for release mode (FPS counter) and I'm leaving it there just to test this build later tonight.

yanchith (approved):
Looks good!

PR #27, opened 2019-09-25
Shrink wrap

yanchith:
This adds the shrink wrap operation in the iterative style, as we discussed in person.

Please only have a look at `shrink_wrap.rs` and provide comments, other changes are just development support atm... the question is what
should we do with the changes to `lib.rs`, `input.rs`... I can throw them away, but will likely end up re-implementing them rather soon for the next operation. Unless we decide it is time to do the interpreter?

### DIFF
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -164,6 +164,10 @@ impl Geometry {
         &self.vertices
     }
 
+    pub fn vertices_mut(&mut self) -> &mut [Point3<f32>] {
+        &mut self.vertices
+    }
+
     pub fn normals(&self) -> &[Vector3<f32>] {
         &self.normals
     }
@@ -638,6 +642,27 @@ pub fn compute_centroid(geometries: &[Geometry]) -> Point3<f32> {
     }
 
     centroid / (vertex_count as f32)
+}
+
+pub fn find_closest_point(position: &Point3<f32>, geometry: &Geometry) -> Option<Point3<f32>> {
+    let vertices = geometry.vertices();
+    if vertices.is_empty() {
+        return None;
+    }
+
+    let mut closest = vertices[0];
+    // FIXME: @Optimization benchmark `distance` vs `distance_squared`
+    // with branching (0..1, 1..inf)
+    let mut closest_distance = na::distance(position, &closest);
+    for point in &vertices[1..] {
+        let distance = na::distance(position, &point);
+        if distance < closest_distance {
+            closest = *point;
+            closest_distance = distance;
+        }
+    }
+
+    Some(closest)
 }
 
 fn v(x: f32, y: f32, z: f32, translation: [f32; 3], scale: f32) -> Point3<f32> {
--- a/src/operations/shrink_wrap.rs
+++ b/src/operations/shrink_wrap.rs
@@ -1,0 +1,92 @@
+use std::slice;
+use std::u32;
+
+use crate::convert::{cast_u32, cast_usize};
+
+use crate::geometry::{self, Geometry};
+
+pub struct ShrinkWrapParams {
+    pub geometry: Geometry,
+    pub sphere_density: u32,
+    pub step: u32,
+}
+
+struct ShrinkWrapState {
+    sphere_geometry: Geometry,
+    current_step: u32,
+}
+
+pub struct ShrinkWrapOp {
+    params: ShrinkWrapParams,
+    state: Option<ShrinkWrapState>,
+}
+
+impl ShrinkWrapOp {
+    pub fn new(params: ShrinkWrapParams) -> Self {
+        Self {
+            params,
+            state: None,
+        }
+    }
+
+    pub fn next_value(&mut self) -> Option<Geometry> {
+        // Note that the operation does a nontrivial amount of work
+        // for its initialization. To make things more flexible for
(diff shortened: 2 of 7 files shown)


yanchith:
I removed the editor integration. If anyone wants to see the integrated version, checkout this commit https://github.com/sub-digital/HURBAN-Selector/pull/27/commits/da534118843699dfcd7b385bd7d75a7ca19043c5

benches/shrink_wrap_bench.rs
```diff
@@ -0,0 +1,57 @@
...
+                b.iter(|| {
+                    let mut op = ShrinkWrapOp::new(ShrinkWrapParams {
+                        geometry: geometry.clone(),
+                        sphere_density: black_box(*density),
+                        step: 0,
+                    });
+
+                    let mut res = None;
+                    while let Some(value) = op.next_value() {
+                        res = Some(value);
```
yanchith: Ooops, I totally forgot to perform the geometry conversion here. Interesting how much overhead there was even when we only cloned the geometry in an operation to produce multiple results.

yanchith: Added one more commit to measure both overheads.

benches/shrink_wrap_bench.rs
```diff
@@ -0,0 +1,98 @@
+use hurban_selector::geometry;
```
ondrowan: Does this file need to have the `_bench` suffix? It's pretty obvious since it's in `benches` directory. I'd follow the same standard as for tests.

src/geometry.rs
```diff
@@ -163,9 +163,17 @@ impl Geometry {
...
 
+    pub fn vertices_mut(&mut self) -> &mut [Point3<f32>] {
+        &mut self.vertices
+    }
+
     pub fn normals(&self) -> &[Vector3<f32>] {
         &self.normals
     }
+
+    pub fn normals_mut(&mut self) -> &mut [Vector3<f32>] {
```
ondrowan: This isn't used anywhere.

src/operations/shrink_wrap.rs
```diff
@@ -0,0 +1,80 @@
...
+            state: None,
+        }
+    }
+
+    pub fn next_value(&mut self) -> Option<Geometry> {
+        if self.state.is_none() {
+            self.init();
+        }
+
+        let state = self.state.as_mut().unwrap();
```
ondrowan: `unwrap` -> `expect`

src/operations/shrink_wrap.rs
```diff
@@ -0,0 +1,80 @@
...
+        Self {
+            params,
+            state: None,
+        }
+    }
+
+    pub fn next_value(&mut self) -> Option<Geometry> {
+        if self.state.is_none() {
+            self.init();
+        }
```
ondrowan, line 44: Why is `init` called on first iteration and not in `new`?

benches/shrink_wrap_bench.rs
```diff
@@ -0,0 +1,98 @@
...
+
+use criterion::{black_box, criterion_group, criterion_main, BenchmarkId, Criterion};
+
+fn bench_shrink_wrap(c: &mut Criterion) {
+    let mut group = c.benchmark_group("shrink_wrap");
+    let geometry = geometry::uv_sphere([0.0, 0.0, 0.0], 1.0, 3, 3);
+
+    for density in (5u32..=25u32).step_by(5) {
+        group.bench_with_input(
+            BenchmarkId::new("Batch + Convert", density.pow(2)),
```
ondrowan: What do these `.pow(2)`s mean?

ondrowan (changes requested):
Criterion looks very nice! We could maybe later include it as a part of our nightly builds when there are more operations to test.

There's couple of questions and suggestions for you before we can merge this.

src/geometry.rs
yanchith: Okay, I can remove it for now and re-add later

src/operations/shrink_wrap.rs
yanchith: A very good question! It will be what we decide it is, but my reasoning for doing it this way is that the interpreter might want to spawn the operation but not yet make it do any work (at least I believe this will be a real use-case  currently). That means creating an operation should be cheap, but this operation does a lot of work in the `init()` method. Thoughts?

benches/shrink_wrap_bench.rs
yanchith: The sphere we shrink wrap around the original geometry will have roughly `n^2` vertices. As I am thinking about it now, maybe we shouldn't really express that here and instead just display what the parameter was. I'll remove the pow calls.

src/operations/shrink_wrap.rs
ondrowan: I cannot imagine doing this for now without more info and examples. But it's not a biggie, it can stay like this, maybe with some extra comment to prevent confusion in the future.

yanchith: Okay, I'll document the current thinking.

ondrowan (dismissed):

yanchith:
OK, I'll squash and merge.

ondrowan (approved):

PR #28, opened 2019-09-27
first attempt

janper:


Cargo.lock
```diff
@@ -2,7 +2,7 @@
 # It is not intended for manual editing.
```
yanchith: I see no changes in the `cargo.toml`, did you update this file on purpose? If not, please revert it to original. 

Accidental update might have been caused by `cargo update`. The only time you want to update this file on purpose without also modifying `cargo.toml` is to update to a newer dependency version because of a fix the new version provides. Personally, I would also update the toml file even in this case.

src/geometry.rs
```diff
@@ -209,6 +219,15 @@ impl TriangleFace {
             normals: (ni1, ni2, ni3),
         }
     }
+
+    /// Generates 3 edges from the respective triangular face
+    pub fn to_edges(&self) -> Vec<Edge> {
```
yanchith: Instead of `Vec<Edge>` I would make the return type an `[Edge; 3]`. You prevent a heap allocation and get more static checks for free.

src/mesh-analysis.rs
```diff
@@ -0,0 +1,17 @@
+use nalgebra as na;
+use nalgebra::base::Vector3;
+use nalgebra::geometry::Point3;
+use std::collections::HashSet;
```
yanchith: We usually split imports into groups as follows:

1) `pub` imports (none here) 
2) std imports, if any (`std::collections::HashSet` in this case)
3) external crate imports, if any (all `nalgebra` imports here)
4) `crate` imports, if any (`crate::geometry::{Geometry, NormalStrategy, TriangleFace}` import here)
5) `super` imports

... and separate the groups by a newline. This way, the most local import group is last and the most global is first. Concretely in this case, it would be:

```rust

use nalgebra as na;
use nalgebra::base::Vector3;

use crate::geometry::{Geometry, NormalStrategy, TriangleFace};
```

src/mesh-analysis.rs
```diff
@@ -0,0 +1,17 @@
+use nalgebra as na;
```
yanchith: This file is not part of the compilation and in fact contains multiple compilation errors (from what I see). In rust, not all files are compiled, but only those referenced in the project by `mod` declarations.

There are multiple problems here:

1) The filename should be `mesh_analysis.rs`, not `mesh-analysis.rs`, because `-` is not a valid identifier character in rust and modules are not string tokens, but identifier tokens. 
2) You need to add a `mod mesh_analysis;` to `lib.rs` (search for `mod` in the code to find the section to put it in)

src/geometry.rs
```diff
@@ -166,6 +166,16 @@ impl Geometry {
...
     }
+
+    pub fn edges(&self) -> Vec<Edge> {
+        self.faces
+        .iter()
+        .flat_map(|face| 
+            match face {
+                Face::Triangle(triangle_face) => triangle_face.to_edges()
+            }
+        ).collect()
```
yanchith: This could just return a borrowed iterator. See `triangle_faces_iter()` and `triangle_faces_len()`.

yanchith (changes requested):
Thanks for the PR! ❤️ 

In the future, please write a short description of what the PR is trying to achieve (might not always be inferrable).

I see no huge problems with the features the PR adds, but have a couple of comments/questions:

- In `edges()`, the same edge will likely be returned multiple times (presumably exactly 2 times for watertight geometry). Are we conceptually distinguishing between `A -> B` and `B -> A` edges, or planning to?

- The CI is not passing for the PR (see the "checks" section below). You can click on the check that did not pass and view the error log in Azure to see what went wrong (spoiler alert, it is a formatting issue).

- Is this PR a part of a longer PR chain leading up to a feature? Are we sure we will be needing the code the PR adds? If so, please mention what that is in the description.

I also have more minor comments inline.

yanchith:
I also just realized, that there are two features implemented here. For the future, let's split it into multiple PRs, unless there is a dependency in between the features.

janper:
> * In `edges()`, the same edge will likely be returned multiple times (presumably exactly 2 times for watertight geometry). Are we conceptually distinguishing between `A -> B` and `B -> A` edges, or planning to?

that's the point of this feature! you can check is the mesh is two-manifold and watertight if you keep all the edges. sometimes they are also called halfedges and maybe we should call them that too.

janper:
> * The CI is not passing for the PR (see the "checks" section below). You can click on the check that did not pass and view the error log in Azure to see what went wrong (spoiler alert, it is a formatting issue).

my editor doesn't format the code automatically (can we fix it?)

janper:
> * Is this PR a part of a longer PR chain leading up to a feature? Are we sure we will be needing the code the PR adds? If so, please mention what that is in the description.

sure, they are my tasks from Trello

Cargo.lock
janper: yes, i used **cargo update** and the rest happened automatically. please let me know what to do to cause no trouble.

src/geometry.rs
janper: can you then `flat_map` `[Edge; 3]`?

src/mesh-analysis.rs
janper: understood

src/geometry.rs
ondrowan: Yes, `flat_map` is `Iterator` method, so you'll have to convert array into iterator using `.iter()` (or `.into_iter()`, although this case is a bit special since arrays themselves don't implement `IntoIterator` and are implicitly coerced into slice first, which itself returns iterator of references anyway, so it's the same as if you used just `.iter()`; [see the difference](https://stackoverflow.com/questions/34733811/what-is-the-difference-between-iter-and-into-iter)) first.

ondrowan:
@janper Since you use VSCode, I'd suggest to use [rust-analyzer](https://github.com/rust-analyzer/rust-analyzer). It doesn't install through official marketplace and you have to go through couple of steps manually, but it's not a big deal and it provides decent editor support. Check their readme and go the `cargo install-ra` way for installation of both server and VSCode extension. Once you install it, change setting `Rust-analyzer › Cargo-watch: Command` to `clippy` so you get clippy (linter) support in editor.

src/geometry.rs
janper: giving up...
```
pub fn edges(&self) -> Vec<Edge> {
        self.triangle_faces_iter()
            .flat_map(|face| face.to_edges().iter())
            .collect()
    }
```
console:
```
error[E0277]: a collection of type `std::vec::Vec<geometry::Edge>` cannot be built from an iterator over elements of type `&geometry::Edge`
   --> src\geometry.rs:173:14
    |
173 |             .collect()
    |              ^^^^^^^ a collection of type `std::vec::Vec<geometry::Edge>` cannot be built from `std::iter::Iterator<Item=&geometry::Edge>`
    |
    = help: the trait `std::iter::FromIterator<&geometry::Edge>` is not implemented for `std::vec::Vec<geometry::Edge>`
```

ondrowan: @janper 
```rust
pub fn edges<'a>(&'a self) -> impl Iterator<Item = Edge> + 'a {      
    self.faces.iter().copied().flat_map(|face| match face {
        Face::Triangle(triangle_face) => triangle_face.to_edges(),
    })
}
```

src/geometry.rs
```diff
@@ -217,6 +236,30 @@ impl From<(u32, u32, u32)> for TriangleFace {
...
+    pub vertices: (u32, u32),
+}
+
+impl Edge {
+    pub fn new(i1: u32, i2: u32) -> Self {
+        Edge {
+            vertices: (i1, i2),
+        }
+    }
+    pub fn equal_bidi(&self, &other: Edge) -> bool {
```
ondrowan: 2nd parameter should be `other: &Edge`.

Cargo.lock
yanchith: Just discard the change.

janper:
> @janper Since you use VSCode, I'd suggest to use [rust-analyzer](https://github.com/rust-analyzer/rust-analyzer). It doesn't install through official marketplace and you have to go through couple of steps manually, but it's not a big deal and it provides decent editor support. Check their readme and go the `cargo install-ra` way for installation of both server and VSCode extension. Once you install it, change setting `Rust-analyzer › Cargo-watch: Command` to `clippy` so you get clippy (linter) support in editor.

```
PS C:\Users\daroi\Documents\rustanalyzer\rust-analyzer> cargo install-ra
    Finished dev [unoptimized + debuginfo] target(s) in 0.97s
     Running `target\debug\ra_tools.exe install-ra`

will run: cargo install --path crates/ra_lsp_server --locked --force
  Installing ra_lsp_server v0.1.0 (C:\Users\daroi\Documents\rustanalyzer\rust-analyzer\crates\ra_lsp_server)
    Updating crates.io index
   Compiling ra_mbe v0.1.0 (C:\Users\daroi\Documents\rustanalyzer\rust-analyzer\crates\ra_mbe)
   Compiling ra_fmt v0.1.0 (C:\Users\daroi\Documents\rustanalyzer\rust-analyzer\crates\ra_fmt)
   Compiling ra_vfs v0.4.0
   Compiling rayon-core v1.6.0
error[E0658]: use of unstable library feature 'copy_within'
   --> C:\Users\daroi\.cargo\registry\src\github.com-1ecc6299db9ec823\ra_vfs-0.4.0\src\lib.rs:433:14
    |
433 |         tail.copy_within(gap_len..idx, 0);
    |              ^^^^^^^^^^^
    |
    = note: for more information, see https://github.com/rust-lang/rust/issues/54236

error: aborting due to previous error

For more information about this error, try `rustc --explain E0658`.
error: Could not compile `ra_vfs`.
warning: build failed, waiting for other jobs to finish...
error: failed to compile `ra_lsp_server v0.1.0 (C:\Users\daroi\Documents\rustanalyzer\rust-analyzer\crates\ra_lsp_server)`, intermediate artifacts can be found at `C:\Users\daroi\Documents\rustanalyzer\rust-analyzer\target`

Caused by:
  build failed
Error: StringError("`cargo install --path crates/ra_lsp_server --locked --force` exited with exit code: 101")
error: process didn't exit successfully: `target\debug\ra_tools.exe install-ra` (exit code: 1)
```

src/geometry.rs
janper:
```
error[E0277]: `[geometry::Edge; 3]` is not an iterator
   --> src\geometry.rs:177:36
    |
177 |         self.faces.iter().copied().flat_map(|face| match face {
    |                                    ^^^^^^^^ borrow the array with `&` or call `.iter()` on it to iterate over it
    |
    = help: the trait `std::iter::Iterator` is not implemented for `[geometry::Edge; 3]`
    = note: arrays are not iterators, but slices like the following are: `&[1, 2, 3]`
    = note: required because of the requirements on the impl of `std::iter::IntoIterator` for `[geometry::Edge; 3]`

error[E0277]: `[geometry::Edge; 3]` is not an iterator
   --> src\geometry.rs:176:35
    |
176 |     pub fn edges<'a>(&'a self) -> impl Iterator<Item = Edge> + 'a {
    |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ borrow the array with `&` or call `.iter()` on it to iterate over it
    |
    = help: the trait `std::iter::Iterator` is not implemented for `[geometry::Edge; 3]`
    = note: arrays are not iterators, but slices like the following are: `&[1, 2, 3]`
    = note: required because of the requirements on the impl of `std::iter::IntoIterator` for `[geometry::Edge; 3]`
    = note: the return type of a function must have a statically known size
```

yanchith (approved):

PR #29, opened 2019-09-28
Find orphans

janper:
Adding a standalone function `has_no_orphans` into `Geometry` where it should semantically belong.


ondrowan (changes requested):
There's couple of things I'd like to point out:

* In both constructors of `Geometry` there's a `FIXME: orphan removal`. Shouldn't this be part of construction of geometry?
* Whatever happens with this, it's a good candidate for a couple of unit tests.

janper:
> There's couple of things I'd like to point out:
> 
> * In both constructors of `Geometry` there's a `FIXME: orphan removal`. Shouldn't this be part of construction of geometry?

I'm on it but first I'm putting together analytic tools, then fixes

> * Whatever happens with this, it's a good candidate for a couple of unit tests.

I can only imagine tests using sample data. And THB, I'm trying to avoid that work.

yanchith:
>> Whatever happens with this, it's a good candidate for a couple of unit tests.

> I can only imagine tests using sample data. And THB, I'm trying to avoid that work.

I understand geometric ops are hard to test in general, but this one actually is really easy to create minimal, understandable test data for. Would be a shame not to do it 😉

janper:
requested tests done, ready for merging

src/geometry.rs
```diff
@@ -800,4 +822,46 @@ mod tests {
             normals.clone(),
         );
     }
+
+    #[test]
+    fn test_orphan_vertex_detection() {
```
yanchith: For future reference, it is better to cover one test case with one test function for better granularity. If this test fails, one does not know exactly why immediately and has to do more investigation to find out. Same for: `test_orphan_normal_detection`

yanchith:
Looks acceptable to me. @ondrowan, if you agree, I'll squash and merge. Or do you want to learn how to squash commits, @janper ? 😸

ondrowan:
I'd still split that test case and use a bit more proper naming of test methods: `test_{name_of_tested_method}_{what_is_being_tested}`. Eg. `test_has_no_orphan_vertices_returns_true_if_there_are_some` and `test_has_no_orphan_vertices_returns_false_if_there_are_none`. Or something along that way.

janper:
requested changes done

ondrowan (approved):

ondrowan:
@janper Squash the commits and 🚢 🇮🇹 !

PR #30, opened 2019-09-29
assert a valid face is being created

janper:
Asserting face validity in constructor. 
A valid face has to be constituted of three distinct vertices.


ondrowan (changes requested):
Tests fail because of this, so they need to be changed as well.

yanchith (dismissed):
A good check to have! Unfortunately the condition is exactly opposite to what it should be in the assert.

The following code is correct and produces no test errors:

```rust
            assert!(
                v.0 != v.1 && v.0 != v.2 && v.1 != v.2,
                "One or more face edges consists of the same vertex"
            );
```

janper:
> A good check to have! Unfortunately the condition is exactly opposite to what it should be in the assert.
> 
> The following code is correct and produces no test errors:
> 
> ```rust
>             assert!(
>                 v.0 != v.1 && v.0 != v.2 && v.1 != v.2,
>                 "One or more face edges consists of the same vertex"
>             );
> ```

oh... fixed!

yanchith:
Very nice! Could you also add a simple test that triggers this assert to solidify the idea?

janper:
passing tests

yanchith:
@janper after your change in https://github.com/sub-digital/HURBAN-Selector/pull/30/commits/1aef9673a5b86f24ccba713ebecdb8952405dcb5 only one of the added asserts is tested currently. Could you add a similar test to the other? (No need for the all the combinations, just something to prove the assert exists)

janper:
added missing tests

ondrowan:
I'd rename the tests to follow our naming convention. `test_traingle_face_new_etc`.

Windows test failed on downloading Rust for some reason, so I forced rebuild.

ondrowan:
One final comment before you can merge. You don't need to assign return value to variable in tests that are gonna panic anyway. Eg.:

```
    #[test]
    #[should_panic(expected = "One or more face edges consists of the same vertex")]
    fn test_triangle_face_new_with_invalid_vertex_indices_1_2_should_panic() {
        let _invalid_face_2 = TriangleFace::new(0, 2, 2);
    }
```

becomes:

```
    #[test]
    #[should_panic(expected = "One or more face edges consists of the same vertex")]
    fn test_triangle_face_new_with_invalid_vertex_indices_1_2_should_panic() {
        TriangleFace::new(0, 2, 2);
    }
```

janper:
requested changes have been made

ondrowan (approved):
Squash the commits and ship it!

PR #31, opened 2019-10-01
Edge valency check

janper:
This is a work-in-progress branch dealing with edge-related analytics.


src/geometry.rs
```diff
@@ -232,28 +248,79 @@ impl From<(u32, u32, u32)> for TriangleFace {
     }
 }
 
-/// A face edge. Contains indices to other geometry data - vertices
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub struct HalfEdge {
+/// Oriented face unoriented edge. Contains indices to other geometry data - vertices
```
yanchith: Possibly some copy-pasta in this comment, shouldn't it be "Oriented face edge"?

src/geometry.rs
```diff
@@ -232,28 +248,79 @@ impl From<(u32, u32, u32)> for TriangleFace {
...
     pub fn new(i1: u32, i2: u32) -> Self {
-        HalfEdge { vertices: (i1, i2) }
+        assert!(
+            i1 != i2,
+            "The oriented edge is constituted of the same vertex"
+        );
+        OrientedEdge { vertices: (i1, i2) }
     }
-    pub fn equal_bidi(self, other: HalfEdge) -> bool {
+    pub fn equal_bidi(self, other: OrientedEdge) -> bool {
```
yanchith: This still needed if you can just get the unoriented edge view?

src/edge_analysis.rs
```diff
@@ -0,0 +1,39 @@
...
+                let one_way_valencies = match edge_valency_map.get(edge) {
+                    Some(v) => *v + 1,
+                    None => 1,
+                };
+                let other_edge = Edge::Oriented(oriented_edge.reverted());
+                let other_way_valencies = match edge_valency_map.get(&other_edge) {
+                    Some(v) => *v + 1,
+                    None => 0,
+                };
+                let valencies = cmp::max(one_way_valencies, other_way_valencies);
```
yanchith: With the current strategy for oriented edges in this method, shouldn't the counts be always be equal on this line? So instead of doing  `cmp::max` you could just `assert_eq!`?

src/mesh_analysis.rs
```diff
@@ -13,3 +13,66 @@ pub fn has_no_orphans(geo: &Geometry) -> bool {
...
+/// Finds non-manifold (errorneous) edges in a mesh edge collection
+#[allow(dead_code)]
+pub fn non_manifold_edges(edge_valencies: HashMap<Edge, u32>) -> Vec<Edge> {
+    let keys = edge_valencies.keys();
+    keys.filter(|key| edge_valencies[key] > 2)
+        .copied()
+        .collect()
+}
+
+/// Finds manifold (inner) edges in a mesh edge collection
```
yanchith: Copy pasta comment?

src/mesh_analysis.rs
```diff
@@ -13,3 +13,66 @@ pub fn has_no_orphans(geo: &Geometry) -> bool {
...
+/// An edge is manifold when its valency is 2
+#[allow(dead_code)]
+pub fn is_mesh_orientable(edges: &[Edge]) -> bool {
```
yanchith: Hmmm, would probably pass in just oriented edges to make the logic simpler. For unoriented the answer is trivially no.

yanchith (commented):
Just as we talked offline, I am only reviewing the logic for this pass. We'll make another pass for the rust minutia later.

I assume the purpose of oriented edges is being able to later compute whether a mesh faces outwards?

You have my logic geometry comments inline, but I also have one over-arching point:

`Geometry` currently always produces either `OrientedEdge`s or `UnorientedEdge`s via the two iterator methods. Is `Edge` necessary? Do we ever expect to be operating on both at the same time? Note that except for wrapping them manually, there is currently no way to even get `Edge`s from `Geometry` (can be remedied with a `From` impl).

PR #32, opened 2019-10-01
Refactor cache and its usage in importer

ondrowan:
Cache was previously part of importer and it was very hard to test it. It's now extracted into a trait and passed into importer. This way it can be swapped for whatever caching mechanism we choose for different scenarios.

One thing is missing in this PR and that's actual (integration) tests if cache is called when loading files. I actually have the code and I'll leave it here as a comment for the future, but I couldn't manage to get the mocking work for `cfg(test)` only. I'm sure I just misunderstood something and I've asked the maintainer of `mockall` for help, but to move on with this refactoring, I decided to create a PR without this feature which can be added later.

```rust
#[test]
fn test_importer_import_obj_cache_sets_models_if_file_was_not_cached_before() {
    let mut cache = importer::MockCache::new();
    cache
        .expect_get_if_not_modified()
        .returning(|_, _| None)
        .times(1);
    cache.expect_get_by_checksum().returning(|_| None).times(1);
    cache.expect_set().returning(|_, _, _| ()).times(1);

    let mut importer = Importer::new(&mut cache);
    let path = "tests/fixtures/valid.obj";

    importer
        .import_obj(&path)
        .expect("Valid obj should be loaded");
}

#[test]
fn test_importer_import_obj_cache_returns_unmodified_file() {
    let mut cache = importer::MockCache::default();
    cache
        .expect_get_if_not_modified()
        .returning(|_, _| Some(vec![]))
        .times(1);
    cache.expect_get_by_checksum().returning(|_| None).times(0);
    cache.expect_set().returning(|_, _, _| ()).times(0);

    let mut importer = Importer::new(&mut cache);
    let path = "tests/fixtures/valid.obj";

    importer
        .import_obj(&path)
        .expect("Valid obj should be loaded");
}

#[test]
fn test_importer_import_obj_cache_returns_the_same_cached_data_from_different_file() {
    let mut cache = importer::MockCache::new();
    cache
        .expect_get_if_not_modified()
        .returning(|_, _| None)
        .times(1);
    cache
        .expect_get_by_checksum()
        .returning(|_| Some(vec![]))
        .times(1);
    cache.expect_set().returning(|_, _, _| ()).times(1);

    let mut importer = Importer::new(&mut cache);
    let path = "tests/fixtures/valid.obj";

    importer
        .import_obj(&path)
        .expect("Valid obj should be loaded");
}
```

Even though you don't like mocking, `mockall` is actually not very noisy and pretty elegant for what we need. It consists only of one attribute on our `Cache` trait and these special expects you can see in test cases above.


src/importer.rs
```diff
@@ -58,26 +58,81 @@ pub struct Model {
...
 
 pub type ImporterResult = Result<Vec<Model>, ImporterError>;
 
-/// `Importer` takes care of importing of obj files and caching of their
-/// internal representations. It holds paths to files, their metadata
-/// (checksums, timestamps) and models of parsed obj files.
+/// An interface for caching of obj files.
+///
+/// The source is expected to be file with path and FileMetadata.
+pub trait Cache {
```
yanchith: Nit: name this `ObjCache` or `ModelCache`?

src/importer.rs
```diff
@@ -58,26 +58,81 @@ pub struct Model {
...
+        self.loaded_models
+            .entry(metadata.checksum)
+            .or_insert_with(|| models.to_vec());
+    }
+}
+
+/// `Importer` takes care of importing of obj files and caching of their
+/// internal representations.
+pub struct Importer<'a, C> {
+    cache: &'a mut C,
```
yanchith: Hmm, any reason the `cache` is not owned by the importer?

src/importer.rs
```diff
@@ -89,65 +144,37 @@ impl Importer {
...
-        file.read_to_end(&mut file_contents)?;
-        let checksum = calculate_checksum(&file_contents);
-
-        let models = match self.loaded_models.entry(checksum) {
-            Entry::Occupied(loaded_model) => {
-                self.path_metadata.insert(
+        let models = match self.cache.get_if_not_modified(path, file_modified) {
+            Some(models) => return Ok(models),
+            None => {
+                let file_size = file.metadata().map(|m| m.len() as usize + 1).unwrap_or(0);
```
yanchith: Please document this +1, this is the third time my brain tripped over it 😂

src/importer.rs
```diff
@@ -89,65 +144,37 @@ impl Importer {
...
+                let mut file_contents = Vec::with_capacity(file_size);
+                file.read_to_end(&mut file_contents)?;
+                let checksum = calculate_checksum(&file_contents);
+
+                let models = match self.cache.get_by_checksum(checksum) {
+                    Some(models) => models.clone(),
```
yanchith, line 164: Hmmm, shouldn't we have a way to just update the `last_modified` for a path in the cache in case the checksum checks out? Something like: we validated the cache entry is still valid for this timestamp. We can leave it for a FIXME, but this could prevent an allocation and de-allocation in the cache itself when you set new data.

yanchith (commented):
Looks nice! Got a few thoughts inline. Also looking forward to see the mocking in action.

src/importer.rs
ondrowan: It used to be until I started to experiement with home-grown mocking and I wanted to see what happened inside of cache object. Now that we're using library for that, I'll see how it behaves, but I'm pretty sure those mocks will need to be passed by reference to be usable in tests.

src/importer.rs
ondrowan: I'm not sure I understand what you mean. Matching checksum does only mean there are contents of some file cached, but it doesn't mean it has come from particular path. That's why path and metadata are cached anyway on line 163.

yanchith: On line 163 you set the models you acquired here back to the cache with new file_modified time. That is an additional drop on line 163 happening because you overwrite the data in cache with the same data. You could instead just say to the cache:

`cache.update_last_modified(path, file_modified)`, causing the cache to update it's metadata and not drop the data in cache only to be replaced with the same data.

yanchith: My bad, just realized the `set` method does different things in different contexts. Carry on!

ondrowan:
It's ready for proper review now. All draft comments should be implemented. Mocked tests are also added.

src/importer.rs
```diff
@@ -58,26 +60,82 @@ pub struct Model {
...
-    last_modified: std::time::SystemTime,
+    last_modified: SystemTime,
 }
 
 pub type ImporterResult = Result<Vec<Model>, ImporterError>;
 
-/// `Importer` takes care of importing of obj files and caching of their
-/// internal representations. It holds paths to files, their metadata
-/// (checksums, timestamps) and models of parsed obj files.
+#[cfg_attr(test, automock)]
```
yanchith: Attributes are usually bellow doc comments.

src/importer.rs
```diff
@@ -416,4 +454,248 @@ mod tests {
...
+
+        assert!(loaded_models.is_none());
+    }
+
+    // The following tests with mocked cache are technically integration tests
+    // and they use fixture data. They're kept here to prevent complications
+    // with automocks not being present in debug build, as it is built without
+    // code marked as `cfg(test)`.
+    //
+    // FIXME: Ideal scenario would be if filesystem was mocked and passed into
```
yanchith, line 642: Oh yeah!

src/importer.rs
```diff
@@ -416,4 +454,248 @@ mod tests {
...
+        let mut cache = MockObjCache::new();
+        cache
+            .expect_get_if_not_modified()
+            .returning(|_, _| None)
+            .times(1);
+        cache
+            .expect_get_by_checksum()
+            .returning(|_| Some(vec![]))
+            .times(1);
+        cache.expect_set().returning(|_, _, _| ()).times(1);
```
yanchith: Is there a way to assert we are given the same data here as we produced in `get_by_checksum`? I believe `mockall` has a `.with()` thing that sounds like it may help.

yanchith (dismissed):
You and your mockery!

I have a couple of nits, but am pre-approving. The `cfg(test)` thingy is meh, but we'll live. The library API is actually pretty nice 🍵

src/importer.rs
ondrowan: Yes, there is and will try to do that. I thought this lazy attempt would be enough, but you're right that we should test that as well.

src/importer.rs
```diff
@@ -680,19 +680,45 @@ mod tests {
...
         cache
             .expect_get_by_checksum()
-            .returning(|_| Some(vec![]))
+            .returning(|_| Some(MODELS.to_vec()))
+            .times(1);
+        cache
+            .expect_set()
+            .with(
+                predicate::eq(path.to_string()),
+                predicate::always(),
```
yanchith: Should this `always` be here? Also, are these predicates AND-ed or OR-ed>

yanchith (dismissed):

src/importer.rs
ondrowan: These are 3 different parameters, so nothing is OR-ed or AND-ed here.

`always` was there just so we don't need to deal with file metadata, but it's probably better if we do this thoroughly. I've changed it.

yanchith (approved):

PR #34, opened 2019-10-07
Edges, edge analytics and edge-based geometry analytics

janper:
This branch implements most (if not all) of the edge-related analytics we'll need. All functions have been tested, some (marked as ~TODO~ or FIXME) will be extended later on when we're implementing respective edge-related modifiers.


src/geometry.rs
```diff
@@ -393,7 +439,7 @@ pub fn cube_smooth_same_len(position: [f32; 3], scale: f32) -> Geometry {
     Geometry::from_triangle_faces_with_vertices_and_normals(faces, vertex_positions, vertex_normals)
 }
 
-pub fn cube_sharp_same_len(position: [f32; 3], scale: f32) -> Geometry {
```
janper, line 404: I suggest removing this function because it doesn't generate a proper watertight mesh. it's being used in one of the tests but that can be replaced by some other function

src/geometry.rs
```diff
@@ -487,6 +533,81 @@ pub fn cube_sharp_same_len(position: [f32; 3], scale: f32) -> Geometry {
     Geometry::from_triangle_faces_with_vertices_and_normals(faces, vertex_positions, vertex_normals)
 }
 
+#[allow(dead_code)]
+pub fn cube_sharp_same_len_welded(position: [f32; 3], scale: f32) -> Geometry {
```
yanchith: This is definitely not `same_len` 😉

yanchith: What is the benefit of this vs `cube_sharp_var_len()`?

src/geometry.rs
yanchith: While I see that the mesh it generates probably makes no sense for geometry modeling, I would prefer to keep if for a while longer. We can add a FIXME to remove, but I want to double check the generated renderer geometry from `cube_sharp_var_len()` is equivalent to this one before we remove.

src/mesh_analysis.rs
```diff
@@ -0,0 +1,395 @@
...
+        border_vertices.insert(edge.vertices.0);
+        border_vertices.insert(edge.vertices.1);
+    });
+    border_vertices
+}
+
+/// Check if all the face normals point the same way.
+/// In a proper watertight mesh each oriented edge
+/// should have a counterpart in a reverted oriented edge
+// TODO: make this work for open meshes
```
yanchith: FYI: we use `TODO` to mark code we need to fix before opening a PR and `FIXME` for minor changes we need to make that are not worth creating an issue in trello for. So this is a actually a FIXME

src/geometry.rs
janper: it generates a broken geometry. please, check if the new one is identical (as far as i'm concerned, it is) and let's please remove it so that we don't use it for further tests.

src/geometry.rs
janper: > This is definitely not `same_len` 😉

it wasn't clear what `same_len` means in this context. i can rename it

janper: > What is the benefit of this vs `cube_sharp_var_len()`?

sharp normals, edgy shading. no benefits, just correct geometry

EDIT: yes, it's the same. I haven't noticed, sorry. we can remove the function I made and keep yours as it's actually nicer. I'd still like to remove the unwelded box from `cube_sharp_var_len()`.

src/mesh_analysis.rs
janper: ok

src/geometry.rs
janper: after a debate with @yanchith marked as `#[deprecated]`

src/edge_analysis.rs
```diff
@@ -0,0 +1,127 @@
...
+}
+
+pub type EdgeCountMap = HashMap<UnorientedEdge, EdgeCount>;
+
+/// Calculate edge valencies = number of faces sharing an edge
+/// 1 -> border edge = acceptable but mesh is not watertight
+/// 2 -> manifold edge = correct
+/// 3 or more -> non-manifold edge = corrupted mesh
+#[allow(dead_code)]
+pub fn edge_valencies(oriented_edges: &[OrientedEdge]) -> EdgeCountMap {
```
yanchith: From the usage in this PR it seems that a slice is unnecessarily restrictive as a parameter here. Instead, the parameter can be `I: IntoIterator<Item=&OrientedEdge>` so there is no need to allocate the the intermediate vector.

src/edge_analysis.rs
```diff
@@ -0,0 +1,127 @@
...
+        if let Some(edge_count) = edge_valency_map.get(&unoriented_edge) {
+            ascending_count += edge_count.ascending_count;
+            descending_count += edge_count.descending_count;
+        }
+
+        let edge_count = EdgeCount {
+            ascending_count,
+            descending_count,
+        };
+        edge_valency_map.insert(unoriented_edge, edge_count);
```
yanchith: Is this meant to overwrite the original value? In that case, a much more efficient approach would be:

```rust
        let unoriented_edge = UnorientedEdge(*edge);
        let mut ascending_count: u32 = 0;
        let mut descending_count: u32 = 0;
        if edge.vertices.0 < edge.vertices.1 {
            ascending_count += 1;
        } else {
            descending_count += 1;
        }
        let edge_count = edge_valency_map.entry(unoriented_edge).or_insert(EdgeCount {
            ascending_count: 0,
            descending_count: 0,
        });
        edge_count.ascending_count += ascending_count;
        edge_count.descending_count += descending_count;
```

src/edge_analysis.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    edge_valency_map
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::geometry::{Geometry, NormalStrategy};
+    use nalgebra::Point3;
```
yanchith, line 49: Please fix the import order.

src/edge_analysis.rs
```diff
@@ -0,0 +1,127 @@
...
+        Point3::new(
+            scale * x + translation[0],
+            scale * y + translation[1],
+            scale * z + translation[2],
+        )
+    }
+
+    fn quad() -> (Vec<(u32, u32, u32)>, Vec<Point3<f32>>) {
+        #[rustfmt::skip]
+            let vertices = vec![
```
yanchith: This is misformatted.

src/edge_analysis.rs
```diff
@@ -0,0 +1,127 @@
...
+            v(-1.0, -1.0,  0.0, [0.0, 0.0, 0.0], 1.0),
+            v( 1.0, -1.0,  0.0, [0.0, 0.0, 0.0], 1.0),
+            v( 1.0,  1.0,  0.0, [0.0, 0.0, 0.0], 1.0),
+            v(-1.0,  1.0,  0.0, [0.0, 0.0, 0.0], 1.0),
+        ];
+
+        #[rustfmt::skip]
+            let faces = vec![
```
yanchith: This is misformatted.

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+    for face in geo.triangle_faces_iter() {
+        used_vertices.insert(face.vertices.0);
+        used_vertices.insert(face.vertices.1);
+        used_vertices.insert(face.vertices.2);
+    }
+    used_vertices.len() == geo.vertices().len()
+}
+
+/// Finds edges with a certain valency in a mesh edge collection
+/// Valency indicated how many faces share the edge
```
yanchith: "indicated" -> "indicates"?

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+fn find_edges_with_valency<'a>(
+    edge_valencies: &'a EdgeCountMap,
+    valency: u32,
+) -> impl Iterator<Item = OrientedEdge> + 'a {
+    edge_valencies
+        .iter()
+        .filter(move |(_edge_wrapper, edge_count)| {
```
yanchith: `_edge_wrapper` -> `_`, unnecessary `move`

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+            edge_count.ascending_count + edge_count.descending_count == valency
+        })
+        .map(|(edge_wrapper, _value)| edge_wrapper.0)
```
yanchith: `edge_wrapper` -> `unoriented_edge`, `_value` -> `_`

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+}
+
+/// Finds non-manifold (errorneous) edges in a mesh edge collection
+#[allow(dead_code)]
+pub fn non_manifold_edges<'a>(
+    edge_valencies: &'a EdgeCountMap,
+) -> impl Iterator<Item = OrientedEdge> + 'a {
+    edge_valencies
+        .iter()
+        .filter(move |(_edge_wrapper, edge_count)| {
```
yanchith: `_edge_wrapper` -> `_`, unnecessary `move`

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+            edge_count.ascending_count + edge_count.descending_count > 2
+        })
+        .map(|(edge_wrapper, _value)| edge_wrapper.0)
```
yanchith: `edge_wrapper` -> `unoriented_edge`, `_value` -> `_`

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+    });
+    border_vertices
+}
+
+/// Check if all the face normals point the same way.
+/// In a proper watertight mesh each oriented edge
+/// should have a counterpart in a reverted oriented edge
+#[allow(dead_code)]
+pub fn is_mesh_orientable(edge_valencies: &EdgeCountMap) -> bool {
+    edge_valencies.iter().all(|(_edge_wrapper, edge_count)| {
```
yanchith: `_edge_wrapper` -> `_`

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+        border_vertices.insert(edge.vertices.1);
+    });
+    border_vertices
+}
+
+/// Check if all the face normals point the same way.
+/// In a proper watertight mesh each oriented edge
+/// should have a counterpart in a reverted oriented edge
+#[allow(dead_code)]
+pub fn is_mesh_orientable(edge_valencies: &EdgeCountMap) -> bool {
```
yanchith, line 84: Does the doc comment match the code? Is this intended to be used on watertight meshes only, or any mesh? Hmm, either I am missing something, or the condition: `edge_count.ascending_count <= 1 && edge_count.descending_count <= 1` is not correct.

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+    edge_valencies.iter().all(|(_edge_wrapper, edge_count)| {
+        edge_count.ascending_count <= 1 && edge_count.descending_count <= 1
+    })
+}
+
+/// The mesh is watertight if there is no border or non-manifold edge,
+/// in other words, all edge valencies are 2
+#[allow(dead_code)]
+pub fn is_mesh_watertight(edge_valencies: &EdgeCountMap) -> bool {
+    edge_valencies.iter().all(|(_edge_wrapper, edge_count)| {
```
yanchith: `_edge_wrapper` -> `_`

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+        edge_count.ascending_count + edge_count.descending_count == 2
```
yanchith: Shouldn't this be: `edge_count.ascending_count == 1 && edge_count.descending_count == 1`? (A stronger condition)

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+    })
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
yanchith: Please fix import order.

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+        )
+    }
+
+    fn n(x: f32, y: f32, z: f32) -> Vector3<f32> {
+        Vector3::new(x, y, z)
+    }
+
+    fn quad() -> (Vec<(u32, u32, u32)>, Vec<Point3<f32>>) {
+        #[rustfmt::skip]
+            let vertices = vec![
```
yanchith: Misformatted.

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+            v(-1.0, -1.0,  0.0, [0.0, 0.0, 0.0], 1.0),
+            v( 1.0, -1.0,  0.0, [0.0, 0.0, 0.0], 1.0),
+            v( 1.0,  1.0,  0.0, [0.0, 0.0, 0.0], 1.0),
+            v(-1.0,  1.0,  0.0, [0.0, 0.0, 0.0], 1.0),
+        ];
+
+        #[rustfmt::skip]
+            let faces = vec![
```
yanchith: Misformatted

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+    }
+
+    fn non_manifold_shape() -> (Vec<(u32, u32, u32)>, Vec<Point3<f32>>) {
+        #[rustfmt::skip]
+            let vertices = vec![
+            v(-1.0, -1.0,  0.0, [0.0, 0.0, 0.0], 1.0),
+            v( 1.0, -1.0,  0.0, [0.0, 0.0, 0.0], 1.0),
+            v( 1.0,  1.0,  0.0, [0.0, 0.0, 0.0], 1.0),
+            v(-1.0,  1.0,  0.0, [0.0, 0.0, 0.0], 1.0),
+            v(0.0,  0.0,  1.0, [0.0, 0.0, 0.0], 1.0),
```
yanchith: Misformatted

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+            (2, 3, 0),
+            (2, 4, 0),
+        ];
+
+        (faces, vertices)
+    }
+
+    fn cube_sharp_broken_winding(position: [f32; 3], scale: f32) -> Geometry {
+        #[rustfmt::skip]
+            let vertex_positions = vec![
```
yanchith: Misformatted

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+            let faces = vec![
+            (0, 1, 2),
+            (2, 3, 0),
+            (2, 4, 0),
+        ];
+
+        (faces, vertices)
+    }
+
+    fn cube_sharp_broken_winding(position: [f32; 3], scale: f32) -> Geometry {
```
yanchith: Broken how? Variable?

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+            n( 1.0,  0.0,  0.0),
+            // left
+            n(-1.0,  0.0,  0.0),
+            n(-1.0,  0.0,  0.0),
+            n(-1.0,  0.0,  0.0),
+            n(-1.0,  0.0,  0.0),
+        ];
+
+        #[rustfmt::skip]
+            let faces = vec![
```
yanchith: Misformatted

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+            v( 1.0,  1.0, -1.0, position, scale),
+            // front
+            v(-1.0, -1.0, -1.0, position, scale),
+            v( -1.0, -1.0, 1.0, position, scale),
+            v( 1.0, -1.0,  1.0, position, scale),
+            v(1.0, -1.0,  -1.0, position, scale),
+        ];
+
+        #[rustfmt::skip]
+            let vertex_normals = vec![
```
yanchith: Misformatted

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+        let oriented_edges_border_correct = vec![
+            OrientedEdge::new(0, 1),
+            OrientedEdge::new(1, 2),
+            OrientedEdge::new(2, 3),
+            OrientedEdge::new(3, 0),
+        ];
+
+        let oriented_edges_border_check = border_edges(&edge_valency_map);
+
+        for o_e in oriented_edges_border_check {
```
yanchith: If you iterate in this order, the test doesn't prove anything. If `border_edges()` returned `(0, 1)` 4 times, this test would still pass but the logic would be incorrect. Instead the `_correct` data should be in the outer loop.

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+            OrientedEdge::new(2, 3),
+            OrientedEdge::new(3, 0),
+        ];
+        let oriented_edges_with_valency_2_correct =
+            vec![OrientedEdge::new(2, 0), OrientedEdge::new(0, 2)];
+
+        let oriented_edges_with_valency_1 = find_edges_with_valency(&edge_valency_map, 1);
+        let oriented_edges_with_valency_2 = find_edges_with_valency(&edge_valency_map, 2);
+
+        for o_e in oriented_edges_with_valency_1 {
```
yanchith: If you iterate in this order, the test doesn't prove anything. If `find_edges_with_valency()` returned `(0, 1)` 4 times, this test would still pass but the logic would be incorrect. Instead the `_correct` data should be in the outer loop.

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+            assert!(oriented_edges_with_valency_1_correct
+                .iter()
+                .any(|e| *e == o_e));
+        }
+
+        for o_e in oriented_edges_with_valency_2 {
```
yanchith: If you iterate in this order, the test doesn't prove anything. If `find_edges_with_valency()` returned `(0, 2)` 2 times, this test would still pass but the logic would be incorrect. Instead the `_correct` data should be in the outer loop.

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+        );
+        let oriented_edges: Vec<OrientedEdge> = geometry.oriented_edges_iter().collect();
+        let edge_valency_map = edge_valencies(&oriented_edges);
+
+        let oriented_edges_manifold_correct =
+            vec![OrientedEdge::new(2, 0), OrientedEdge::new(0, 2)];
+
+        let oriented_edges_manifold_check = manifold_edges(&edge_valency_map);
+
+        for o_e in oriented_edges_manifold_check {
```
yanchith: Same as above: the correct data should be in the outer loop.

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+        let oriented_edges: Vec<OrientedEdge> = geometry.oriented_edges_iter().collect();
+        let edge_valency_map = edge_valencies(&oriented_edges);
+
+        let oriented_edges_non_manifold_correct =
+            vec![OrientedEdge::new(2, 0), OrientedEdge::new(0, 2)];
+
+        let oriented_edges_non_manifold_check: Vec<_> =
+            non_manifold_edges(&edge_valency_map).collect();
+
+        for o_e in &oriented_edges_non_manifold_check {
```
yanchith: Same as above: correct data should be in the outer loop.

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+            NormalStrategy::Sharp,
+        );
+        let oriented_edges: Vec<OrientedEdge> = geometry.oriented_edges_iter().collect();
+        let edge_valency_map = edge_valencies(&oriented_edges);
+
+        let border_vertex_indices_correct = vec![0, 1, 2, 3];
+
+        let order_vertex_indices_check = border_vertex_indices(&edge_valency_map);
+
+        for v_i in order_vertex_indices_check {
```
yanchith: Same as above: correct data should be in the outer loop.

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+            assert!(border_vertex_indices_correct.iter().any(|v| *v == v_i));
+        }
+    }
+
+    #[test]
+    fn test_mesh_analysis_is_mesh_orientable_returns_true() {
```
yanchith: `test_mesh_analysis_is_mesh_orientable_returns_true_watertight_mesh`

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+            NormalStrategy::Sharp,
+        );
+        let oriented_edges: Vec<OrientedEdge> = geometry.oriented_edges_iter().collect();
+        let edge_valency_map = edge_valencies(&oriented_edges);
+
+        assert!(is_mesh_orientable(&edge_valency_map));
+    }
+
+    #[test]
+    fn test_mesh_analysis_is_mesh_orientable_returns_false_because_is_wrong() {
```
yanchith: `test_mesh_analysis_is_mesh_orientable_returns_false_for_nonorientable_mesh`

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+        let geometry = cube_sharp_broken_winding([0.0, 0.0, 0.0], 1.0);
+
+        let oriented_edges: Vec<OrientedEdge> = geometry.oriented_edges_iter().collect();
+        let edge_valency_map = edge_valencies(&oriented_edges);
+
+        assert!(!is_mesh_orientable(&edge_valency_map));
+    }
+
+    #[test]
+    fn test_mesh_analysis_is_mesh_watertight_returns_true() {
```
yanchith: `test_mesh_analysis_is_mesh_watertight_returns_true_for_watertight_mesh`

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
...
+        let geometry = geometry::cube_sharp_var_len([0.0, 0.0, 0.0], 1.0);
+        let oriented_edges: Vec<OrientedEdge> = geometry.oriented_edges_iter().collect();
+        let edge_valency_map = edge_valencies(&oriented_edges);
+
+        assert!(is_mesh_watertight(&edge_valency_map));
+    }
+
+    #[test]
+    fn test_mesh_analysis_is_mesh_watertight_returns_false_because_is_open() {
```
yanchith: `test_mesh_analysis_is_mesh_watertight_returns_false_for_open_mesh`

src/geometry.rs
```diff
@@ -1,11 +1,14 @@
 use std::collections::HashSet;
 
 use arrayvec::ArrayVec;
+use std::cmp;
```
yanchith: Please fix import order

src/geometry.rs
```diff
@@ -174,9 +177,25 @@ impl Geometry {
...
+    pub fn unoriented_edges_iter<'a>(&'a self) -> impl Iterator<Item = UnorientedEdge> + 'a {
         self.triangle_faces_iter()
-            .flat_map(|face| ArrayVec::from(face.to_edges()).into_iter())
+            .flat_map(|face| ArrayVec::from(face.to_unoriented_edges()).into_iter())
+    }
+
+    /// Genus of a mesh is the number of holes in topology / conectivity
+    /// The mesh must be triangular and watertight
+    /// V - E + F = 2 (1 - G)
+    pub fn mesh_genus(&self, edges: &HashSet<UnorientedEdge>) -> i32 {
```
yanchith: This shouldn't be a method on `Geometry` if you need to provide external data, possibly coming from another mesh. Let's make it a function in `mesh_analysis.rs`. Also the fact it takes a hash set only to get it's `len` is worrisome.

src/geometry.rs
```diff
@@ -258,25 +286,47 @@ impl From<(u32, u32, u32)> for TriangleFace {
...
+        );
+        OrientedEdge { vertices: (i1, i2) }
+    }
+
+    pub fn is_reverted(self, other: OrientedEdge) -> bool {
+        self.vertices.0 == other.vertices.1 && self.vertices.1 == other.vertices.0
+    }
+}
+
+impl From<(u32, u32)> for OrientedEdge {
```
yanchith: Is this used? Let's not add all impls we can think of before they are used. They can be added later trivially.

src/geometry.rs
```diff
@@ -816,13 +2899,154 @@ mod tests {
...
+        let unoriented_edge_other_way = UnorientedEdge(OrientedEdge::new(1, 0));
+        let mut hasher_1 = DefaultHasher::new();
+        let mut hasher_2 = DefaultHasher::new();
+        unoriented_edge_one_way.hash(&mut hasher_1);
+        unoriented_edge_other_way.hash(&mut hasher_2);
+        assert_eq!(hasher_1.finish(), hasher_2.finish());
+    }
+
+    #[test]
+    fn test_unoriented_edge_hash_returns_false_because_different() {
```
yanchith: This test is incorrect. The hash can be the same even if data is different. It can start failing randomly when underlying implementation of `DefaultHasher` changes. Let's remove it.

src/geometry.rs
```diff
@@ -816,13 +2899,154 @@ mod tests {
...
+    }
+
+    #[test]
+    #[should_panic(expected = "The oriented edge is constituted of the same vertex")]
+    fn test_oriented_edge_constructor_consists_of_the_same_vertex_should_panic() {
+        OrientedEdge::new(0, 0);
+    }
+
+    #[test]
+    fn test_oriented_edge_constructor_doesnnt_consist_of_the_same_vertex_should_pass() {
```
yanchith: `does_not`

src/geometry.rs
```diff
@@ -754,6 +807,2036 @@ mod tests {
         (faces, vertices, normals)
     }
 
+    fn torus() -> (Vec<(u32, u32, u32)>, Vec<Point3<f32>>) {
```
yanchith: Please move all these data fixtures to `test_geometry_fixtures.rs` and add this text on top:

```
// FIXME: move to data fixture files once we are able to serialize and deserialize stuff.
```

src/mesh_analysis.rs
```diff
@@ -0,0 +1,413 @@
+use std::collections::HashSet;
+
+use crate::edge_analysis::EdgeCountMap;
+use crate::geometry::{Geometry, OrientedEdge};
+
+/// Check if all the vertices of geometry are referenced in geometry's faces
+#[allow(dead_code)]
+pub fn has_no_orphans(geo: &Geometry) -> bool {
```
yanchith: This is already implemented on `Geometry`, why have it here as well?

src/geometry.rs
```diff
@@ -864,6 +3088,71 @@ mod tests {
...
+            UnorientedEdge(OrientedEdge::new(1, 2)),
+            UnorientedEdge(OrientedEdge::new(2, 0)),
+            UnorientedEdge(OrientedEdge::new(2, 3)),
+            UnorientedEdge(OrientedEdge::new(3, 0)),
+            UnorientedEdge(OrientedEdge::new(0, 2)),
+        ];
+        let unoriented_edges_to_check: Vec<UnorientedEdge> =
+            geometry.unoriented_edges_iter().collect();
+
+        assert!(unoriented_edges_to_check
```
yanchith: This doesn't test much. Swap the inner loop with the outer loop.

src/geometry.rs
```diff
@@ -864,6 +3088,71 @@ mod tests {
...
+            .iter()
+            .all(|u_e| unoriented_edges_correct.iter().any(|e| e == u_e)));
+
+        let len_1 = unoriented_edges_to_check.len();
+        let len_2 = unoriented_edges_correct.len();
+        assert_eq!(
+            len_1, len_2,
+            "unoriented_edges_to_check.len() = {}, unoriented_edges_correct.len() = {}",
```
yanchith: I think you can remove the message.

src/geometry.rs
```diff
@@ -864,6 +3088,71 @@ mod tests {
...
+        assert!(oriented_edges_to_check
+            .iter()
+            .all(|o_e| oriented_edges_correct.iter().any(|e| e == o_e)));
+
+        let len_1 = oriented_edges_to_check.len();
+        let len_2 = oriented_edges_correct.len();
+
+        assert_eq!(
+            len_1, len_2,
+            "oriented_edges_to_check.len() = {}, oriented_edges_correct.len() = {}",
```
yanchith: You can remove this assert message.

src/geometry.rs
```diff
@@ -864,6 +3088,71 @@ mod tests {
...
+            OrientedEdge::new(0, 1),
+            OrientedEdge::new(1, 2),
+            OrientedEdge::new(2, 0),
+            OrientedEdge::new(2, 3),
+            OrientedEdge::new(3, 0),
+            OrientedEdge::new(0, 2),
+        ];
+        let oriented_edges_to_check: Vec<OrientedEdge> = geometry.oriented_edges_iter().collect();
+
+        assert!(oriented_edges_to_check
```
yanchith: Same as above: you need to swap the inner and outer loops for this test to actually test anything.

yanchith (changes requested):
Whew, what a large PR 🥇 

I think the library logic is mostly correct, but see inline comments for my suggestions and questions. Logic of some tests is almost certainly incorrect (also see inline comments).

There's a ton of formatting issues as well because `#[rustfmt::skip]` usage and then not formatting the data manually.

Also... there's about 3 different features contained in the PR... let's make it 3 different PRs in the future, so we can iterate better. I see little reason why this had to come in one package and not subsequent smaller ones.

src/edge_analysis.rs
janper:
```
error[E0191]: the value of the associated type `IntoIter` (from the trait `std::iter::IntoIterator`) must be specified
  --> src\edge_analysis.rs:21:39
   |
21 | pub fn edge_valencies(oriented_edges: dyn IntoIterator<Item = &OrientedEdge>) -> EdgeCountMap {
   |                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ associated type `IntoIter` must be specified
```
any suggestion?

src/mesh_analysis.rs
janper: Resolved IRL

src/mesh_analysis.rs
janper: true that

src/mesh_analysis.rs
janper: yes. for the sake of testing, this mesh has a broken (non-matching) winding

src/mesh_analysis.rs
janper: true, trying to fix but test fails and due to ownership i don't know how to look into the generated data. help needed!

src/geometry.rs
janper: >Also the fact it takes a hash set only to get it's `len` is worrisome.

what do you suggest? asking for the length only would move the logic outside of the function. also the same applies to `&self` - that's also needed just for the lengths of its fields

janper: > This shouldn't be a method on Geometry if you need to provide external data, possibly coming from another mesh.

OK

src/geometry.rs
janper: How can i test the custom hash function though?
```
impl Hash for UnorientedEdge {
    fn hash<H: Hasher>(&self, state: &mut H) {
        cmp::min(self.0.vertices.0, self.0.vertices.1).hash(state);
        cmp::max(self.0.vertices.0, self.0.vertices.1).hash(state);
    }
}
```

src/mesh_analysis.rs
janper: this function was moved one of the previous PRs, probably got forgotten here.

src/geometry.rs
yanchith: > what do you suggest?

~My bad: we can just compute the number of edges each time as we already have the data. Let's keep this a method and just take the count from the iterator you can readily spawn.~

~This can stay a method, but it should take just self.~

Nevermind! I forgot we need to deduplicate. I stick by my original suggestion. Make it a standalone function taking 3 numbers.

src/geometry.rs
yanchith: With prop tests. https://github.com/BurntSushi/quickcheck

The property we are testing is this: https://doc.rust-lang.org/std/hash/trait.Hash.html#hash-and-eq `k1 == k2 -> hash(k1) == hash(k2)`

Let's not do it yet. Enough to add a FIXME for now. But this test needs to be removed.

src/edge_analysis.rs
```diff
@@ -3,53 +3,55 @@ use std::collections::HashMap;
...
-/// ascending_count contains number of edges oriented from lower index to higher
-/// descending_count contains number of edges oriented from higher index to lower
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub struct EdgeCount {
-    pub ascending_count: u32,
-    pub descending_count: u32,
+/// ascending_edges contains edges oriented from lower index to higher
+/// descending_edges contains edges oriented from higher index to lower
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SimilarEdges {
```
yanchith, line 9: ~Why this change? What was wrong about only keeping the counts only?~

Sorry, resolved on slack.

src/convert.rs
```diff
@@ -17,6 +17,7 @@ where
 ///
 /// # Panics
 /// Panics if the conversion returns an error.
+#[allow(dead_code)]
```
yanchith: After implementing other changes I am proposing, this allow is unnecessary. Let's remove.

src/mesh_analysis.rs
```diff
@@ -74,159 +74,50 @@ pub fn border_vertex_indices(edge_valencies: &EdgeCountMap) -> HashSet<u32> {
...
+    edge_valencies.iter().all(|(_, edge_count)| {
+        edge_count.ascending_edges.len() == 1 && edge_count.descending_edges.len() == 1
     })
 }
 
+/// Genus of a mesh is the number of holes in topology / conectivity
+/// The mesh must be triangular and watertight
+/// V - E + F = 2 (1 - G)
+#[allow(dead_code)]
+pub fn mesh_genus(vertex_count: i32, edge_count: i32, face_count: i32) -> i32 {
```
yanchith: Let's make the parameters `u32` since they are counts and do the casts internally in this function. The caller doesn't need to know.

src/mesh_analysis.rs
```diff
@@ -74,159 +74,50 @@ pub fn border_vertex_indices(edge_valencies: &EdgeCountMap) -> HashSet<u32> {
...
-        ];
-
-        (faces, vertices)
-    }
+    use crate::geometry::{self, cube_sharp_var_len, Geometry, NormalStrategy, UnorientedEdge};
+    use crate::test_geometry_fixtures::{
+        cube_sharp_mismatching_winding, double_torus, non_manifold_shape, quad, torus, triple_torus,
+    };
 
-    fn cube_sharp_broken_winding(position: [f32; 3], scale: f32) -> Geometry {
```
yanchith: Sorry if I wasn't clear. I only meant to move the large fixtures (torus, double_torus, triple_torus) to the separate file. We should keep the others here.

src/mesh_analysis.rs
```diff
@@ -248,19 +139,20 @@ mod tests {
...
+        let oriented_edges_with_valency_1: Vec<_> =
+            find_edges_with_valency(&edge_valency_map, 1).collect();
+        let oriented_edges_with_valency_2: Vec<_> =
+            find_edges_with_valency(&edge_valency_map, 2).collect();
 
-        for o_e in oriented_edges_with_valency_1 {
-            assert!(oriented_edges_with_valency_1_correct
-                .iter()
-                .any(|e| *e == o_e));
+        assert_eq!(&oriented_edges_with_valency_1.len(), &cast_usize(4));
```
yanchith: Unnecessary `&` on both side of the assert, and unnecessary `cast_usize` (99% sure)

src/mesh_analysis.rs
```diff
@@ -248,19 +139,20 @@ mod tests {
...
+        assert_eq!(&oriented_edges_with_valency_2.len(), &cast_usize(2));
+
+        for o_e in oriented_edges_with_valency_1_correct {
+            assert!(&oriented_edges_with_valency_1.iter().any(|e| *e == o_e));
```
yanchith: Unnecessary `&`, also below.

src/mesh_analysis.rs
```diff
@@ -327,11 +219,11 @@ mod tests {
...
             non_manifold_edges(&edge_valency_map).collect();
 
-        for o_e in &oriented_edges_non_manifold_check {
-            assert!(oriented_edges_non_manifold_correct.iter().any(|e| e == o_e));
+        for o_e in &oriented_edges_non_manifold_correct {
+            assert!(oriented_edges_non_manifold_check.iter().any(|e| e == o_e));
         }
 
-        assert_eq!(oriented_edges_non_manifold_check.len(), 1);
+        assert_eq!(oriented_edges_non_manifold_check.len(), 3);
```
yanchith, line 343: Did this change to 3, because there is now no de-duplication in the `non_manifold_edges()`?

yanchith (commented):
Thank you for the changes. A few new minor inconsistencies were introduced, but we are getting close!

src/mesh_analysis.rs
yanchith: Alternatively, we could keep them as `usize` intead of `u32`, since that's most likely what they'll be.

src/mesh_analysis.rs
janper: why though?

src/mesh_analysis.rs
janper: because 1 was a wrong assumption. since we're asking for oriented edges, they are in fact 3 (therefore the edge is non-manifold). there is, however, just one edge, appearing 3 times as unoriented or twice plus one as oriented

yanchith: kk, thought so, thanks!

src/mesh_analysis.rs
yanchith: Multiple reasons:

- From the clarity perspective It's good to have (small, for some definition of small) test data close to the tests that use them
- By having the test data in behind `#[cfg(test)]`, we don't have to compile them for non-test builds

For large test data we will eventually have fixture files we will be loading them from.

yanchith:
Good enough for me! @ondrowan ?

Also, let's resolve conflicts.

janper:
> Also, let's resolve conflicts.

rebase onto master, squash, merge?

ondrowan (dismissed):
I wrote down some things that need to be changed, but it can wait. I'll change them later.

yanchith (dismissed):

yanchith (approved):
Squash and shipit!

PR #35, opened 2019-10-12
Implement interpreter with variable cache and invalidation

yanchith:
This is the first iteration of our Rooty-Lang interpreter. What's
missing:

- The ability to pause and resume the interpreter
- Iterative funcs a.k.a. Generators

Be aware that there is the barest amount of program
verification implemented, and that is mostly runtime. Compile time
verification is yet to come. Invalid programs may crash the
interpreter.

The interpreter is integrated into the editor with `InterpreterServer`,
which works very similarly to what the Importer worker does.


src/geometry.rs
```diff
@@ -25,7 +25,7 @@ pub enum NormalStrategy {
 ///
 /// The geometry data lives in right-handed coordinate space with the
 /// XY plane being the ground and Z axis growing upwards.
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Default)]
```
ondrowan: Is this `Default` needed? It doesn't seem to do anything at first glance.

src/interpreter_funcs.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+pub fn global_definitions() -> Vec<Box<dyn Func>> {
+    // IMPORTANT: Do not re-order, ever!
+    // The index into this array serves as the unique function
+    // identifier throughout the program.
+    vec![
+        Box::new(FuncImplCreateUvSphere),
+        Box::new(FuncImplShrinkWrap),
+    ]
```
ondrowan: This seems to be rather fragile and I'd feel better if it was a hash map with unique identifiers. Just in case there would be later a valid reason to reorder it.

src/interpreter_server.rs
```diff
@@ -0,0 +1,226 @@
...
+    },
+    Shutdown,
+}
+
+struct Response {
+    request_id: RequestId,
+    data: InterpreterResponse,
+}
+
+pub struct InterpreterServer {
```
ondrowan, line 67: Add docstrings for server and friendly structures (where it makes sense).

src/interpreter_server.rs
```diff
@@ -0,0 +1,226 @@
...
+                            "Interpreter server received request 'SetProg' with {}",
+                            prog,
+                        );
+                        interpreter.set_prog(prog);
+                        response_sender
+                            .send(Response {
+                                request_id,
+                                data: InterpreterResponse::Completed,
+                            })
+                            .expect("Interpreter server failed to send response");
```
ondrowan: I think it'd be a bit lighter and more readable if only responses were produced by this `match` and then sent after it.

ondrowan: And since `match` arms are gonna be much shorter, you can remove the blank lines between them to match our style (for short `match` arms) elsewhere.

src/interpreter_server.rs
```diff
@@ -0,0 +1,226 @@
...
+            .send(Request::Command {
+                request_id,
+                data: request,
+            })
+            .expect("Interpreter client failed to send request");
+
+        request_id
+    }
+
+    pub fn poll_response(&mut self) -> Result<(RequestId, InterpreterResponse), PollResponseError> {
```
ondrowan: `self` doesn't need to be mutable here.

src/interpreter_server.rs
```diff
@@ -0,0 +1,226 @@
...
+            Ok((response.request_id, response.data))
+        }
+    }
+}
+
+impl Drop for InterpreterServer {
+    fn drop(&mut self) {
+        self.request_sender
+            .send(Request::Shutdown)
+            .expect("Interpreter clinet failed to send shutdown request");
```
ondrowan: "clinet" -> "client"

src/interpreter/mod.rs
```diff
@@ -0,0 +1,1320 @@
...
+
+/// A runtime error.
+#[derive(Debug, PartialEq)]
+pub enum RuntimeError {
+    UndeclaredVarUse(usize, ast::VarIdent),
+    UndeclaredFuncUse(usize, ast::FuncIdent),
+    ArgCountMismatch(usize, ast::CallExpr, usize, usize),
+    ArgTyMismatch(usize, ast::CallExpr, bool, Ty, Ty),
+    ReturnTyMismatch(usize, ast::CallExpr, Ty, Ty),
+}
```
ondrowan: At first glance I'm not sure what are any of the fields. I'd write some general docstring description for the type and its variants. It'd be maybe even worth considering using structures with fields for self-documenting purposes.

src/interpreter/mod.rs
```diff
@@ -0,0 +1,1320 @@
...
+pub use self::func::{Func, FuncFlags, ParamInfo};
+pub use self::value::{Ty, Value};
+
+pub mod ast;
+pub mod func;
+pub mod value;
+
+/// A name resolution error.
+#[derive(Debug, PartialEq)]
+pub enum ResolveError {}
```
ondrowan, line 15: Are there ever gonna be variants for this and `TypecheckError`? Is there any advantage of using enum instead of unit structure?

src/interpreter/mod.rs
```diff
@@ -0,0 +1,1320 @@
...
+        self.env.clear();
+        self.pc = 0;
+        self.prog = ast::Prog::default();
+    }
+
+    pub fn push_prog_stmt(&mut self, stmt: ast::Stmt) {
+        self.prog.push_stmt(stmt);
+    }
+
+    pub fn pop_prog_stmt(&mut self) {
```
ondrowan, line 224: This behaves differently than `pop` methods in `std`. I think it would be cleaner if it returned `Option`. The same goes for `Prog::pop_stmt`.

src/interpreter/mod.rs
```diff
@@ -0,0 +1,1320 @@
...
+        // FIXME: @Diagnostics Implement type-checking
+
+        Ok(())
+    }
+
+    pub fn interpret(&mut self) -> Result<Value, InterpretError> {
+        assert!(
+            !self.prog.stmts().is_empty(),
+            "Can not execute empty program",
+        );
```
ondrowan, line 274: Since only `interpret_up_until` is called from this function, I think this assert should be handled there.

src/interpreter/mod.rs
```diff
@@ -0,0 +1,1320 @@
...
+
+    pub fn interpret_up_until(&mut self, index: usize) -> Result<Value, InterpretError> {
+        self.resolve()?;
+        self.typecheck()?;
+
+        let max_index = self.prog.stmts().len() - 1;
+        assert!(
+            max_index >= index,
+            "Can not execute past the program lenght",
+        );
```
ondrowan, line 288: Shouldn't this be handled before resolve and typecheck to prevent extra work?

src/interpreter/mod.rs
```diff
@@ -0,0 +1,1320 @@
...
+                    .expect("Value must have been populated or already cached");
+
+                var_info.value.clone()
+            }
+        };
+
+        Ok(value)
+    }
+
+    fn invalidate(&mut self) {
```
ondrowan, line 347: This function deserves a docstring. Could the comment below be moved to it?

src/interpreter/mod.rs
```diff
@@ -0,0 +1,1320 @@
...
+) -> Result<(), RuntimeError> {
+    let var_ident = var_decl.ident();
+
+    // This is a false positive. Bad Clippy, bad! Rewriting the code
+    // to use the entry API would fail borrowchecking (and cause
+    // pointer invalidation if it didn't!). The entry would create a
+    // long-lived mutable borrow which simply can't be live when we
+    // call into `eval_call_expr`, which borrows again. Note that
+    // having `eval_call_expr` guarded by the map access is the whole
+    // point here.
```
ondrowan, line 450: Have you created an issue in clippy repo for this? If not, please do so and link it here, so we can fix it later (and you'll get to write one of your favourite FIXMEs as a bonus).

src/interpreter/mod.rs
```diff
@@ -0,0 +1,1320 @@
...
+        let prog = ast::Prog::new(vec![ast::Stmt::VarDecl(ast::VarDeclStmt::new(
+            ast::VarIdent(0),
+            ast::CallExpr::new(ast::FuncIdent(func_id), vec![]),
+        ))]);
+
+        let mut interpreter = Interpreter::new(vec![Box::new(func)]);
+        interpreter.set_prog(prog);
+
+        let _ = interpreter.interpret_up_until(2);
+    }
```
ondrowan, line 907: It'd probably make sense to test also execution of empty program and valid partial execution.

src/interpreter/func.rs
```diff
@@ -0,0 +1,75 @@
...
+    /// Unset bits are always the safe default. Set bits may trigger
+    /// interpreter optimizations. Incorrectly set bits may result in
+    /// bugs.
+    pub struct FuncFlags: u8 {
+        /// The function is pure (referentially transparent), if its
+        /// result only depends on its arguments. Pure functions may
+        /// not be re-run by the interpreter, if their inputs did not
+        /// change. This flag triggers optimizations. Setting it
+        /// incorrectly might produce stale results.
+        const PURE = 0b_0000_0001;
```
ondrowan, line 17: Is there ever going to be anything but fn's purity as a flag?

src/interpreter/func.rs
```diff
@@ -0,0 +1,75 @@
...
+    /// for more.
+    ///
+    /// [`ParamInfo`]: struct.ParamInfo.html
+    fn param_info(&self) -> &[ParamInfo];
+
+    /// Information about the function's return type.
+    ///
+    /// Used for static and dynamic typecheking. See [`Ty`] for more.
+    ///
+    /// [`Ty`]: enum.Ty.html
```
ondrowan: This link does not work.

src/interpreter/func.rs
```diff
@@ -0,0 +1,75 @@
...
+    /// [`Nil`]: enum.Ty.html#variant.Nil
+    pub optional: bool,
+}
+
+/// An interface describing a function as seen by the interpreter.
+///
+/// Functions are pieces of callable code. They can receive parameters
+/// and must produce a return value, even if [`Nil`].
+///
+/// [`Nil`]: enum.Ty.html#variant.Nil
```
ondrowan: This link does not work.

src/interpreter/func.rs
```diff
@@ -0,0 +1,75 @@
...
+pub trait Func {
+    /// Information about the function behaviour.
+    ///
+    /// See [`Funcflags`] for more.
```
ondrowan: Nit: Capitalize to `FuncFlags`.

ondrowan (changes requested):
It looks good overall, but there are couple of things to be addressed before we can merge it. I hope I haven't half arsed any part of it, but just to be sure, expect that I find some new things on the next sweep.

src/geometry.rs
yanchith: 🤔 it doesn't, weird.

src/interpreter_funcs.rs
yanchith: Was half thinking to do this myself. The result is a bit more verbose, but hopefully more robust. Let's do it!

src/interpreter/mod.rs
yanchith: These fields might change a lot yet, so I would wait with the docstrings, but I agree that struct enum variants could be more readable here, something like: 

```rust
pub enum RuntimeError {
    UndeclaredVarUse {
        line: usize,
        var: ast::VarIdent
    }
    /// ...
}
```

src/interpreter/mod.rs
yanchith: Yes! (Although depending on our milestones and deadlines it may not happen very soon). 

The gist will be the same as for the runtime error: The most complete information possible will be given so that the UI can pin-point and display the error message on a concrete widget.

I already have a WIP version of the name resolving and its corresponding errors. Some errors will move from the `RuntimeError` to the `ResolveError`, and will be impossible to encounter during program execution (modulo bugs). Similar should later happen for `TypecheckError`.

src/interpreter/mod.rs
yanchith: True, but depends on whether we view popping the last statement off of an empty program as an error, or not. In our case the one calling this should know whether this will succeed beforehand. I initially thought it would be better not to hide this logic error by preventing a crash, but do not feel strongly about it - if you insist, I'll change it.

src/interpreter/mod.rs
yanchith: This is actually a different check and the one in `interpret_up_until` wouldn't catch this. But actually this check is also needed there, so I'll move it.

The check in `interpret_up_until` would underflow the `usize` if the program is empty, therefore it needs to be a separate check. `usize` underflows are panics with debug assertions and wrapping arithmetic (I think ?) without them.

EDIT: scratch that... need to check in both functions, as the underflow can happen in `interpret()` too.

src/interpreter/mod.rs
yanchith: We are going to crash anyway, but there's no harm in it.

src/interpreter/mod.rs
yanchith: Now I have https://github.com/rust-lang/rust-clippy/issues/4674

src/interpreter/func.rs
yanchith: I believe so, but can't really tell when. I expect all these flags to be heuristics the interpreter can use to optimize _something_ about how the function is executed. If you feel strongly about this, I can remove flags now and add them back once we need more than one.

yanchith:
@ondrowan I think I addressed or answered most of your points, please have another look and let me know your decision on the few points I asked questions about.

ondrowan (approved):

PR #36, opened 2019-10-12
Use distance squared in a few places and remove FIXMEs

yanchith:
Turns out I was not thinking right when I made those FIXME's. If x1 >
x2, x1^2 > x2^2 for values above zero, which distances are.


janper (approved):
Cool. This is a standard way of comparing distances and thus should work well. 
I'd like to see some comparative benchmarks between `distance` and `distance_squared` in Rust once we are not busy with other things.

ondrowan (approved):
🚢 🇮🇹

PR #37, opened 2019-10-14
Remove orphan vertices and normals

janper:
The pull request deals with orphan vertices and normals - data not referenced in any indexing structure (in our case a face) and thus not used by our software (yet and probably never). 

Two new functions are introduced:

- `remove_orphan_vertices(faces: Vec<(u32, u32, u32)>, vertices: Vec<Point3<f32>>) -> (Vec<(u32, u32, u32)>, Vec<Point3<f32>>)`
- `remove_orphan_normals(faces: Vec<TriangleFace>, normals: Vec<Vector3<f32>>) -> (Vec<TriangleFace>, Vec<Vector3<f32>>)`

The functions are intended to be used when the mesh geometry is being created, therefore the input and the output is not the geometry itself but rather its source data. The two functions are independent and can be used separately. For sake of performance, there could be a compound function removing both types of orphans at once.

The main struggle is the renumbering of the data indices (in our case `TriangleFace`s) to match the reduced set of vertices or normals respectively. The functions first create maps of old index -> new index, a list of reduced data (vertices or normals), then renumber the face data and return both, the reduced dataset (vertices or normals) and renumbered faces. 

 The orphan removal is optional (because we don't know why do they exist in the first place; they may have been created intentionally), therefore there are two new geometry constructors: `from_triangle_faces_with_vertices_and_computed_normals_remove_orphans`, which removes orphan vertices and `from_triangle_faces_with_vertices_and_normals_remove_orphans`, which removes both types or orphans, the vertices and the normals.

### DIFF
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -364,11 +364,11 @@ fn remove_orphan_vertices(
     faces: Vec<(u32, u32, u32)>,
     vertices: Vertices,
 ) -> (Vec<(u32, u32, u32)>, Vertices) {
-    let mut vertices_reduced: Vertices = Vec::new();
+    let mut vertices_reduced: Vertices = Vec::with_capacity(vertices.len());
     let original_vertex_len = vertices.len();
     let unused_vertex_marker = vertices.len();
     let mut old_new_vertex_map: Vec<usize> = vec![unused_vertex_marker; original_vertex_len];
-    let mut faces_renumbered: Vec<(u32, u32, u32)> = Vec::new();
+    let mut faces_renumbered: Vec<(u32, u32, u32)> = Vec::with_capacity(faces.len());
 
     for face in faces {
         let old_vertex_index_0 = cast_usize(face.0);
@@ -408,6 +408,9 @@ fn remove_orphan_vertices(
         ));
     }
 
+    faces_renumbered.shrink_to_fit();
+    vertices_reduced.shrink_to_fit();
+
     (faces_renumbered, vertices_reduced)
 }
 
@@ -416,11 +419,11 @@ fn remove_orphan_normals(
     faces: Vec<TriangleFace>,
     normals: Normals,
 ) -> (Vec<TriangleFace>, Normals) {
-    let mut normals_reduced: Normals = Vec::new();
+    let mut normals_reduced: Normals = Vec::with_capacity(normals.len());
     let original_normal_len = normals.len();
     let unused_normal_marker = normals.len();
     let mut old_new_normal_map: Vec<usize> = vec![unused_normal_marker; original_normal_len];
-    let mut faces_renumbered: Vec<TriangleFace> = Vec::new();
+    let mut faces_renumbered: Vec<TriangleFace> = Vec::with_capacity(faces.len());
 
     for face in faces {
         let old_normal_index_0 = cast_usize(face.normals.0);
@@ -463,6 +466,9 @@ fn remove_orphan_normals(
         ));
     }
 
+    faces_renumbered.shrink_to_fit();
+    normals_reduced.shrink_to_fit();
+
     (faces_renumbered, normals_reduced)
 }
 
@@ -471,16 +477,17 @@ fn remove_orphan_vertices_and_normals(
     vertices: Vertices,
     normals: Normals,
 ) -> (Vec<TriangleFace>, Vertices, Normals) {
-    let mut vertices_reduced: Vertices = Vec::new();
+    let mut vertices_reduced: Vertices = Vec::with_capacity(vertices.len());
     let original_vertex_len = vertices.len();
     let unused_vertex_marker = vertices.len();
     let mut old_new_vertex_map: Vec<usize> = vec![unused_vertex_marker; original_vertex_len];
 
-    let mut normals_reduced: Normals = Vec::new();
+    let mut normals_reduced: Normals = Vec::with_capacity(normals.len());
     let original_normal_len = normals.len();
     let unused_normal_marker = normals.len();
     let mut old_new_normal_map: Vec<usize> = vec![unused_normal_marker; original_normal_len];
-    let mut faces_renumbered: Vec<TriangleFace> = Vec::new();
+
+    let mut faces_renumbered: Vec<TriangleFace> = Vec::with_capacity(faces.len());
 
     for face in faces {
         let old_vertex_index_0 = cast_usize(face.vertices.0);
@@ -552,6 +559,10 @@ fn remove_orphan_vertices_and_normals(
             cast_u32(new_normal_index_2),
         ));
     }
+
+    faces_renumbered.shrink_to_fit();
+    vertices_reduced.shrink_to_fit();
+    normals_reduced.shrink_to_fit();
 
     (faces_renumbered, vertices_reduced, normals_reduced)
 }
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -16,7 +16,6 @@ pub enum NormalStrategy {
     // FIXME: add `Smooth`
 }
 
-pub type FaceVertexTuples = Vec<(u32, u32, u32)>;
 pub type Vertices = Vec<Point3<f32>>;
 pub type Normals = Vec<Vector3<f32>>;
 
@@ -46,7 +45,7 @@ impl Geometry {
     /// # Panics
     /// Panics if faces refer to out-of-bounds vertices.
     pub fn from_triangle_faces_with_vertices_and_computed_normals(
-        faces: FaceVertexTuples,
+        faces: Vec<(u32, u32, u32)>,
         vertices: Vertices,
         normal_strategy: NormalStrategy,
     ) -> Self {
@@ -105,7 +104,7 @@ impl Geometry {
     /// # Panics
     /// Panics if faces refer to out-of-bounds vertices.
     pub fn from_triangle_faces_with_vertices_and_computed_normals_remove_orphans(
-        faces: FaceVertexTuples,
+        faces: Vec<(u32, u32, u32)>,
         vertices: Vertices,
         normal_strategy: NormalStrategy,
     ) -> Self {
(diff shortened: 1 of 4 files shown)


src/geometry.rs
```diff
@@ -96,6 +96,67 @@ impl Geometry {
...
+    /// # Panics
+    /// Panics if faces refer to out-of-bounds vertices.
+    pub fn from_triangle_faces_with_vertices_and_computed_normals_remove_orphans(
+        faces: Vec<(u32, u32, u32)>,
+        vertices: Vec<Point3<f32>>,
+        normal_strategy: NormalStrategy,
+    ) -> Self {
+        let (faces_purged, vertices_purged) = remove_orphan_vertices(faces, vertices);
+
+        let mut normals = Vec::with_capacity(faces_purged.len());
```
yanchith: I think this can just re-use `from_triangle_faces_with_vertices_and_computed_normals` from here on.

src/geometry.rs
```diff
@@ -146,6 +207,70 @@ impl Geometry {
...
+
+        let (faces_purged, normals_purged) = remove_orphan_normals(
+            faces_prepurged
+                .iter()
+                .map(|f| TriangleFace::new(f.0, f.1, f.2))
+                .collect(),
+            normals,
+        );
+
+        let vertices_range = 0..cast_u32(vertices_purged.len());
```
yanchith: I think this should just re-use `from_triangle_faces_with_vertices_and_normals` from here on.

src/geometry.rs
```diff
@@ -322,6 +447,88 @@ impl Hash for UnorientedEdge {
...
+                    .get(&f.2)
+                    .expect("Vertex index not found"),
+            )
+        })
+        .collect();
+
+    (recomputed_faces, reduced_vertices)
+}
+
+#[allow(dead_code)]
```
yanchith, line 417: Let's remove this `allow` and add it to the constructor that uses it.

src/geometry.rs
```diff
@@ -322,6 +447,88 @@ impl Hash for UnorientedEdge {
     }
 }
 
+#[allow(dead_code, clippy::type_complexity)]
```
yanchith: Let's remove all these allows, they shouldn't be the default 😉 

If clippy complains about the type signature of `(Vec<(u32, u32, u32)>, Vec<Point3<f32>>)`, let's do what it suggests: create a type alias for it.

src/geometry.rs
```diff
@@ -322,6 +447,88 @@ impl Hash for UnorientedEdge {
...
+fn remove_orphan_vertices(
+    faces: Vec<(u32, u32, u32)>,
+    vertices: Vec<Point3<f32>>,
+) -> (Vec<(u32, u32, u32)>, Vec<Point3<f32>>) {
+    let mut old_to_new_vertex_index: HashMap<u32, u32> = HashMap::new();
+    let mut reduced_vertices: Vec<Point3<f32>> = Vec::new();
+    let mut new_index = 0;
+    #[allow(clippy::needless_range_loop)]
```
yanchith: Remove this allow and do what clippy says:

```rust
    for (original_index, vertex) in vertices.iter().enumerate() {
        let o_i = cast_u32(original_index);
        if faces.iter().any(|f| f.0 == o_i || f.1 == o_i || f.2 == o_i) {
            old_to_new_vertex_index.insert(o_i, new_index);
            reduced_vertices.push(*vertex);
            new_index += 1;
        }
    }
```

src/geometry.rs
```diff
@@ -322,6 +447,88 @@ impl Hash for UnorientedEdge {
...
+
+#[allow(dead_code)]
+fn remove_orphan_normals(
+    faces: Vec<TriangleFace>,
+    normals: Vec<Vector3<f32>>,
+) -> (Vec<TriangleFace>, Vec<Vector3<f32>>) {
+    let mut old_to_new_normal_index: HashMap<u32, u32> = HashMap::new();
+    let mut reduced_normals: Vec<Vector3<f32>> = Vec::new();
+    let mut new_index = 0;
+    #[allow(clippy::needless_range_loop)]
```
yanchith: Remove allow, do what clippy suggests.

src/geometry.rs
```diff
@@ -1141,4 +1348,106 @@ mod tests {
 
         assert!(!geometry_with_orphans.has_no_orphan_normals());
     }
+    #[test]
+    fn test_remove_orphan_vertices() {
+        let (faces, vertices, _normals) = quad_with_normals();
```
yanchith: Use `quad` instead of `quad_with_normals`

src/geometry.rs
```diff
@@ -1141,4 +1348,106 @@ mod tests {
...
+        let extra_vertex = vec![v(0.0, 0.0, 0.0, [0.0, 0.0, 0.0], 1.0)];
+        let vertices_extended = [&extra_vertex[..], &vertices[..]].concat();
+        let faces_renumbered_to_match_extend_vertices: Vec<_> = faces
```
yanchith: Can't you simply append the vertex instead of prepending it so you don't have to renumber? Or is this on purpose?

src/geometry.rs
```diff
@@ -1141,4 +1348,106 @@ mod tests {
...
+        assert_eq!(vertices_purged, vertices);
+    }
+
+    #[test]
+    fn test_remove_orphan_normals() {
+        let (faces, _vertices, normals) = quad_with_normals();
+        let extra_normal = vec![n(0.0, 0.0, 0.0)];
+        let normals_extended = [&extra_normal[..], &normals[..]].concat();
+
+        let faces_renumbered_to_match_extend_normals: Vec<_> = faces
```
yanchith, line 1414: Same as above. Why prepend and renumber when you can append?

src/geometry.rs
```diff
@@ -1141,4 +1348,106 @@ mod tests {
...
+        assert_eq!(faces_length, faces_purged_length);
+        assert_eq!(normals_purged, normals);
+    }
+
+    #[test]
+    fn test_geometry_from_triangle_faces_with_vertices_and_computed_normals_remove_orphans() {
+        let (faces, vertices, _normals) = quad_with_normals();
+        let extra_vertex = vec![v(0.0, 0.0, 0.0, [0.0, 0.0, 0.0], 1.0)];
+        let vertices_extended = [&extra_vertex[..], &vertices[..]].concat();
+        let faces_renumbered_to_match_extend_vertices: Vec<_> = faces
```
yanchith: Same as above. Why prepend and renumber when you can append?

src/geometry.rs
```diff
@@ -1141,4 +1348,106 @@ mod tests {
...
+
+        let faces_purged_length = &faces_purged.len();
+
+        assert_eq!(faces_length, faces_purged_length);
+        assert_eq!(normals_purged, normals);
+    }
+
+    #[test]
+    fn test_geometry_from_triangle_faces_with_vertices_and_computed_normals_remove_orphans() {
+        let (faces, vertices, _normals) = quad_with_normals();
```
yanchith: `quad_with_normals` -> `quad`

src/geometry.rs
```diff
@@ -1141,4 +1348,106 @@ mod tests {
...
+        let extra_vertex = vec![v(0.0, 0.0, 0.0, [0.0, 0.0, 0.0], 1.0)];
+        let vertices_extended = [&extra_vertex[..], &vertices[..]].concat();
+        let faces_renumbered_to_match_extend_vertices: Vec<_> = faces
+            .iter()
+            .map(|f| (f.vertices.0 + 1, f.vertices.1 + 1, f.vertices.2 + 1))
+            .collect();
+
+        let geometry =
+            Geometry::from_triangle_faces_with_vertices_and_computed_normals_remove_orphans(
+                faces_renumbered_to_match_extend_vertices.clone(),
```
yanchith: No need to clone faces and vertices

src/geometry.rs
```diff
@@ -1141,4 +1348,106 @@ mod tests {
...
+
+    #[test]
+    fn test_geometry_from_triangle_faces_with_vertices_and_normals_remove_orphans() {
+        let (faces, vertices, normals) = quad_with_normals();
+        let extra_vertex = vec![v(0.0, 0.0, 0.0, [0.0, 0.0, 0.0], 1.0)];
+        let vertices_extended = [&extra_vertex[..], &vertices[..]].concat();
+        let extra_normal = vec![n(0.0, 0.0, 0.0)];
+        let normals_extended = [&extra_normal[..], &normals[..]].concat();
+
+        let faces_renumbered_to_match_extend_vertices_and_normals: Vec<_> = faces
```
yanchith, line 1504: Same as above. Why prepend and renumber when you can append?

src/geometry.rs
```diff
@@ -1141,4 +1348,106 @@ mod tests {
...
+                    f.vertices.2 + 1,
+                    f.normals.0 + 1,
+                    f.normals.1 + 1,
+                    f.normals.2 + 1,
+                )
+            })
+            .collect();
+
+        let geometry = Geometry::from_triangle_faces_with_vertices_and_normals_remove_orphans(
+            faces_renumbered_to_match_extend_vertices_and_normals.clone(),
```
yanchith: No need to clone these things

yanchith (changes requested):
As we talked in person, let's not do the `O(m*n)` implementation if we can avoid it.

Also let's match the constructors for now: have one orphan removal function for vertices-only, one for normals-only (which will be unused for now), and one compound for both. It would be great, if we could join the vertices-only and normals-only functions into just one generic function... but I don't want to torture you that much. I would also accept not doing the normals-only function, since it doesn't seem like we will need it anytime soon.

You may also want to remove the FIXMEs regarding orphan removal in the other geometry constructors.

src/geometry.rs
janper: this is on purpose. removing an orphan vertex from the end of the list wouldn't trigger faces to renumber its references into the list of vertices.

src/geometry.rs
janper: same as above

src/geometry.rs
janper: same as above

src/geometry.rs
janper: same as above

janper:
> As we talked in person, let's not do the `O(m*n)` implementation if we can avoid it.

I hope I've come up with an optimized algorithm for the orphan removal. Unless I understand the implementation of certain data structures wrong. The code isn't very elegant and i can imagine how to reduce the number of lines but i'm worried it would become unreadable whatsoever. Any improvement suggestions are welcome though!

src/geometry.rs
```diff
@@ -16,6 +16,10 @@ pub enum NormalStrategy {
     // FIXME: add `Smooth`
 }
 
+pub type FaceVertexTuples = Vec<(u32, u32, u32)>;
```
yanchith: This name is a bit weird. Can we leave this type unnamed? I checked and clippy doesn't throw a hissy fit.

src/geometry.rs
```diff
@@ -447,86 +361,186 @@ impl Hash for UnorientedEdge {
...
+    let mut vertices_reduced: Vertices = Vec::new();
+    let original_vertex_len = vertices.len();
+    let unused_vertex_marker = vertices.len();
+    let mut old_new_vertex_map: Vec<usize> = vec![unused_vertex_marker; original_vertex_len];
+    let mut faces_renumbered: FaceVertexTuples = Vec::new();
+    for face in faces {
+        let old_vertex_index_0 = cast_usize(face.0);
+        let new_vertex_index_0 = if old_new_vertex_map[old_vertex_index_0] == unused_vertex_marker {
+            vertices_reduced.push(vertices[old_vertex_index_0]);
+            let new_index = vertices_reduced.len() - 1;
```
yanchith: I would move this `let new_index =` one line above, and get rid of the `- 1`. The current form triggered the underflow detector in my brain 👻 . Same for all analogous code in this functions and the others.

src/geometry.rs
```diff
@@ -447,86 +361,186 @@ impl Hash for UnorientedEdge {
...
-                    .expect("Vertex index not found"),
-            )
-        })
-        .collect();
-
-    (recomputed_faces, reduced_vertices)
+    faces: FaceVertexTuples,
+    vertices: Vertices,
+) -> (FaceVertexTuples, Vertices) {
+    let mut vertices_reduced: Vertices = Vec::new();
```
yanchith: Let's create some sensible breathing space for this function (and the normals-only one) and insert a few newlines here and there. The style in the vertex and normal orphan removal function is good to apply here too.

yanchith (commented):
Almost good to go! See the few minor comments inline.

yanchith (dismissed):
Nice! 👍 We can squash and merge once CI passes, unless @ondrowan wants to have a look also.

ondrowan (approved):

yanchith (approved):
<3

PR #38, opened 2019-10-16
Implement interpreter result value-sets

yanchith:
Now each interpreting returns a list of all used/unused values in
addition to the last value.

Should enable us to fulfill various UI requirements.


src/interpreter/mod.rs
```diff
@@ -316,7 +336,53 @@ impl Interpreter {
...
+        }
+
+        Ok(ValueSet {
+            last_value,
+            used_values,
+            unused_values,
+        })
+    }
+
+    fn compute_unused_vars_up_until(&self, index: usize) -> HashSet<ast::VarIdent> {
```
ondrowan, line 375: This function could use a docstring.

src/interpreter/mod.rs
```diff
@@ -1261,10 +1432,24 @@ mod tests {
...
         let value = interpreter.interpret().unwrap();
-        assert_eq!(value, Value::Boolean(true));
+        assert_eq!(
+            value,
+            ValueSet {
+                last_value: Value::Boolean(true),
+                used_values: vec![Value::Boolean(true)],
+                unused_values: vec![Value::Boolean(true)],
+            }
+        );
```
ondrowan: I think it'd be cleaner if all these tests cared only about `value.last_value` as before, so they don't test multiple things. Cases that test explicitly for used and unused values could be added. This way it'd be also easier to maintain them in case the structure changes.

ondrowan (changes requested):

src/interpreter/mod.rs
yanchith: Makes sense to add tests specifically for the new structure and keep old tests on `last_value`, I'll do it.

yanchith:
@ondrowan addressed, and also incorporated the change we talked about IRL - used and unused variable values are now tupled with their variable idents.

src/interpreter/mod.rs
ondrowan: I think there are some typos in it now. "Compute" -> "Computes" and it seems there's a word missing in second part of the sentence.

ondrowan:
I'll approve this once the typos are fixed and master is merged to this branch.

src/interpreter/mod.rs
yanchith: > "Compute" -> "Computes"

Not a typo. It was on purpose, but I just checked the std doc comments and this style is not used, so I'll stop using it. Don't know why I thought this was the official style.

I'll fix the rest.

ondrowan:
Rebase and 🚢 :it:

ondrowan (approved):

PR #40, opened 2019-10-17
Calculate element to element topologies

janper:
This pull request contains most of the topology (connectivity) calculation needed for further mesh crowling algorithms:

* `face_to_face_topology` returns a map, where the key is index into the list of mesh's own faces for each face and the value is a set of neighboring face indices

* `edge_to_face_topology` returns a map, where the key is index into the provided set of unoriented edges for each edge and the value is a set of face indices containing the unoriented edge

* `vertex_to_face_topology` returns a map, where the key is index into the list of mesh's own vertices for each vertex and the value is a set of faces containing the vertex indices

* `vertex_to_edge_topology `returns a map, where the key is index into the list of mesh's own vertices for each vertex and the value is a set of provided edge indices containing the vertex

* `vertex_to_vertex_topology `returns a map, where the key is index into the list of mesh's own vertices for each vertex and the value is a set of mesh's own vertex indices directly linked via an edge with the respective vertex


src/geometry.rs
```diff
@@ -215,16 +215,30 @@ impl Geometry {
...
+    /// Extracts unoriented edges from all mesh faces
     pub fn unoriented_edges_iter<'a>(&'a self) -> impl Iterator<Item = UnorientedEdge> + 'a {
         self.triangle_faces_iter()
             .flat_map(|face| ArrayVec::from(face.to_unoriented_edges()).into_iter())
     }
 
+    /// Genus of a mesh is the number of holes in topology / conectivity
+    /// The mesh must be triangular and watertight
+    /// V - E + F = 2 (1 - G)
+    pub fn mesh_genus(&self, edges: &HashSet<UnorientedEdge>) -> i32 {
```
yanchith: I think this came from beyond the grave - there already is a `mesh_genus` function in `mesh_analysis.rs`.

src/geometry.rs
```diff
@@ -339,9 +461,12 @@ impl OrientedEdge {
     pub fn is_reverted(self, other: OrientedEdge) -> bool {
         self.vertices.0 == other.vertices.1 && self.vertices.1 == other.vertices.0
     }
+
+    pub fn contains_vertex(self, vertex_index: u32) -> bool {
+        self.vertices.0 == vertex_index || self.vertices.1 == vertex_index
+    }
 }
 
-/// Implements orientation indifferent hash and equal methods
```
yanchith, line 344: Probably a bad merge somewhere down the line: let's not remove this comment.

src/geometry.rs
```diff
@@ -352,7 +477,6 @@ impl PartialEq for UnorientedEdge {
     }
 }
 
-// FIXME: test
```
yanchith: Let's not remove this

src/geometry.rs
```diff
@@ -1039,6 +1165,28 @@ mod tests {
         (faces, vertices, normals)
     }
 
+    fn tessellated_triangle() -> (Vec<(u32, u32, u32)>, Vec<Point3<f32>>) {
+        #[rustfmt::skip]
+            let vertices = vec![
```
yanchith: Not formatted correctly. The point of using `#[rustfmt::skip]` is to format the thing manually, if you know better than the formatter.

src/geometry.rs
```diff
@@ -1039,6 +1165,28 @@ mod tests {
...
+            Point3::new(-2.0, -2.0, 0.0),
+            Point3::new(0.0, -2.0, 0.0),
+            Point3::new(2.0, -2.0, 0.0),
+            Point3::new(-1.0, 0.0, 0.0),
+            Point3::new(1.0, 0.0, 0.0),
+            Point3::new(0.0, 2.0, 0.0),
+        ];
+
+        #[rustfmt::skip]
```
yanchith: No need for the `#[rustfmt::skip]` here at all.

src/geometry.rs
```diff
@@ -1386,6 +1534,187 @@ mod tests {
...
+        let geometry = Geometry::from_triangle_faces_with_vertices_and_computed_normals(
+            faces.clone(),
+            vertices.clone(),
+            NormalStrategy::Sharp,
+        );
+        let edges: HashSet<UnorientedEdge> = geometry.unoriented_edges_iter().collect();
+
+        let genus = geometry.mesh_genus(&edges);
+        assert_eq!(genus, 2);
+    }
```
yanchith: Did all of these mesh genus tests come back alive? There's copies of them in `mesh_analysis.rs`

Also the one below, but the screen wasn't large enough for me to select them all 😂

yanchith (commented):
Let's clean this up (fix bad merge artifacts) before I do a proper review.

One thing though: I have an uneasy feeling about the `HashSet` inside the `HashMap`. I feel this deserves its own data structure. I'll think about it today.

yanchith:
Hmmm, maybe `smallvec` can be useful here. If we manage to not use `usize`. https://docs.rs/smallvec/0.6.10/smallvec/

Just out of interest, what are the usual/expected neighborhood counts for various topologies?

EDIT: a quick investigation shows that a `HashSet` is a very large type and it's not so hard to beat: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&code=use%20std%3A%3Acollections%3A%3AHashSet%3B%0Ause%20std%3A%3Amem%3B%0A%0Ause%20smallvec%3A%3ASmallVec%3B%20%2F%2F%200.6.10%0A%0Afn%20main()%20%7B%0A%20%20%20%20dbg!(mem%3A%3Asize_of%3A%3A%3CHashSet%3Cusize%3E%3E())%3B%0A%0A%20%20%20%20dbg!(mem%3A%3Asize_of%3A%3A%3CSmallVec%3C%5Busize%3B%205%5D%3E%3E())%3B%0A%20%20%20%20dbg!(mem%3A%3Asize_of%3A%3A%3CSmallVec%3C%5Bu32%3B%2010%5D%3E%3E())%3B%0A%7D

If we used a 5 element smallvec of usize, we are still on par with the hash set. If we switched to u32, we can hold 10 elements before needing to allocate.

janper:
> Just out of interest, what are the usual/expected neighborhood counts for various topologies?

mostly just 2 or 4, however if the mesh contains a triangle fan, it can be even hundreds and the mesh will still be valid.

src/geometry.rs
```diff
@@ -8,6 +8,8 @@ use nalgebra as na;
 use nalgebra::base::Vector3;
 use nalgebra::geometry::Point3;
 
+use smallvec::SmallVec;
```
yanchith: Let's move the import to the previous import group :wink:

src/geometry.rs
```diff
@@ -244,6 +250,138 @@ impl Geometry {
         }
         used_normals.len() == self.normals().len()
     }
+
+    /// Calculates topological relations (neighborhood) of mesh face -> faces.
+    /// Returns a Map (key: face index, value: list of its neighboring faces indices)
+    pub fn face_to_face_topology(&self) -> HashMap<usize, SmallVec<[usize; 16]>> {
```
yanchith: If we are going to keep using `usize` as indices, let's have at most 8 elements. Applies to all the other smallvecs in this PR.

src/geometry.rs
```diff
@@ -244,6 +250,138 @@ impl Geometry {
...
+    }
+
+    /// Calculates topological relations (neighborhood) of mesh edge -> faces.
+    /// Returns a Map (key: edge index, value: list of its neighboring faces indices)
+    pub fn edge_to_face_topology(
+        &self,
+        edges: &HashSet<UnorientedEdge>,
+    ) -> HashMap<usize, SmallVec<[usize; 16]>> {
+        let mut e2f: HashMap<usize, SmallVec<[usize; 16]>> = HashMap::new();
+        for (from_counter, e) in edges.iter().enumerate() {
```
yanchith: This is dubious. `HashSet`s have randomization and reseeding, so no two instances will contain the elements in the same order.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&code=use%20std%3A%3Acollections%3A%3AHashSet%3B%0Ause%20std%3A%3Aiter%3A%3AFromIterator%3B%0A%0Afn%20main()%20%7B%0A%20%20%20%20let%20data%3A%20Vec%3Cu8%3E%20%3D%20vec!%5B0%2C%201%2C%202%2C%203%2C%204%2C%205%2C%206%2C%207%2C%208%2C%209%5D%3B%0A%20%20%20%20let%20set1%3A%20HashSet%3Cu8%3E%20%3D%20HashSet%3A%3Afrom_iter(data.iter().copied())%3B%0A%20%20%20%20let%20set2%3A%20HashSet%3Cu8%3E%20%3D%20HashSet%3A%3Afrom_iter(data.iter().copied())%3B%0A%20%20%20%20%0A%20%20%20%20println!(%22--%20SET1%20--%22)%3B%0A%20%20%20%20for%20(i%2C%20elem)%20in%20set1.iter().enumerate()%20%7B%0A%20%20%20%20%20%20%20%20println!(%22%7B%7D%3A%20%7B%7D%22%2C%20i%2C%20elem)%3B%0A%20%20%20%20%7D%0A%0A%20%20%20%20println!(%22--%20SET2%20--%22)%3B%0A%20%20%20%20for%20(i%2C%20elem)%20in%20set2.iter().enumerate()%20%7B%0A%20%20%20%20%20%20%20%20println!(%22%7B%7D%3A%20%7B%7D%22%2C%20i%2C%20elem)%3B%0A%20%20%20%20%7D%0A%7D

Will we be able to guarantee we always have the same instance of the set around? Seems tricky to do. Shouldn't we have a way of creating and using the edges in a stable order? I imagine they could be a `Vec`, but we would use a `HashSet` during the creation to perform deduplication.

Also applies to the other method using edges.

src/geometry.rs
```diff
@@ -244,6 +250,138 @@ impl Geometry {
         }
         used_normals.len() == self.normals().len()
     }
+
+    /// Calculates topological relations (neighborhood) of mesh face -> faces.
+    /// Returns a Map (key: face index, value: list of its neighboring faces indices)
+    pub fn face_to_face_topology(&self) -> HashMap<usize, SmallVec<[usize; 16]>> {
+        let mut f2f: HashMap<usize, SmallVec<[usize; 16]>> = HashMap::new();
+        for (from_counter, f) in self.triangle_faces_iter().enumerate() {
```
yanchith: I know we only have triangle faces so far, but once we add e.g. quads, this will continue to compile but will no longer be correct. Shouldn't we iterate `self.faces` instead so we are forced to handle all kinds of faces?

Also applies everywhere else this is used in this PR.

src/geometry.rs
```diff
@@ -313,6 +451,19 @@ impl TriangleFace {
...
+    pub fn contains_vertex(&self, vertex_index: u32) -> bool {
+        self.vertices.0 == vertex_index
+            || self.vertices.1 == vertex_index
+            || self.vertices.2 == vertex_index
+    }
+
+    /// Does the face contain the specific unoriented edge
+    pub fn contains_unoriented_edge(&self, unoriented_edge: UnorientedEdge) -> bool {
+        let [o_e_0, o_e_1, o_e_2] = &self.to_unoriented_edges();
+        o_e_0 == &unoriented_edge || o_e_1 == &unoriented_edge || o_e_2 == &unoriented_edge
```
yanchith: All ampersands in this function's body seem extraneous.

yanchith (changes requested):
Very nice tests and very understandable PR ❤️ 

Just out of curiosity, are there any better-than-quadratic algorithms for building the topologies?

I have also raised some issues we will have to resolve inline, or at least talk to them. I think the most important one is the order dependent hash-set iteration.

janper:
> Just out of curiosity, are there any better-than-quadratic algorithms for building the topologies?

vertex-to-vertex is not quadratic already. in fact, only face-to-face needs to be quadratic i think, but then we wouldn't overuse the already calculated edges and the logic would be slightly more complex. all and all, i think optimization is not worth the effort.

src/geometry.rs
janper: this can be any `Iterator` or directly a `Vec` in fact. it will work for duplicates as well, so it can be the responsibility of the caller to purge or keep the duplicates.

src/geometry.rs
janper: True that but i also expect you'll want to put all the topology-computing functions outside the `Geometry` and this is the only face getter right now. Shall we add a general face getter, jut like the one for vertices?

src/edge_analysis.rs
```diff
@@ -61,7 +61,6 @@ mod tests {
     }
 
     fn quad() -> (Vec<(u32, u32, u32)>, Vertices) {
-        #[rustfmt::skip]
```
yanchith: Let's not do this. It is in an unrelated file.

src/mesh_topology_analysis.rs
```diff
@@ -0,0 +1,313 @@
...
+/// Calculates topological relations (neighborhood) of mesh face -> faces.
+/// Returns a Map (key: face index, value: list of its neighboring faces indices)
+#[allow(dead_code)]
+pub fn face_to_face_topology(geometry: &Geometry) -> HashMap<usize, SmallVec<[usize; 8]>> {
+    let mut f2f: HashMap<usize, SmallVec<[usize; 8]>> = HashMap::new();
+    for (from_counter, face) in geometry.faces().iter().enumerate() {
+        match face {
+            Face::Triangle(f) => {
+                let [f_e_0, f_e_1, f_e_2] = f.to_unoriented_edges();
+                for (to_counter, t_f) in geometry.triangle_faces_iter().enumerate() {
```
yanchith: Forgot to use `geometry.faces()` here instead of `geometry.triangle_faces_iter()`.

I have also created an issue to remove `geometry.triangle_faces_iter()` https://trello.com/c/vFi2zRtW/151-remove-geometrytrianglefacesiter

src/mesh_topology_analysis.rs
```diff
@@ -0,0 +1,313 @@
...
+    }
+    v2e
+}
+
+/// Calculates topological relations (neighborhood) of mesh vertex -> vertex.
+/// Returns a Map (key: vertex index, value: list of its neighboring vertices indices)
+#[allow(dead_code)]
+pub fn vertex_to_vertex_topology(geometry: &Geometry) -> HashMap<usize, SmallVec<[usize; 8]>> {
+    let mut v2v: HashMap<usize, SmallVec<[usize; 8]>> = HashMap::new();
+    for f in geometry.triangle_faces_iter() {
```
yanchith: Let's use `geometry.faces()` here

yanchith (commented):
Almost there! Just got a few nits to fix. @ondrowan you can start your review now if you want.

yanchith (approved):
Let's wait for @ondrowan 's review also, but I am ok with this.

ondrowan (approved):

ondrowan:
🚢 🇮🇹

PR #41, opened 2019-10-19
Replace test donuts with their low poly versions

janper:
The previous test geometry was unnecessarily heavy and (therefore) placed in a separate module, which was outside of the test scope. 
This commit replaces the geometries with their equivalents with the lowest possible number of vertices and faces, which makes it quite short and thus possible to include into the test scope.


yanchith (approved):
❤️

PR #42, opened 2019-10-21
Add basic name resolution

yanchith:
This allows us to remove a few runtime errors, and make them
"compile-time" errors instead. When we decide to do static
typechecking in the future, this can act as the guideline for adding
it.

As far as the frontend/UI is concerned, name resolution errors should
ideally never happen. Maybe the sensible strategy for the UI is to panic on
these, but otherwise make sure correct programs are constructed.

### DIFF
--- a/src/interpreter/mod.rs
+++ b/src/interpreter/mod.rs
@@ -14,7 +14,7 @@ pub mod value;
 /// A name resolution error.
 #[derive(Debug, PartialEq)]
 pub enum ResolveError {
-    VarRedefinition { line: usize, var: ast::VarIdent },
+    VarRedefinition { line: usize, var: VarIdent },
     UndeclaredVarUse { line: usize, var: VarIdent },
     UndeclaredFuncUse { line: usize, func: FuncIdent },
 }
--- a/src/interpreter/mod.rs
+++ b/src/interpreter/mod.rs
@@ -287,6 +287,12 @@ impl Interpreter {
         *stmt_mut = stmt;
     }
 
+    /// Runs name resolution on the currently set program.
+    ///
+    /// More concretely, this statically verifies that:
+    /// - Each used function is registered with the interpreter
+    /// - Each variable is defined before it is used
+    /// - Variables are not re-declared
     pub fn resolve(&mut self) -> Result<(), ResolveError> {
         if self.last_resolve_epoch == self.epoch {
             return Ok(());
@@ -337,6 +343,11 @@ impl Interpreter {
         Ok(())
     }
 
+    /// Interprets the whole currently set program and returns the
+    /// used/unused values after the last statment.
+    ///
+    /// # Panics
+    /// Panics if the currently set program is empty.
     pub fn interpret(&mut self) -> InterpretResult {
         assert!(
             !self.prog.stmts().is_empty(),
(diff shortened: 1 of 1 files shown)


src/interpreter/mod.rs
```diff
@@ -13,11 +13,27 @@ pub mod value;
 
 /// A name resolution error.
 #[derive(Debug, PartialEq)]
-pub enum ResolveError {}
+pub enum ResolveError {
+    VarRedefinition { line: usize, var: ast::VarIdent },
```
ondrowan: There's no need to use `ast::` here. Let's be consistent.

src/interpreter/mod.rs
```diff
@@ -547,13 +598,12 @@ fn eval_var_decl_stmt(
 }
 
 fn eval_expr(
```
ondrowan, line 622: It's not the scope of this PR, but wouldn't it make sense for these `eval_` functions to be methods of `Interpreter`? Most of them need `env` or `funcs` and as far as I see, they aren't used anywhere but in scope of `Interpreter`.

src/interpreter/mod.rs
```diff
@@ -271,11 +283,50 @@ impl Interpreter {
...
 
     pub fn resolve(&mut self) -> Result<(), ResolveError> {
-        // FIXME: @Diagnostics Implement name resolution
+        if self.last_resolve_epoch == self.epoch {
+            return Ok(());
+        }
+
+        let mut var_scope = HashSet::new();
+
+        for (line, stmt) in self.prog.stmts().iter().enumerate() {
```
ondrowan: Shouldn't we count lines from 1? Seems to be more standard.

src/interpreter/mod.rs
```diff
@@ -271,11 +283,50 @@ impl Interpreter {
             .get_mut(index)
             .expect("Expected the program to have index-th element");
 
+        self.epoch += 1;
         *stmt_mut = stmt;
     }
 
     pub fn resolve(&mut self) -> Result<(), ResolveError> {
```
ondrowan, line 300: Now `resolve` actually does something, it'd deserve a docstring.

ondrowan (changes requested):

src/interpreter/mod.rs
yanchith: I am glad you asked! They were originally (long time ago) a part of the interpreter, but a pretty nasty borrowchecker problem surfaced. Unfortunately I now don't remember what it was or even if it's still there 😂

src/interpreter/mod.rs
yanchith: 🤔 either that or we can rename it to `stmt_index` or something. Thoughts?

src/interpreter/mod.rs
yanchith: Just checked. This is the problem:

```rust
for (line, stmt) in self.prog.stmts()[range].iter().enumerate() {
    self.eval_stmt(line, stmt)?;
    self.pc += 1;
}
```

The for loop creates a shared borrow on `self.prog`, but `self.eval_stmt(line, stmt)` creates a unique borrow on `self`, which conflicts. I had the execution part of the interpreter initially artificially split to separate struct so that this doesn't happen, but it was far less readable and more boilerplatey, so I went for the simple functions instead.

src/interpreter/mod.rs
ondrowan: I don't know what's usual for this kind of problem. I'd personally just count from 1.

yanchith: Actually the more I think about it, the less I believe the "line" name does us justice.

Also just realized the problem with counting from 1 is that reporting the error in the UI will once again have to subtract 1 to get the index of the UI widget where to display the error. Maybe we should not pretend this is text based?

yanchith: @ondrowan if you agree, I'd rename occurrences of `line` to `stmt_index`.

ondrowan: That'd probably make sense.

ondrowan (approved):

PR #44, opened 2019-10-21
Mark FPS window as debug mode only

ondrowan:


yanchith (approved):
I'd later resurrect this as part of the "renderer options" window

PR #45, opened 2019-10-22
Operation: Laplacian smoothing

janper:
This commit implements Laplacian smoothing routine and creates an interpreter operation utilizing it. 
Laplacian smoothing (relaxation) is an iterative mesh transformation resulting in a smoother mesh without adding new vertices or faces. Each vertex is moved to an average position of its neighbors. The routine is usually applied in several (3-10) consecutive iterations, too many iterations, however, cause the model to "disappear" into a small round amorphous goo.
The tests of the Laplacian smoothing use data snapshots of the same algorithm implemented in Grasshopper. The tests also prove the vertex-to-vertex topology computation works as expected.


yanchith:
Nice, but once again: this is not rebased/merged correctly and contains changes already included in other branches. Please fix before we can review.

yanchith:
Also: I quickly skimmed it, and instead of `allow(dead_code)` I would strongly prefer, if this was visually checkable. There is a hardcoded program in `lib.rs` you can use to run the thing.

janper:
> Also: I quickly skimmed it, and instead of `allow(dead_code)` I would strongly prefer, if this was visually checkable. There is a hardcoded program in `lib.rs` you can use to run the thing.

This needs a proper input model for a visual test to be relevant. The model I'm using now relaxes into an ugly mess, but it's a correct result. I know, because I made the same algorithm in Grasshopper and compared it with industry standard algorithms on both, simple and proper models.

janper:
Ready for review
laplacian

src/geometry.rs
```diff
@@ -366,7 +366,6 @@ impl OrientedEdge {
     }
 }
 
-/// Implements orientation indifferent hash and equal methods
```
yanchith: Probably removed by a bad merge?

src/geometry.rs
```diff
@@ -377,7 +376,6 @@ impl PartialEq for UnorientedEdge {
     }
 }
 
-// FIXME: test
```
yanchith: Removed by a bad merge?

src/interpreter_funcs.rs
```diff
@@ -77,11 +78,50 @@ impl Func for FuncImplShrinkWrap {
...
+
+    fn return_ty(&self) -> Ty {
+        Ty::Geometry
+    }
+
+    fn call(&self, args: &[Value]) -> Value {
+        let geometry = args[0].unwrap_geometry();
+        let iterations = args[1].unwrap_uint();
+
+        assert!(
```
yanchith: This will crash the interpreter. We must not do this yet.

This is something we will have the ability to do later. Most ideally, we would specify min and max value in the `ParamInfo` which both the UI and the interpreter would have to uphold. Once we have that in place, this can become an assert again.

For now let's make this a clamp and then the cast, and also add the FIXME so we know we need to come back to it once we support value ranges..

Also: I think we shouldn't force the number of iterations to be above zero, but rather handle that as a valid edge case. While not useful, the user might have a reason to do it we don't antiicipate yet (maybe performance for prototyping).

src/mesh_smoothing.rs
```diff
@@ -0,0 +1,582 @@
+use nalgebra::geometry::Point3;
+
+use crate::convert::cast_u32;
+use crate::geometry::Geometry;
+use crate::mesh_topology_analysis::vertex_to_vertex_topology;
+
+/// # Mesh relaxation / Laplacian smoothing
```
yanchith: This is not usually how these doc comments are formatted. When in doubt, always check some doc comments in std (`rustup doc --std`).

More concretely, the first line of a function doc is usually not a markdown title. The first sentence or a couple of sentences should summarize what the function does, with more elaborate explanation following later (separated by a newline).

This is a good example: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts

Same applies to the other doc comment.

src/mesh_smoothing.rs
```diff
@@ -0,0 +1,582 @@
...
+/// The more iterations, the smoother result.
+/// Too many iterations may cause slow calculation time.
+///
+/// The algorithm is based on replacing each vertex position
+/// with an average position of its immediate neighbors
+#[allow(dead_code)]
+pub fn laplacian_smoothing_with_anchors(
+    geometry: &Geometry,
+    iterations: u8,
+    fixed_vertex_indices: &[u32],
```
yanchith, line 40: Let's also mention in the doc comment, what this does.

src/mesh_smoothing.rs
```diff
@@ -0,0 +1,582 @@
...
+            {
+                let mut average_position: Point3<f32> = Point3::origin();
+                for neighbor_index in neighbors_indices {
+                    average_position += geometry_vertices[*neighbor_index].coords;
+                }
+                average_position /= neighbors_indices.len() as f32;
+                vertices[*current_vertex_index] = average_position;
+            }
+        }
+        geometry_vertices = vertices.clone();
```
yanchith: If you restructure the code to clone one of `vertices`/`geometry_vertices` this at the beginning of each iteration only, you can get rid of this clone which is unused in the last iteration.

src/mesh_smoothing.rs
```diff
@@ -0,0 +1,582 @@
...
+        for (current_vertex_index, neighbors_indices) in vertex_to_vertex_topology.iter() {
+            if fixed_vertex_indices
+                .iter()
+                .all(|i| *i != cast_u32(*current_vertex_index))
+            {
+                let mut average_position: Point3<f32> = Point3::origin();
+                for neighbor_index in neighbors_indices {
+                    average_position += geometry_vertices[*neighbor_index].coords;
+                }
+                average_position /= neighbors_indices.len() as f32;
```
yanchith, line 62: Do we hope here the geometry is without orphans? If it has a vertex with no neighbors, this will produce an `inf` value. Not sure what do do, maybe the easiest way is to add an explicit branch and not relax a vertex without neighbors.

src/mesh_smoothing.rs
```diff
@@ -0,0 +1,582 @@
...
+                vertices[*current_vertex_index] = average_position;
+            }
+        }
+        geometry_vertices = vertices.clone();
+    }
+
+    Geometry::from_triangle_faces_with_vertices_and_normals(
+        geometry.triangle_faces_iter().collect(),
+        vertices,
+        geometry.normals().to_vec(),
```
yanchith, line 72: What is more correct? Re-use stale normals after relaxation, or force re-compute of sharp normals?

src/mesh_smoothing.rs
```diff
@@ -0,0 +1,582 @@
...
+#[cfg(test)]
+mod tests {
+    use nalgebra::distance_squared;
+
+    use crate::edge_analysis::edge_valencies;
+    use crate::geometry::{Geometry, NormalStrategy, OrientedEdge, TriangleFace, Vertices};
+    use crate::mesh_analysis::border_vertex_indices;
+
+    use super::*;
+
```
yanchith, line 87: Can you please add a `FIXME: snapshot testing` here? We should at some point have a standard set of testing geometries for these things and not define them in code.

src/mesh_smoothing.rs
```diff
@@ -0,0 +1,582 @@
...
+    #[test]
+    fn test_laplacian_smoothing_preserves_face_vertex_normal_count() {
+        let (faces, vertices) = torus();
+        let geometry = Geometry::from_triangle_faces_with_vertices_and_computed_normals(
+            faces.clone(),
+            vertices.clone(),
+            NormalStrategy::Sharp,
+        );
+        let relaxed_geometry_1 = laplacian_smoothing(&geometry, 1);
```
yanchith, line 420: Let's also test 0. Apart from not being useful, the alorithm has no reason to break apart for 0 iterations. The usual algorithmic test cases are 0, 1 and a really big number (10 suffices as that in this case).

src/mesh_smoothing.rs
```diff
@@ -0,0 +1,582 @@
...
+    Geometry::from_triangle_faces_with_vertices_and_normals(
+        geometry.triangle_faces_iter().collect(),
+        vertices,
+        geometry.normals().to_vec(),
+    )
+}
+
+#[cfg(test)]
+mod tests {
+    use nalgebra::distance_squared;
```
yanchith: Functions are usually not imported like this (with a very few exceptions). Instead, use the qualified name in the code. Applies to all of the imports.

Usages in code usually look like: `let b = mesh_analysis::border_vertices(...);`

src/mesh_smoothing.rs
```diff
@@ -0,0 +1,582 @@
...
+        assert_eq!(relaxed_geometry_1_i_faces, test_geometry_1_i_faces);
+        assert_eq!(relaxed_geometry_3_i_faces, test_geometry_3_i_faces);
+
+        const TOLERANCE_SQUARED: f32 = 0.01 * 0.01;
+
+        let relaxed_geometry_1_i_vertices = relaxed_geometry_1_i.vertices();
+        let test_geometry_1_i_vertices = test_geometry_1_i.vertices();
+
+        for i in 0..test_geometry_1_i_vertices.len() {
+            assert!(
```
yanchith, line 502: I think the `distance_squared` asserts here should become `approx` asserts: https://docs.rs/approx/0.3.2/approx/
We already have `approx` as a dependency.

From the top of my head, I don't know whether to test for relative difference or ULPs, but we should probably pick one of these strategies and stick with it.

Same for every `distance_squared` assert from here to the end of file.

src/mesh_smoothing.rs
```diff
@@ -0,0 +1,582 @@
...
+        let (faces, vertices) = shape_for_smoothing_with_anchors();
+        let geometry = Geometry::from_triangle_faces_with_vertices_and_computed_normals(
+            faces.clone(),
+            vertices.clone(),
+            NormalStrategy::Sharp,
+        );
+
+        let oriented_edges: Vec<OrientedEdge> = geometry.oriented_edges_iter().collect();
+        let edge_valency_map = edge_valencies(&oriented_edges);
+        let fixed_vertex_indices: Vec<_> = border_vertex_indices(&edge_valency_map)
```
yanchith: This can be written more succinctly by using `FromItarator`.

```rust
use std::iter::FromIterator;
// ... later ...
let vector = Vec::from_iter(set.into_iter());
```

src/mesh_smoothing.rs
```diff
@@ -0,0 +1,582 @@
...
+        for i in 0..test_geometry_vertices.len() {
+            assert!(
+                distance_squared(&test_geometry_vertices[i], &relaxed_geometry_vertices[i])
+                    < TOLERANCE_SQUARED
+            );
+        }
+    }
+
+    #[test]
+    fn test_laplacian_smoothing_with_anchors_find_border_vertices() {
```
yanchith, line 566: Does this test the same thing as the previous test, but only finds the anchor points algorithmically?

src/mesh_smoothing.rs
```diff
@@ -0,0 +1,582 @@
...
+pub fn laplacian_smoothing_with_anchors(
+    geometry: &Geometry,
+    iterations: u8,
+    fixed_vertex_indices: &[u32],
+) -> Geometry {
+    let vertex_to_vertex_topology = vertex_to_vertex_topology(geometry);
+    let mut geometry_vertices: Vec<Point3<f32>> = Vec::from(geometry.vertices());
+    let mut vertices = geometry_vertices.clone();
+
+    for _ in 0..iterations {
```
yanchith, line 50: If we have 0 iterations, let's just clone and return the original geometry.

src/interpreter_funcs.rs
```diff
@@ -77,11 +78,50 @@ impl Func for FuncImplShrinkWrap {
     }
 }
 
+pub struct FuncImplLaplacianSmoothing;
+impl Func for FuncImplLaplacianSmoothing {
+    fn flags(&self) -> FuncFlags {
+        FuncFlags::PURE
```
yanchith, line 144: Just a note. I actually spent a non-trivial amount of time on this review making sure this holds 😂 These things are tricky.

yanchith (changes requested):
Very relaxing ❤️ 😉 

The logic seems correct, but I do have a few suggestions for improving it as well.

Also: it now seems like rust is a bit less of a problem for you now, so I started commenting on a few things I previously omitted so I wouldn't overwhelm you, like the style of doc comments.

src/interpreter_funcs.rs
janper: 0 is a valid param now. clamping and removing the assert.

src/mesh_smoothing.rs
janper: good point. this actually destroys the orphans. i've changed the logic so that it preserver the orphans just like the fixed vertices

src/mesh_smoothing.rs
janper: neither of the two. we don't know if the mesh was created with sharp normals or not. also we don't know if the same normal is being reused for each face vertex. 
this is a matter of a decision but i think this should compute smooth normals for the result, as this is a smoothing algorithm. adding a `FIXME`

src/mesh_smoothing.rs
janper: yes

src/interpreter_funcs.rs
janper: so, is it correct or not?

janper:
Ready for another review

src/mesh_smoothing.rs
```diff
@@ -1,49 +1,104 @@
...
     for _ in 0..iterations {
+        geometry_vertices = vertices.clone();
         for (current_vertex_index, neighbors_indices) in vertex_to_vertex_topology.iter() {
-            let mut average_position: Point3<f32> = Point3::origin();
-            for neighbor_index in neighbors_indices {
-                average_position += geometry_vertices[*neighbor_index].coords;
+            if fixed_vertex_indices
+                .iter()
+                .all(|i| *i != cast_u32(*current_vertex_index))
+                || neighbors_indices.is_empty()
```
yanchith: Shouldn't this be `!neighbors_indices.is_empty()`?

EDIT: Actually, something else entirely: 

```rust
            if fixed_vertex_indices
                .iter()
                .all(|i| *i != cast_u32(*current_vertex_index))
                && !neighbors_indices.is_empty()
```

src/mesh_smoothing.rs
```diff
@@ -1,49 +1,104 @@
...
+    // FIXME: Calculate smooth normals for the result once we support them
     Geometry::from_triangle_faces_with_vertices_and_normals(
-        geometry.triangle_faces_iter().collect(),
+        faces,
         vertices,
         geometry.normals().to_vec(),
     )
 }
 
+fn filter_triangular_faces(geometry: &Geometry) -> Vec<TriangleFace> {
```
yanchith: If this is only used in tests, let's move it to the test module.

src/interpreter_funcs.rs
yanchith: I think it is, because it does not use the random order in the hashmap 🤔

src/mesh_smoothing.rs
yanchith: Hmm, is it really necessary to test then?

yanchith (commented):
I think I found one more logic error, but otherwise seems good!

yanchith (dismissed):

src/interpreter_funcs.rs
```diff
@@ -151,6 +189,10 @@ pub fn global_definitions() -> HashMap<FuncIdent, Box<dyn Func>> {
 
     funcs.insert(FUNC_ID_CREATE_UV_SPHERE, Box::new(FuncImplCreateUvSphere));
     funcs.insert(FUNC_ID_SHRINK_WRAP, Box::new(FuncImplShrinkWrap));
+    funcs.insert(
```
yanchith, line 193: Let's keep the same order here as above

yanchith (approved):

PR #46, opened 2019-10-22
Remove triangle_faces_iter and triangle_faces_count from Geometry

yanchith:
It seems this filtering view is virtually never what we want to do,
when we ever have nontriangular faces. It was allowing us to not care
about other kind of faces, potentially leading to incorrect behaviour
in the future.

### DIFF
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -184,25 +184,6 @@ impl Geometry {
         )
     }
 
-    /// Return a view of all triangle faces in this geometry. Skip all
-    /// other types of faces.
-    pub fn triangle_faces_iter<'a>(&'a self) -> impl Iterator<Item = TriangleFace> + 'a {
-        self.faces.iter().copied().map(|index| match index {
-            Face::Triangle(f) => f,
-        })
-    }
-
-    /// Return count of all triangle faces in this geometry. Skip all
-    /// other types of faces.
-    pub fn triangle_faces_len(&self) -> usize {
-        self.faces
-            .iter()
-            .filter(|index| match index {
-                Face::Triangle(_) => true,
-            })
-            .count()
-    }
-
     pub fn faces(&self) -> &[Face] {
         &self.faces
     }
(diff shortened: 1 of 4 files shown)


yanchith:
I realized I still want to add some cleanup to this: Currently, it is very impractical to create new geometry based on these changes, so I'll fix that as well. I'll ping you when this is ready for review.

yanchith:
@ondrowan @janper This is now reviewable again.

src/geometry.rs
```diff
@@ -39,168 +39,235 @@ pub struct Geometry {
...
+        // To avoid one additional cloning of this collection, we
+        // first materialize it into its final structure (Vec<Face>),
+        // and later assert the only kind of faces there are
+        // triangles.
+        let faces_collection: Vec<_> = faces
+            .into_iter()
+            .enumerate()
+            .map(|(i, (i1, i2, i3))| {
+                let normal_index = cast_u32(i);
+                TriangleFace::new_separate(i1, i2, i3, normal_index, normal_index, normal_index)
```
janper: Will the face reference the same normal for each vertex when the `NormalStrategy` is `Smooth`?

src/geometry.rs
```diff
@@ -39,168 +39,235 @@ pub struct Geometry {
...
+
+                }
+                // FIXME: once we add other kinds of faces, they must panic here
+                // _ => panic!("Face must be triangular, we just created it"),
+            }
         }
 
-        assert_eq!(normals.len(), faces.len());
-        assert_eq!(normals.capacity(), faces.len());
+        assert_eq!(normals_collection.len(), faces_collection.len());
```
janper: This also won't be true for `Smooth` normals.

src/geometry.rs
```diff
@@ -39,168 +39,235 @@ pub struct Geometry {
...
+            remove_orphan_vertices(faces.into_iter().collect(), vertices.into_iter().collect());
+        Self::from_triangle_faces_with_vertices_and_computed_normals(
             faces_purged,
             vertices_purged,
             normal_strategy,
         )
     }
 
-    /// Create new triangle face geometry from provided faces,
+    /// Creates new triangular geometry from provided triangle faces,
```
janper: Such a geometry is usually called a *triangulated mesh*

src/geometry.rs
```diff
@@ -221,35 +288,60 @@ impl Geometry {
...
-            .flat_map(|face| ArrayVec::from(face.to_unoriented_edges()).into_iter())
+        self.faces.iter().flat_map(|face| match face {
+            Face::Triangle(triangle_face) => {
+                ArrayVec::from(triangle_face.to_unoriented_edges()).into_iter()
+            }
+        })
+    }
+
+    /// Returns whether the geometry contains exclusively triangular faces.
+    pub fn is_triangular(&self) -> bool {
```
janper: I suggest `is_triangulated` if we agree on this convention

src/mesh_analysis.rs
```diff
@@ -101,9 +101,13 @@ pub fn is_mesh_watertight(edge_valencies: &EdgeCountMap) -> bool {
...
-/// V - E + F = 2 (1 - G)
+/// Computes the mesh genus.
+///
+/// Genus of a mesh is the number of holes in topology / connectivity.
+/// The mesh **must** be triangular and watertight for this to produce
+/// usable results.
+///
+/// The genus (G) is computed as: `V - E + F = 2*(1 - G)`.
 #[allow(dead_code)]
 pub fn mesh_genus(vertex_count: usize, edge_count: usize, face_count: usize) -> i32 {
```
janper: Maybe `triangulated_mesh_genus` to make sure there is no mistake

janper (changes requested):
This is certainly a useful change and a step towards custom (quad or polygonal) mesh faces.

There seems to be one logical mistake in the geometry constructor, I think.

Also we should come up with some naming convention. I started naming my geometry-scope functions `mesh_...`, you call them `geometry_...`. I don't mind either name, but I think we should only pick one and rename all the mismatching functions, variables and rewrite the documentation. If we stick to *geometry*, I'll rename my parts.

There is already an convention for naming a mesh that consists only of triangular faces: **triangulated mesh** (or in our case **triangulated geometry**). I suggest following this convention because *triangular geometry* sounds like something with a triangular shape.

yanchith:
@janper 

> There seems to be one logical mistake in the geometry constructor, I think.

Please do tell 😂

yanchith:
@janper I totally agree with your naming suggestions. I suggest to do it in 2 phases:

1) This PR, I'll make sure everything is "triangulated", but not rename "geometry" to "mesh"
2) I file an issue in trello for the "geometry" to "mesh" rename and we do it when we have less contention (we can also synchronize when we do it). This change would otherwise dramatically increase the number of files this PR affects and we would have merge hell.

src/geometry.rs
yanchith: This will definitely have to be fixed once we add `NormalStrategy::Smooth`, but I would prefer not to do it now, but rather when we implement the smooth strategy (so we can test the fixes are correct). Many other small implementation things will have to be revised once that happens.

For the record, this was already implemented for a while, the only reason it shows up on the diff is because I moved the function impls around.

janper: it seems inconsistent though - there is a separate logic for the normal strategy and not for normal creation. shouldn't all future features be treated the same way (either way)?

janper:
> 2. I file an issue in trello for the "geometry" to "mesh" rename and we do it when we have less contention (we can also synchronize when we do it). This change would otherwise dramatically increase the number of files this PR affects and we would have merge hell.

I don't insist on calling it a **mesh**. It can be a **geometry**. But sure, let's do it when there are no related PRs open.

janper:
> @janper
> 
> > There seems to be one logical mistake in the geometry constructor, I think.
> 
> Please do tell 😂

told you - the normal creation logic

src/geometry.rs
yanchith: Aaah, so that's what you mean by that. Ok, I can wrap it in the match statement if that makes it more consistent for you.

Do note though that probably the whole structure of this function will have to be reworked for smooth normals.

janper: exactly what i wanted to propose

yanchith:
@janper addressed

src/geometry.rs
```diff
@@ -39,168 +39,244 @@ pub struct Geometry {
...
+    where
+        F: IntoIterator<Item = (u32, u32, u32)>,
+        V: IntoIterator<Item = Point3<f32>>,
+    {
+        match normal_strategy {
+            NormalStrategy::Sharp => {
+                // To avoid one additional cloning of this collection, we
+                // first materialize it into its final structure (Vec<Face>),
+                // and later assert the only kind of faces there are
+                // triangles.
```
ondrowan, line 62: Is this a typo?

janper (dismissed):
Thanks!

ondrowan (dismissed):

src/geometry.rs
yanchith: I don't see any 😱

ondrowan (approved):

PR #47, opened 2019-10-22
Find mesh border edge loops

janper:
Find mesh border edge loops, which are circular sequences of edges (closed polylines) around the mesh border. Border edges are contained exclusively in a single face. The feature will be later needed for parametrization of meshes before morphing, for splitting into patches and for physical simulations we might need in some morphing algorithms.

A mesh can have:
* no border edge loop, when it's watertight
* single border edge loop, when it's a patch or a mesh with a single hole (topologically also a patch)
* more border edge loops, when it has more holes or consists of more mesh patches.

A border edge loop always consists of 3 or more border edges (for mesh patches made of a single triangular face or for meshes with a single triangle missing).

In an edge case when more border edge loops share a single vertex (checkerboard scenario), there is no resolution strategy, so the result is left unpredictable, yet correct. In the future, the function can detect the case and throw a warning.

Also, in this pull request most of the valency-related types and functions are renamed more suitably.

### DIFF
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -385,6 +385,12 @@ impl Hash for UnorientedEdge {
     }
 }
 
+impl UnorientedEdge {
+    pub fn share_vertex(self, other: UnorientedEdge) -> bool {
+        self.0.contains_vertex(other.0.vertices.0) || self.0.contains_vertex(other.0.vertices.1)
+    }
+}
+
 fn remove_orphan_vertices(
     faces: Vec<(u32, u32, u32)>,
     vertices: Vertices,
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -385,6 +385,12 @@ impl Hash for UnorientedEdge {
     }
 }
 
+impl UnorientedEdge {
+    pub fn shares_vertex(self, other: UnorientedEdge) -> bool {
+        other.0.contains_vertex(self.0.vertices.0) || other.0.contains_vertex(self.0.vertices.1)
+    }
+}
+
 fn remove_orphan_vertices(
     faces: Vec<(u32, u32, u32)>,
     vertices: Vertices,
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -370,6 +370,12 @@ impl OrientedEdge {
 #[derive(Debug, Clone, Copy, Eq)]
 pub struct UnorientedEdge(pub OrientedEdge);
 
+impl UnorientedEdge {
+    pub fn shares_vertex(self, other: UnorientedEdge) -> bool {
+        other.0.contains_vertex(self.0.vertices.0) || other.0.contains_vertex(self.0.vertices.1)
+    }
+}
+
 impl PartialEq for UnorientedEdge {
     fn eq(&self, other: &Self) -> bool {
         (self.0.vertices.0 == other.0.vertices.0 && self.0.vertices.1 == other.0.vertices.1)
@@ -382,12 +388,6 @@ impl Hash for UnorientedEdge {
     fn hash<H: Hasher>(&self, state: &mut H) {
         cmp::min(self.0.vertices.0, self.0.vertices.1).hash(state);
         cmp::max(self.0.vertices.0, self.0.vertices.1).hash(state);
-    }
-}
-
-impl UnorientedEdge {
-    pub fn shares_vertex(self, other: UnorientedEdge) -> bool {
-        other.0.contains_vertex(self.0.vertices.0) || other.0.contains_vertex(self.0.vertices.1)
     }
 }
 
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -385,12 +385,6 @@ impl Hash for UnorientedEdge {
     }
 }
 
-impl UnorientedEdge {
-    pub fn share_vertex(self, other: UnorientedEdge) -> bool {
-        self.0.contains_vertex(other.0.vertices.0) || self.0.contains_vertex(other.0.vertices.1)
-    }
-}
-
 fn remove_orphan_vertices(
     faces: Vec<(u32, u32, u32)>,
     vertices: Vertices,
--- a/src/mesh_analysis.rs
+++ b/src/mesh_analysis.rs
@@ -74,11 +74,18 @@ pub fn border_vertex_indices(edge_sharing: &EdgeSharingMap) -> HashSet<u32> {
     border_vertices
 }
 
+/// Finds continuous loops of border edges,
+/// starting from a random edge.
+/// The mesh may contain holes or islands,
+/// therefore it may have an unknown number of edge loops.
+/// If two edge loops meet at a single vertex, the result
+/// may be unpredictable and erratic.
 #[allow(dead_code)]
 pub fn edge_loops(edge_sharing: &EdgeSharingMap) -> Vec<Vec<UnorientedEdge>> {
     let mut border_edges: Vec<_> = border_edges(edge_sharing)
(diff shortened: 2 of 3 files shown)


src/geometry.rs
```diff
@@ -385,6 +385,12 @@ impl Hash for UnorientedEdge {
     }
 }
 
+impl UnorientedEdge {
```
yanchith: Let's move this impl above the trait impls. Up for discussion, but I usually provide this first and we should be consistent.

src/mesh_analysis.rs
```diff
@@ -64,25 +64,59 @@ pub fn non_manifold_edges<'a>(
...
 
-    border_edges(edge_valencies).for_each(|edge| {
+    border_edges(edge_sharing).for_each(|edge| {
         border_vertices.insert(edge.vertices.0);
         border_vertices.insert(edge.vertices.1);
     });
     border_vertices
 }
 
+/// Finds continuous loops of border edges,
```
yanchith: Please separate the summary from the rest by a newline. Also could use some text re-flow (most text editors support it)

src/mesh_analysis.rs
```diff
@@ -64,25 +64,59 @@ pub fn non_manifold_edges<'a>(
...
+/// therefore it may have an unknown number of edge loops.
+/// If two edge loops meet at a single vertex, the result
+/// may be unpredictable and erratic.
+#[allow(dead_code)]
+pub fn border_edge_loops(edge_sharing: &EdgeSharingMap) -> Vec<Vec<UnorientedEdge>> {
+    let mut border_edges: Vec<_> = border_edges(edge_sharing).map(UnorientedEdge).collect();
+
+    let mut edge_loops: Vec<Vec<UnorientedEdge>> = Vec::new();
+
+    while !border_edges.is_empty() {
```
yanchith: Can be combined to `while let Some(edge) = border_edges.pop() {`

src/mesh_analysis.rs
```diff
@@ -64,25 +64,59 @@ pub fn non_manifold_edges<'a>(
...
+        if let Some(edge) = border_edges.pop() {
+            let mut current_loop: Vec<UnorientedEdge> = vec![edge];
+
+            while current_loop.len() < 3
+                || !current_loop[current_loop.len() - 1].shares_vertex(current_loop[0])
+            {
+                for (i, other_edge) in border_edges.iter().enumerate() {
+                    if current_loop[current_loop.len() - 1].shares_vertex(*other_edge) {
+                        current_loop.push(border_edges.remove(i));
```
yanchith: Is this the source of erratic behavior from the doc comment? Overlapping loops can "steal" border edges from each other?

src/mesh_analysis.rs
```diff
@@ -114,7 +148,7 @@ mod tests {
     use nalgebra::base::Vector3;
     use nalgebra::geometry::Point3;
 
-    use crate::edge_analysis::edge_valencies;
+    use crate::edge_analysis::edge_sharing;
```
yanchith: Let's not import the function but rather use `edge_analysis::edge_sharing()`

src/mesh_analysis.rs
```diff
@@ -653,4 +720,79 @@ mod tests {
         );
         assert_eq!(genus, 3);
     }
+
+    #[test]
+    fn test_border_edge_loops_should_be_one() {
```
yanchith: Let's name this a bit more explicitely (if you see this test failing in the console, the name doesn't tell you anything). I suggest: `test_border_edge_loops_returns_one_for_xyz`. Same for other test names.

src/mesh_analysis.rs
```diff
@@ -653,4 +720,79 @@ mod tests {
...
+        let (faces, vertices) = tessellated_triangle();
+        let geometry = Geometry::from_triangle_faces_with_vertices_and_computed_normals(
+            faces.clone(),
+            vertices.clone(),
+            NormalStrategy::Sharp,
+        );
+
+        let oriented_edges: Vec<OrientedEdge> = geometry.oriented_edges_iter().collect();
+        let edge_sharing_map = edge_sharing(&oriented_edges);
```
yanchith: Now I am wondering why `edge_sharing` takes oriented edges, only to convert them to unoriented internally. Let's not fix it now, but create a trello issue for later.

src/mesh_analysis.rs
```diff
@@ -653,4 +720,79 @@ mod tests {
...
+            UnorientedEdge(OrientedEdge::new(2, 4)),
+            UnorientedEdge(OrientedEdge::new(4, 5)),
+            UnorientedEdge(OrientedEdge::new(5, 3)),
+            UnorientedEdge(OrientedEdge::new(3, 0)),
+        ];
+
+        let calculated_loops = border_edge_loops(&edge_sharing_map);
+
+        assert_eq!(calculated_loops.len(), 1);
+        for edge in correct_loop {
```
yanchith, line 746: Hmmm, can't we in this case assert also the order? In other words: `assert_eq!(calculated_loops[0], correct_loop);`?

Alternatively, let's at least also assert the length of the loop.

src/mesh_analysis.rs
```diff
@@ -653,4 +720,79 @@ mod tests {
...
+            UnorientedEdge(OrientedEdge::new(3, 0)),
+            UnorientedEdge(OrientedEdge::new(6, 7)),
+            UnorientedEdge(OrientedEdge::new(7, 8)),
+            UnorientedEdge(OrientedEdge::new(8, 6)),
+        ];
+
+        let calculated_loops = border_edge_loops(&edge_sharing_map);
+
+        assert_eq!(calculated_loops.len(), 2);
+        for edge in correct_loop {
```
yanchith: Why doesn't this match concrete correct data with concrete computed data? There are 2 loops, but only one correct loop?

yanchith (changes requested):
Good job 👍 

I have a few minor improvement suggestions for the implementation code and a few not-so-minor suggestions for the test code, though.

Also a note: you left the word valency instead of sharing i many functions. Is that intentional (to distinguish them) or not?

src/mesh_analysis.rs
janper: no, that would be an error. see the picture - how would the edge loop look like in this case? and what should it look like?
image

src/mesh_analysis.rs
janper: there had to be some decision made. i suggest implementing the function for both, oriented and unoriented edge

src/mesh_analysis.rs
janper: you're right. the only problem is that I don't know which edge comes first.
let's check the length though

src/mesh_analysis.rs
janper: same reason - i don;t know where does the loop start. but let's split it into two vectors and check the lengths

janper:
> Also a note: you left the word valency instead of sharing i many functions. Is that intentional (to distinguish them) or not?

Exactly: the word *valency* is left where we request specifically the valency and removed where we deal with edges (potentially) shared by more faces.

src/mesh_analysis.rs
yanchith: Well that's what I was trying to say. Currently it is implementation dependent: whatever border edge is encountered first changes the result for the rest of them.

yanchith: In this picture I see at least 8 unique edge loops... Not sure if you want to return all or not.

yanchith (approved):

PR #49, opened 2019-10-23
Add transform func

yanchith:
Also, `Value::Float3` is added as a prerequisite.

There is some integration code in `lib.rs` for visual testing. I can remove it prior to merging.

### DIFF
--- a/src/interpreter/ast.rs
+++ b/src/interpreter/ast.rs
@@ -149,6 +149,7 @@ pub enum LitExpr {
     Int(i32),
     Uint(u32),
     Float(f32),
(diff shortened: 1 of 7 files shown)


janper:
Looks okay to me. It was a bit more trouble than expected, wasn't it?

janper (approved):

yanchith:
> It was a bit more trouble than expected, wasn't it?

Not sure what you mean. My previous PR did most of the grunt work (triangle faces -> faces). The only thing here that I had to add is the new `Float3` type, which was rather trivial.

PR #50, opened 2019-10-24
Change meaning of MSAA disabled to 1

yanchith:
Just noticed this in `main.rs` - 1 sample means MSAA is disabled, not 0. Readme is correct though.


janper:
Sure

ondrowan (approved):

PR #51, opened 2019-10-24
Add more project description to the readme

yanchith:
I probably half-assed this, so @janper please check this :))


janper (approved):
It seems sufficient for the moment. 
We'll get a more detailed description (including name explanation) next week. Then I'll also add my comments and descriptions. We should also credit the client, ourselves and resources properly.
Let's keep the Trello task open and revisit it once we have more input.

PR #52, opened 2019-10-24
Compare visually similar geometries

janper:
Check if two mesh geometries are visually similar. They can differ in vertex and normal order and the start of face winding.

### DIFF
--- a/src/mesh_analysis.rs
+++ b/src/mesh_analysis.rs
@@ -7,8 +7,9 @@ use crate::convert::{cast_i32, cast_usize};
 use crate::edge_analysis::EdgeSharingMap;
 use crate::geometry::{Face, Geometry, OrientedEdge, UnorientedEdge};
 
-/// Finds edges with a certain valency in a mesh edge collection
-/// Valency indicates how many faces share the edge
+/// Finds edges with a certain valency in a mesh edge collection 
+///
+/// An edge valency indicates how many faces share the edge.
 fn find_edges_with_valency<'a>(
     edge_sharing: &'a EdgeSharingMap,
     valency: usize,
@@ -28,6 +29,7 @@ fn find_edges_with_valency<'a>(
 }
 
 /// Finds border edges in a mesh edge collection
+///
 /// An edge is border when its valency is 1
 #[allow(dead_code)]
 pub fn border_edges<'a>(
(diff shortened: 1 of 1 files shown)


src/mesh_analysis.rs
```diff
@@ -144,6 +147,78 @@ pub fn triangulated_mesh_genus(vertex_count: usize, edge_count: usize, face_coun
     1 - (cast_i32(vertex_count) - cast_i32(edge_count) + cast_i32(face_count)) / 2
 }
 
+/// Checks if two geometries are visually identical.
+///
+/// The two geometries can
+/// possibly contain the same vertices and normals bound to the same faces, but all
+/// of these can be in different order, including shifts in face winding. This
```
yanchith: > ... including shifts in face winding.

I understood this comment that we consider geometries with opposite winding still identical, a long as they are consistent, but the code doesn't seem to support that. Or?

src/mesh_analysis.rs
```diff
@@ -144,6 +147,78 @@ pub fn triangulated_mesh_genus(vertex_count: usize, edge_count: usize, face_coun
...
+                    && self.vertices.1 == other.vertices.0
+                    && self.vertices.2 == other.vertices.1
+                    && self.normals.0 == other.normals.2
+                    && self.normals.1 == other.normals.0
+                    && self.normals.2 == other.normals.1)
+        }
+    }
+
+    let mut unpacked_faces_1 = g_1.faces().iter().map(|face| match face {
+        Face::Triangle(f) => Some(UnpackedFace {
```
yanchith: I think the `Some` wrapping is unnecessary, same below.

yanchith (changes requested):
Mostly fine, I've just got a simple nit for the implementation and one clarification request for the doc comment.

src/mesh_analysis.rs
janper: no. a face with an opposite winding is a different face. "shifted winding" means (1, 2, 3) = (2, 3, 1) = (3, 1, 2) and that is still the same face

src/mesh_analysis.rs
janper: mind blown!

src/mesh_analysis.rs
yanchith: I understand, I just meant that if I didn't know it beforehand, the doc text wouldn't tell me :)

yanchith (approved):

PR #53, opened 2019-10-25
Operation/separate geometries

janper:
Split isolated chunks of mesh geometry into separate mesh geometries.
Add more verbose assertion messages.

### DIFF
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -89,15 +89,15 @@ impl Geometry {
 
                             assert!(
                                 vertices_range.contains(&v1),
-                                "Faces reference out of bounds position data: Vertex 1 at from_triangle_faces_with_vertices_and_computed_normals"
+                                "Faces reference out of bounds position data"
                             );
                             assert!(
                                 vertices_range.contains(&v2),
-                                "Faces reference out of bounds position data: Vertex 2 at from_triangle_faces_with_vertices_and_computed_normals"
+                                "Faces reference out of bounds position data"
                             );
                             assert!(
                                 vertices_range.contains(&v3),
-                                "Faces reference out of bounds position data: Vertex 3 at from_triangle_faces_with_vertices_and_computed_normals"
+                                "Faces reference out of bounds position data"
                             );
 
                             let face_normal = compute_triangle_normal(
@@ -222,15 +222,15 @@ impl Geometry {
                     let n = triangle_face.normals;
                     assert!(
                         vertices_range.contains(&v.0),
-                        "Faces reference out of bounds position data: Vertex 1 at from_faces_with_vertices_and_normals"
+                        "Faces reference out of bounds position data"
                     );
                     assert!(
                         vertices_range.contains(&v.1),
-                        "Faces reference out of bounds position data: Vertex 2 at from_faces_with_vertices_and_normals"
+                        "Faces reference out of bounds position data"
                     );
                     assert!(
                         vertices_range.contains(&v.2),
-                        "Faces reference out of bounds position data: Vertex 3 at from_faces_with_vertices_and_normals"
+                        "Faces reference out of bounds position data"
                     );
                     assert!(
                         normals_range.contains(&n.0),
@@ -1268,9 +1268,7 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(
-        expected = "Faces reference out of bounds position data: Vertex 3 at from_triangle_faces_with_vertices_and_computed_normals"
-    )]
+    #[should_panic(expected = "Faces reference out of bounds position data")]
     fn test_geometry_from_triangle_faces_with_vertices_and_computed_normals_bounds_check() {
         let (_, vertices) = quad();
         #[rustfmt::skip]
@@ -1311,9 +1309,7 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(
-        expected = "Faces reference out of bounds position data: Vertex 3 at from_faces_with_vertices_and_normals"
-    )]
+    #[should_panic(expected = "Faces reference out of bounds position data")]
     fn test_geometry_from_triangle_faces_with_vertices_and_normals_bounds_check() {
         let (_, vertices, normals) = quad_with_normals();
         #[rustfmt::skip]
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -371,17 +371,23 @@ impl fmt::Display for Geometry {
             .enumerate()
             .map(|(i, f)| format!("{}: {}", i, f))
             .collect();
-        // TODO: Display vectors properly
+        let normals: Vec<_> = self
+            .normals
+            .iter()
+            .enumerate()
+            .map(|(i, n)| format!("{}: ({}, {}, {})", i, n.x, n.y, n.z))
+            .collect();
         write!(
             f,
             "Vertex count: {}, Normal count: {}, Face count: {}\n\
              Vertices: {:?}\n\
-             Normals not formatted yet\n\
+             Normals: {:?}\n\
              Faces: {:?}",
             self.vertices.len(),
             self.normals.len(),
             self.faces.len(),
             vertices,
+            normals,
             faces,
         )
     }
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -358,7 +358,7 @@ impl Geometry {
 }
 
 impl fmt::Display for Geometry {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let vertices: Vec<_> = self
             .vertices
             .iter()
@@ -379,15 +379,12 @@ impl fmt::Display for Geometry {
             .collect();
         write!(
             f,
-            "Vertex count: {}, Normal count: {}, Face count: {}\n\
-             Vertices: {:?}\n\
-             Normals: {:?}\n\
-             Faces: {:?}",
+            "G( V({}): {:?}, N({}): {:?}, F({}): {:?} )",
             self.vertices.len(),
-            self.normals.len(),
-            self.faces.len(),
             vertices,
+            self.normals.len(),
             normals,
+            self.faces.len(),
             faces,
         )
     }
@@ -406,9 +403,9 @@ impl From<TriangleFace> for Face {
 }
 
 impl fmt::Display for Face {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
-            Face::Triangle(face) => write!(f, "T({})", face),
+            Face::Triangle(face) => write!(f, "{}", face),
         }
     }
 }
@@ -490,10 +487,10 @@ impl From<(u32, u32, u32)> for TriangleFace {
 }
 
 impl fmt::Display for TriangleFace {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(
             f,
-            "V: ({}, {}, {}); N: ({}, {}, {})",
+            "T(V: ({}, {}, {}); N: ({}, {}, {}))",
             self.vertices.0,
             self.vertices.1,
             self.vertices.2,
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -88,15 +88,15 @@ impl Geometry {
 
                             assert!(
                                 vertices_range.contains(&v1),
-                                "Faces reference out of bounds position data"
+                                "Faces reference out of bounds position data: Vertex 1 at from_triangle_faces_with_vertices_and_computed_normals"
                             );
                             assert!(
                                 vertices_range.contains(&v2),
-                                "Faces reference out of bounds position data"
+                                "Faces reference out of bounds position data: Vertex 2 at from_triangle_faces_with_vertices_and_computed_normals"
                             );
                             assert!(
                                 vertices_range.contains(&v3),
-                                "Faces reference out of bounds position data"
+                                "Faces reference out of bounds position data: Vertex 3 at from_triangle_faces_with_vertices_and_computed_normals"
                             );
 
                             let face_normal = compute_triangle_normal(
@@ -221,15 +221,15 @@ impl Geometry {
                     let n = triangle_face.normals;
                     assert!(
                         vertices_range.contains(&v.0),
-                        "Faces reference out of bounds position data"
+                        "Faces reference out of bounds position data: Vertex 1 at from_faces_with_vertices_and_normals"
                     );
                     assert!(
                         vertices_range.contains(&v.1),
-                        "Faces reference out of bounds position data"
+                        "Faces reference out of bounds position data: Vertex 2 at from_faces_with_vertices_and_normals"
                     );
                     assert!(
                         vertices_range.contains(&v.2),
-                        "Faces reference out of bounds position data"
+                        "Faces reference out of bounds position data: Vertex 3 at from_faces_with_vertices_and_normals"
                     );
                     assert!(
                         normals_range.contains(&n.0),
@@ -1211,7 +1211,9 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "Faces reference out of bounds position data")]
+    #[should_panic(
+        expected = "Faces reference out of bounds position data: Vertex 3 at from_triangle_faces_with_vertices_and_computed_normals"
+    )]
     fn test_geometry_from_triangle_faces_with_vertices_and_computed_normals_bounds_check() {
         let (_, vertices) = quad();
         #[rustfmt::skip]
@@ -1252,7 +1254,9 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "Faces reference out of bounds position data")]
+    #[should_panic(
+        expected = "Faces reference out of bounds position data: Vertex 3 at from_faces_with_vertices_and_normals"
+    )]
     fn test_geometry_from_triangle_faces_with_vertices_and_normals_bounds_check() {
         let (_, vertices, normals) = quad_with_normals();
         #[rustfmt::skip]
--- a/src/interpreter_funcs.rs
+++ b/src/interpreter_funcs.rs
@@ -6,7 +6,8 @@ use nalgebra::base::Vector3;
 use crate::convert::{cast_u32, cast_u8};
 use crate::geometry;
 use crate::interpreter::{Func, FuncFlags, FuncIdent, ParamInfo, Ty, Value};
-use crate::mesh_smoothing::laplacian_smoothing;
+use crate::mesh_smoothing;
+use crate::mesh_tools;
 use crate::operations::shrink_wrap::{self, ShrinkWrapParams};
 use crate::operations::transform;
 
@@ -169,8 +170,34 @@ impl Func for FuncImplLaplacianSmoothing {
             iterations_unclamped
         };
 
-        let value = laplacian_smoothing(geometry, cast_u8(iterations));
+        let value = mesh_smoothing::laplacian_smoothing(geometry, cast_u8(iterations));
         Value::Geometry(Arc::new(value))
+    }
+}
+
+pub struct FuncImplSeparateIsolatedMeshes;
+impl Func for FuncImplSeparateIsolatedMeshes {
+    fn flags(&self) -> FuncFlags {
+        FuncFlags::PURE
+    }
+    fn param_info(&self) -> &[ParamInfo] {
+        &[ParamInfo {
+            ty: Ty::Geometry,
+            optional: false,
+        }]
+    }
+
+    fn return_ty(&self) -> Ty {
+        Ty::Geometry
+    }
+
+    fn call(&self, args: &[Value]) -> Value {
+        let geometry = args[0].unwrap_geometry();
+
+        let values = mesh_tools::separate_isolated_meshes(geometry);
+
+        // FIXME: This returns a slice of Geometries. Return all of them
+        Value::Geometry(Arc::new(values[0].clone()))
     }
 }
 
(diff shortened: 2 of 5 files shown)


src/geometry.rs
```diff
@@ -356,6 +357,42 @@ impl Geometry {
...
+            .iter()
+            .enumerate()
+            .map(|(i, n)| format!("{}: ({}, {}, {})", i, n.x, n.y, n.z))
+            .collect();
+        write!(
+            f,
+            "Vertex count: {}, Normal count: {}, Face count: {}\n\
+             Vertices: {:?}\n\
+             Normals: {:?}\n\
+             Faces: {:?}",
```
yanchith: For what purpose are you adding this `Display` impl?

I don't agree we should dump it all the data here. It will be useless for anything but the smallest geometries. Note that there is a `Debug` impl for `Geometry` that does dump everything.

Also, when you are implementing `Display` for something, you usually want to indicate somehow what that something is. My suggestion:

```rust
write!(
    f,
    "G( V: {}, N: {}, F {} )",
    self.vertices.len(),
    self.normals.len(),
    self.faces.len(),
);
```

src/geometry.rs
```diff
@@ -221,15 +222,15 @@ impl Geometry {
...
                     );
                     assert!(
                         vertices_range.contains(&v.1),
-                        "Faces reference out of bounds position data"
+                        "Faces reference out of bounds position data: Vertex 2 at from_faces_with_vertices_and_normals"
                     );
                     assert!(
                         vertices_range.contains(&v.2),
-                        "Faces reference out of bounds position data"
+                        "Faces reference out of bounds position data: Vertex 3 at from_faces_with_vertices_and_normals"
```
yanchith: What made you implement this for vertices but not the normals? 😉 Also, I would make the messages shorter - they now exceed 100 columns. Maybe: "Vertex 3 in faces references out of bounds position data". You don't need to include the name of the function - you'll get that in the stack trace.

src/geometry.rs
```diff
@@ -356,6 +357,42 @@ impl Geometry {
     }
 }
 
+impl fmt::Display for Geometry {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
yanchith: Why the `<'_>`? Shouldn't be necessary. Also in the other display impls.

src/geometry.rs
```diff
@@ -368,6 +405,14 @@ impl From<TriangleFace> for Face {
     }
 }
 
+impl fmt::Display for Face {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Face::Triangle(face) => write!(f, "T({})", face),
```
yanchith: This doesn't compose. What you want to do is:

`Face::Triangle(face) => write!(f, "{}", face),` in here, and include the "T" in the `Display` impl for `TriangleFace`:

```rust
write!(
    f,
    "T(V: ({}, {}, {}); N: ({}, {}, {}))",
    self.vertices.0,
    self.vertices.1,
    self.vertices.2,
    self.normals.0,
    self.normals.1,
    self.normals.2,
)
```

src/mesh_tools.rs
```diff
@@ -0,0 +1,279 @@
+use std::collections::HashMap;
+
+use smallvec::SmallVec;
+
+use crate::convert::{cast_u32, cast_usize};
+use crate::geometry::Geometry;
+use crate::mesh_topology_analysis::face_to_face_topology;
+
+fn crawl_faces(
```
yanchith: Private functions should usually be positioned below the public functions.

Also... I don't believe it needs to be recursive at all once you remove all the copies and allocations you don't need.

src/mesh_tools.rs
```diff
@@ -0,0 +1,279 @@
...
+    let face_to_face = face_to_face_topology(geometry);
+    let faces = geometry.faces();
+    let vertices: Vec<_> = geometry.vertices().to_vec();
+    let normals: Vec<_> = geometry.normals().to_vec();
+
+    let mut available_faces: Vec<u32> = face_to_face.keys().map(|key| cast_u32(*key)).collect();
+
+    let mut patches: Vec<Geometry> = Vec::new();
+
+    while !available_faces.is_empty() {
```
yanchith: Can't this be change into `while let` with everything happening in the condition?

src/mesh_tools.rs
```diff
@@ -0,0 +1,279 @@
...
+}
+
+/// Crawls the geometry to find continuous patches of geometry.
+/// Returns a vector of new separated geometries.
+#[allow(dead_code)]
+pub fn separate_isolated_meshes(geometry: &Geometry) -> Vec<Geometry> {
+    let face_to_face = face_to_face_topology(geometry);
+    let faces = geometry.faces();
+    let vertices: Vec<_> = geometry.vertices().to_vec();
+    let normals: Vec<_> = geometry.normals().to_vec();
```
yanchith: Vertices and normals don't need to be vectors. Slices suffice. You can use `.to_vec()` down below to allocate.

Also, you have shadowing with `faces` down the line. Le't not define these at all and use `geometry.faces()` and friends. Alternatively: provide non-shadowing name for faces.

src/mesh_tools.rs
```diff
@@ -0,0 +1,279 @@
...
+                &face_to_face,
+                &mut available_faces,
+                &mut separate_faces,
+            );
+        }
+
+        let faces: Vec<_> = separate_faces
+            .iter()
+            .map(|face_index| faces[cast_usize(*face_index)])
+            .collect();
```
yanchith: Let's not `collect` when iterator is enough.

src/mesh_tools.rs
```diff
@@ -0,0 +1,279 @@
...
+
+        patches.push(
+            Geometry::from_faces_with_vertices_and_normals_remove_orphans(
```
yanchith, line 20: Hmm, can this be done more efficiently than letting is sort itself out later? May be fine for now, but this may not perform very well on bigger data.

src/mesh_tools.rs
```diff
@@ -0,0 +1,279 @@
...
+        ];
+
+        let vertex_normals = vec![n(0.0, 0.0, 1.0)];
+
+        let faces = vec![TriangleFace::new_separate(0, 1, 2, 0, 0, 0)];
+
+        Geometry::from_triangle_faces_with_vertices_and_normals(faces, vertices, vertex_normals)
+    }
+
+    pub fn cube_sharp_var_len(position: [f32; 3], scale: f32) -> Geometry {
```
yanchith: Why is this copied here? Is there a difference between this function and the one defined in `geometry.rs`? If so, it should be documented. If not, let's remove it from here.

src/mesh_tools.rs
```diff
@@ -0,0 +1,279 @@
...
+            .iter()
+            .copied()
+            .filter(|n| available_faces.iter().any(|a_f| a_f == n))
+            .collect();
+        let af: Vec<_> = available_faces
+            .iter()
+            .copied()
+            .filter(|f| neighbors.iter().all(|n| n != f))
+            .collect();
+        *available_faces = af;
```
yanchith: This is needless allocation for no good reason. Try to find out if there is some kind of swap_remove for `&mut` iterators over vectors and ideally use that.

src/mesh_tools.rs
```diff
@@ -0,0 +1,279 @@
...
+    separate_faces: &mut Vec<u32>,
+) {
+    separate_faces.push(current_face_index);
+
+    if let Some(all_neighbors) = face_to_face.get(&current_face_index) {
+        let neighbors: Vec<u32> = all_neighbors
+            .iter()
+            .copied()
+            .filter(|n| available_faces.iter().any(|a_f| a_f == n))
+            .collect();
```
yanchith: This doesn't need to be a `Vec` if you only iterate it here and then discard it. Keep it an iterator.

src/mesh_analysis.rs
```diff
@@ -223,7 +223,7 @@ pub fn are_visually_identical(g_1: &Geometry, g_2: &Geometry) -> bool {
     g_1.faces().len() == g_2.faces().len()
         && g_1.vertices().len() == g_2.vertices().len()
         && g_1.normals().len() == g_2.normals().len()
-        && unpacked_faces_1.all(|f| unpacked_faces_2.any(|g| f == g))
+        && unpacked_faces_1.all(|f| unpacked_faces_2.clone().any(|g| f == g))
```
yanchith, line 259: Does this means the tests for this function were insufficient? Maybe we should add more tests that cover the bug that was just fixed? Ideally separate this out of the PR and let's do it first.

src/mesh_tools.rs
```diff
@@ -0,0 +1,279 @@
...
+/// Crawls the geometry to find continuous patches of geometry.
+/// Returns a vector of new separated geometries.
+#[allow(dead_code)]
+pub fn separate_isolated_meshes(geometry: &Geometry) -> Vec<Geometry> {
+    let face_to_face = face_to_face_topology(geometry);
+    let faces = geometry.faces();
+    let vertices: Vec<_> = geometry.vertices().to_vec();
+    let normals: Vec<_> = geometry.normals().to_vec();
+
+    let mut available_faces: Vec<u32> = face_to_face.keys().map(|key| cast_u32(*key)).collect();
```
yanchith: I'd name both this and separate_faces: `available_face_indices` and `separate_face_indices`.

yanchith (changes requested):
This is just a preliminary review. The PR was a bit hard to read and understand, so we will have to come back to it.

First of all: I don't think the crawling needs to be recursive at all. It may be that all the cloning forced you to go the recursive way, but the result is both unreadable and inefficient.

Also the `Display` impls seems unrelated to the current PR (or maybe I just didn't notice). Let's split them out.

See inline comments for more concrete suggestions.

src/geometry.rs
janper: I agree it might be irrelevant to the rest of the PR but it was a great tool for debugging. The `Debug` dump wasn't easy enough to read, whereas this was and helped me to find errors in both, the functions and their tests. I believe this will be a very useful way of tracking possible errors in the upcoming geometry modifiers.

I don't mind splitting the current PR into two, just have no idea how.

src/geometry.rs
janper: I got the error in the test report (i didn't get it for normals, so I've forgotten) and there you don't see the stacktrace, afaik. It was also very useful for debugging.

src/geometry.rs
janper: I've got the line from Stack Overflow ;)

src/mesh_tools.rs
janper: ~not sure I understand~

src/mesh_tools.rs
janper: I thought about it but at the end of the day you'll need to perform the exact same logic here anyway. So reusing the existing function seemed fair.

src/mesh_analysis.rs
janper: not sure why do the original tests pass but they shoudn't. i'll add more tests catching this bug.

src/mesh_tools.rs
janper: Okay, I can imagine renumbering the faces as you crawl over it. It will be a hell of an algorithm, though! Not sure if it is even better optimized as you might need some supplemental data structures to mark used vertices and normals.

src/mesh_tools.rs
janper: Spent about 30 minutes on this, don't know how without causing some error somewhere else.
`&` `*` `clone()` `copied()`

src/mesh_analysis.rs
janper: tests now prove it works (wouldn't pass with the former version)

src/geometry.rs
janper: Meh. Let's keep it the way it was. We'd have to be way more verbose to make this change useful.

src/mesh_tools.rs
yanchith: It's alright. I know it's better for our time to re-use. Was just wandering.

src/mesh_tools.rs
janper: How else would you do it?

src/mesh_tools.rs
ondrowan: If this is going to happen, create a Trello task for it and let's resolve it as wontfix for now.

src/mesh_tools.rs
```diff
@@ -0,0 +1,200 @@
...
+            assert!(mesh_analysis::are_visually_identical(
+                &calculated_geometries[1],
+                &geometry_triangle_correct
+            ));
+            assert!(mesh_analysis::are_visually_identical(
+                &calculated_geometries[0],
+                &geometry_island_correct
+            ));
+        }
+    }
```
ondrowan: What should this `if..else` do? Only one of its branches runs anyway. The second one also asserts the exact thing that's tested in `if` so it probably doesn't make sense.

src/mesh_tools.rs
janper: I actually still have some doubts if there is a more efficient way. I can imagine saving one pass through the filtered faces and copying (do we?) the vertices and normals as arguments into the function. The rest would be exactly the same, only the code will be way more complex.

src/mesh_tools.rs
```diff
@@ -0,0 +1,200 @@
...
+        for neighbor_index in neighbor_indices {
+            crawl_faces(
+                neighbor_index,
+                face_to_face,
+                available_face_indices,
+                separate_face_indices,
+            );
+        }
+    }
+}
```
ondrowan, line 57: I have a played a bit with optimization here and the outline looks like this: https://gist.github.com/ondrowan/d635981610ed0ff9394f56c3a81f6809

It uses depth-first search which removes the need for recursion and vector allocations. It's however not finished and there is space for improvement, as seen in notes.

ondrowan (changes requested):

src/mesh_tools.rs
janper: it's not the same thing! it's comparing either 0<->0, 1<->1 or 0<->1, 1<->0

ondrowan: I'm a bit lost now. What I wanted to say is that only the first branch of the condition ever runs since result of the first `mesh_analysis::are_visually_identical` is always the same.

janper: true, but at the beginning i didn't know which branch is actually executed. and i still don't know, tbh. i can try and figure and remove the other one but should i? the point here is that i don;t know which separate mesh comes first, so i'm ready for both cases.

janper: the important thing is that one is equal to one of the test samples and the other to the other one.

ondrowan: Well, tests should be deterministic, so let's split this into 2 separate tests for each of the branches.

janper: okay, i'll do that

janper: actually, after many tests, it seems that the order of the two separate meshes changes each time. @yanchith mentioned some "seeds" used by HashSets - maybe that's why it changes each run?

janper:
```
// NOTE: very random capacity, I bet there's some clever math to figure this out
    let mut index_stack = Vec::with_capacity(face_to_face.len() / 2);
```
This seems about right. In a proper mesh, in the worst case scenario, with each iteration two more faces are added and one removed from the stack. This is happening until a half for the distinct mesh has been crowled _(is that so?)_ and then the size decreases. Therefore it intuitively seems like the potential maximum capacity of the stack is half of the biggest mesh chunk, which is half of the entire mesh if there is just one chunk. 
Long story short, `face_to_face.len() / 2` might not only be a good estimation but also a real maximum capacity.
Unless i'm wrong ;)

janper:
```
// NOTE: Worst case scenario capacity (everything is connected)
    let mut connected_face_indices = HashSet::with_capacity(face_to_face.len());
```
If this really makes any difference, you can pass the number of already processed faces from the calling function and subtract it from the `face_to_face.len()`. But then, does it really affect performance?

janper:
```
// NOTE: I have a hunch we can somehow get rid of this set
    let mut available_face_indices: HashSet<u32> = face_to_face.keys().cloned().collect();
```
It will be eventually needed one way or another. You can also compute it as a difference between all faces and already used faces, which doesn't sound like a performance improvement. 
However, instead of adding and removing items from this set, you can just have an availability flag map (boolean pattern). that way there are no changes to the structure itself (and it can be also reused) but only changes to the items. Not sure if ti does improve performance though.
Another thing is the HashSet itself - any list would be sufficient for the logic and if it is actually faster, we can switch to that. According to what @yanchith told me about sets, it seems like it might be actually quite fast for random access?

src/mesh_tools.rs
yanchith: If I understand correctly, the test is deterministic with the branch, as it is now. I agree that it is a bit non-obvious, but it seems correct to me.

src/interpreter_funcs.rs
```diff
@@ -169,18 +170,45 @@ impl Func for FuncImplLaplacianSmoothing {
...
-        let value = laplacian_smoothing(geometry, cast_u8(iterations));
+        let value = mesh_smoothing::laplacian_smoothing(geometry, cast_u8(iterations));
         Value::Geometry(Arc::new(value))
     }
 }
 
+pub struct FuncImplSeparateIsolatedMeshes;
+impl Func for FuncImplSeparateIsolatedMeshes {
+    fn flags(&self) -> FuncFlags {
+        FuncFlags::empty()
```
yanchith: If no flag is provided let's omit the implementation altogether - there is a default impl in the trait.

src/interpreter_funcs.rs
```diff
@@ -169,18 +170,45 @@ impl Func for FuncImplLaplacianSmoothing {
...
+        Ty::Geometry
+    }
+
+    fn call(&self, args: &[Value]) -> Value {
+        let geometry = args[0].unwrap_geometry();
+
+        let values = mesh_tools::separate_isolated_meshes(geometry);
+
+        // FIXME: This returns a slice of Geometries. Return all of them
+        Value::Geometry(Arc::new(values[0].clone()))
```
yanchith, line 197: I think the `.clone()` here can be removed.

src/mesh_tools.rs
```diff
@@ -0,0 +1,196 @@
...
+use crate::convert::cast_usize;
+use crate::geometry::Geometry;
+use crate::mesh_topology_analysis::face_to_face_topology;
+
+/// Crawls the geometry to find continuous patches of geometry.
+/// Returns a vector of new separated geometries.
+#[allow(dead_code)]
+pub fn separate_isolated_meshes(geometry: &Geometry) -> Vec<Geometry> {
+    let face_to_face = face_to_face_topology(geometry);
+    let mut available_face_indices: HashSet<u32> = face_to_face.keys().cloned().collect();
```
yanchith: Rust nit: use `.copied()` instead of `.cloned()`. Does not make any runtime difference here, but expresses intent more clearly (`.copied()` is not available for items that are `!Copy`, but `.cloned()` is).

src/mesh_tools.rs
yanchith: I am totally fine with the current approach, was just musing in my sickly delirium. I would even skip the trello task - if this becomes a problem, we'll see it on our profiling and fix it.

src/mesh_tools.rs
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    patches
+}
+
+fn crawl_faces(
+    start_face_index: u32,
+    face_to_face: &HashMap<u32, SmallVec<[u32; 8]>>,
+) -> HashSet<u32> {
+    let mut index_stack = Vec::with_capacity(face_to_face.len() / 2);
```
yanchith: We can benchmark this to be sure, but I have a hunch that amortized re-allocation will be better than a bad estimate. I'd keep this a `Vec::new()`, or even better: `vec![start_face_index]`.

src/mesh_tools.rs
```diff
@@ -0,0 +1,196 @@
...
+    index_stack.push(start_face_index);
+
+    let mut connected_face_indices = HashSet::with_capacity(face_to_face.len());
```
yanchith: Once again... let's default to `HashSet::new()`, if we can't estimate the capacity. (Or provide a benchmark that proves the point here, but that's going to be difficult).

src/mesh_tools.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    while !index_stack.is_empty() {
```
yanchith: Rust nit: can be `while let Some(current_face_index) = index_stack.pop() {` instead.

src/mesh_tools.rs
```diff
@@ -0,0 +1,196 @@
...
+        let current_face_index = index_stack.pop().unwrap();
+
+        if !connected_face_indices.contains(&current_face_index) {
+            connected_face_indices.insert(current_face_index);
```
yanchith: Rust nit: `HashSet::insert` returns whether the call caused the insertion or not, so this can be simplified to save some electricity:
```rust
if connected_face_indices.insert(current_face_index) {
    // ...
}
```
https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.insert

yanchith (changes requested):
Wow! This PR progressed very far ❤️ 

I have a few nits inline and a general question:

I now see that the behavior is not pure, which is a shame... Can we think of something to make it pure or do we leave that for the future?

src/interpreter_funcs.rs
janper: I suggest to resolve the purity of this later. We can leave a `FIXME`

src/interpreter_funcs.rs
janper: move occurs because value has type `geometry::Geometry`, which does not implement the `Copy` trait

src/interpreter_funcs.rs
yanchith: I agree with adding the task or FIXME to do this, but that is not what this comment was about. You can remove the whole function so that the default is used.

yanchith:
@janper almost there. Please resolve merge conflicts and remove the func flags for now. I have created the trello task to resolve purity later: https://trello.com/c/SYBG0nRl/166-find-a-way-to-make-func-separateisolatedmeshes-pure

yanchith (approved):

ondrowan (approved):

PR #54, opened 2019-10-25
Stop Laplacian smoothing with anchors when the geometry becomes stable

janper:
The Laplacian smoothing with anchors can be used to paramterize a mesh patch into a flat disc. Once the edge loop is transformed into the circumference of the target disc, the iterative algorithm of Laplacian smoothing is applied until the mesh reaches a stable spatial configuration.

This function will be later used in more complex operations, therefore it's not exposed as a distinct operation.

### DIFF
--- a/src/interpreter_funcs.rs
+++ b/src/interpreter_funcs.rs
@@ -168,7 +168,7 @@ impl Func for FuncImplLaplacianSmoothing {
         let iterations = args[1].unwrap_uint();
         let vertex_to_vertex_topology = mesh_topology_analysis::vertex_to_vertex_topology(geometry);
 
-        let (g, _, _) = mesh_smoothing::laplacian_smoothing_with_anchors_full(
+        let (g, _, _) = mesh_smoothing::laplacian_smoothing(
             geometry,
             vertex_to_vertex_topology,
             cmp::min(255, iterations),
--- a/src/interpreter_funcs.rs
+++ b/src/interpreter_funcs.rs
@@ -170,7 +170,7 @@ impl Func for FuncImplLaplacianSmoothing {
         let (g, _, _) = mesh_smoothing::laplacian_smoothing_with_anchors_full(
             geometry,
             vertex_to_vertex_topology,
-            cast_u32(cmp::min(255, iterations)),
+            cmp::min(255, iterations),
             &[],
             false,
         );
--- a/src/interpreter_funcs.rs
+++ b/src/interpreter_funcs.rs
@@ -1,12 +1,14 @@
+use std::cmp;
 use std::collections::HashMap;
 use std::sync::Arc;
 
 use nalgebra::base::Vector3;
 
-use crate::convert::{cast_u32, cast_u8};
+use crate::convert::cast_u32;
 use crate::geometry;
 use crate::interpreter::{Func, FuncFlags, FuncIdent, ParamInfo, Ty, Value};
-use crate::mesh_smoothing::laplacian_smoothing;
+use crate::mesh_smoothing;
+use crate::mesh_topology_analysis;
 use crate::operations::shrink_wrap::{self, ShrinkWrapParams};
 use crate::operations::transform;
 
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -2,6 +2,9 @@ pub use crate::renderer::{GpuBackend, Msaa, PresentMode};
 
 use std::time::{Duration, Instant};
 
+#[macro_use]
+extern crate approx;
```
ondrowan: There's no need to do this. Just prefix `relative_eq` with `approx::` and you won't need to import anything.

src/mesh_smoothing.rs
```diff
@@ -39,16 +39,56 @@ pub fn laplacian_smoothing_with_anchors(
...
+fn laplacian_smoothing_with_anchors_full(
+    geometry: &Geometry,
+    max_iterations: u8,
+    fixed_vertex_indices: &[u32],
+    stop_when_stable: bool,
+) -> Geometry {
+    if max_iterations == 0 {
         return geometry.clone();
     }
+    println!("Stop when stable: {}", stop_when_stable);
```
ondrowan: This should be most likely removed.

src/mesh_smoothing.rs
```diff
@@ -60,9 +100,19 @@ pub fn laplacian_smoothing_with_anchors(
                     average_position += geometry_vertices[cast_usize(*neighbor_index)].coords;
```
ondrowan, line 65: Not part of this PR, but I found that closure in condition can be simiplified to `|i| i != current_vertex_index` since `current_vertex_index` is already `u32`. `cast_u32` import won't be needed as well.

src/mesh_smoothing.rs
```diff
@@ -60,9 +100,19 @@ pub fn laplacian_smoothing_with_anchors(
...
                 }
                 average_position /= neighbors_indices.len() as f32;
+                if stop_when_stable {
+                    stable &= relative_eq!(
```
ondrowan: Does this `&` have any special meaning here? Is there a case when `stable` would be `true` and `false` from `relative_eq` would invalidate it?

ondrowan (changes requested):

src/mesh_smoothing.rs
janper: yup, that's just forgotten

src/mesh_smoothing.rs
janper: sure, if some vertices are already stable and some are not, then the iterations need to continue

ondrowan (dismissed):
It looks good to me, but let's wait for @yanchith to review this as well.

yanchith (dismissed):
❤️

janper:
Now I think maybe i should refactor it so that the operation itself runs the logic iteratively. It would be then possible to test the function properly plus in the future when we have the iterative operations with feedback, there will be no need to refactor this. Also the log will be nicer. Wouldn't you mind one more refactor?

yanchith:
@janper 

> Wouldn't you mind one more refactor?

Have no strong opinion on this, but the default for me is to do work later, once we really require it.

src/interpreter_funcs.rs
```diff
@@ -162,15 +164,18 @@ impl Func for FuncImplLaplacianSmoothing {
...
-
-        let value = laplacian_smoothing(geometry, cast_u8(iterations));
-        Value::Geometry(Arc::new(value))
+        let iterations = args[1].unwrap_uint();
+        let vertex_to_vertex_topology = mesh_topology_analysis::vertex_to_vertex_topology(geometry);
+
+        let (g, _, _) = mesh_smoothing::laplacian_smoothing_with_anchors_full(
+            geometry,
+            vertex_to_vertex_topology,
+            cast_u32(cmp::min(255, iterations)),
```
yanchith: This is already a `u32`, no need to cast.

src/mesh_smoothing.rs
```diff
@@ -62,30 +81,41 @@ pub fn laplacian_smoothing_with_anchors(
...
+        max_iterations,
+        fixed_vertex_indices,
+        true,
+    );
+    g
 }
 
-fn laplacian_smoothing_with_anchors_full(
+/// returns (smooth_geometry: Geometry, executed_iterations: u32, stable: bool)
+pub fn laplacian_smoothing_with_anchors_full(
```
yanchith: I think there's now way too many of the `laplacian_smoothing_*` functions. Let's keep just the most generic one (and call it just `laplacian_smoothing`) and use that everywhere. Also make sure its docs are up to date and correctly formatted. Let's keep the tests for the various variants, just inline the function with concrete parameters set there.

src/mesh_smoothing.rs
```diff
@@ -36,40 +47,111 @@ pub fn laplacian_smoothing(geometry: &Geometry, iterations: u8) -> Geometry {
...
+    geometry: &Geometry,
+    vertex_to_vertex_topology: HashMap<u32, SmallVec<[u32; 8]>>,
+    max_iterations: u32,
+    fixed_vertex_indices: &[u32],
+    stop_when_stable: bool,
+) -> (Geometry, u32, bool) {
+    if max_iterations == 0 {
+        return (geometry.clone(), 0, stop_when_stable);
```
yanchith: Consistency nit: I think `false` is more correct than `stop_when_stable` here. The changes could not have stabilized if there were no changes.

yanchith (changes requested):
Nice refactor 👍  but please address inline comments.

src/mesh_smoothing.rs
yanchith: @janper Almost done, but you still forgot the rename to `laplacian_smoothing`

yanchith (approved):

PR #55, opened 2019-10-26
Distinct between identical, similar and visually similar mesh geometries

janper:
This is a more rigorous comparison of two mesh geometries.

Let's consider two mesh geometries IDENTICAL when the data structure looks the same and they look the same in the memory. A simple Eq function should suffice for this comparison.

Let's consider two mesh geometries SIMILAR when they look exaclty the same, have the same features and behave the same but don't necesarilly look the same in memory. Such mesh geometries share the same count of vertices, normals and faces but they can be sotred in different orders. Faces may reference the vertices in different order but identical winding (0->1->2 = 1->2->0 = 2->0->1).

Let's consider two mesh geometries VISUALLY SIMILAR when they look exaclty the same, but neither necesarilly have the same features, nor behave the same, nor look the same in memory. This is usually the case when faces don't reuse (share) the same vertex more times but rather reference an individual vertex with the same coordinates for each of the faces. This can be the case for all the faces/vertices or just some. Such meshes look exactly the same but they are treated differently by this software's functions and have different features - usually at least one of them is not watertight. Not reusing normals doesn't result in different behavior or features but for the sake of this distinction two meshes with different counts of normals are also considered VISUALLY SIMILAR but not SIMILAR.

This distinction is important when using some of our mesh tools. separate_isolated_meshes applied to a mesh geometry, which consists of two or more isolated chunks sharing the same normals, yields two geometires with some of the normals copied to both geometries. Using join_meshes on these separated chings later on doesn't yield the original mesh but rather one with the originally shared normals now copied as individual elemnts referenced by only one chunk. Such joined mesh is VISUALLY SIMILAR to the original one but is not SIMILAR. The same would apply to vertices when a mesh geometry is being chopped into more pieces and then joined again or to unwelded and welded versions of the same geometry.

### DIFF
--- a/src/mesh_analysis.rs
+++ b/src/mesh_analysis.rs
@@ -149,19 +149,54 @@ pub fn triangulated_mesh_genus(vertex_count: usize, edge_count: usize, face_coun
     1 - (cast_i32(vertex_count) - cast_i32(edge_count) + cast_i32(face_count)) / 2
 }
 
-/// Checks if two geometries are visually identical
+/// Checks if two geometries are similar
 ///
-/// Two mesh geometries are identical when the position of each vertex in one
-/// mesh geometry matches a position of exactly one vertex in the other mesh
-/// geometry, when the direction of each normal in one mesh geometry matches a
-/// direction of exactly one normal in the other mesh geometry and each face in
-/// one mesh geometry refers vertices with the same position and normals with
-/// the same direction, both in the same circular order, as exactly one face in
-/// the other mesh geometry. The indices (order in which they are stored) of
-/// vertices, normals and faces can differ but as long as the previous
-/// conditions are met, the mesh geometries are identical.
+/// Two mesh geometries are similar when the position of each vertex in one mesh
+/// geometry matches a position of some vertex in the other mesh geometry, when
+/// the direction of each normal in one mesh geometry matches a direction of
+/// some normal in the other mesh geometry and each face in one mesh geometry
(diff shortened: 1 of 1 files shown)


src/mesh_analysis.rs
```diff
@@ -149,19 +149,54 @@ pub fn triangulated_mesh_genus(vertex_count: usize, edge_count: usize, face_coun
...
+/// some normal in the other mesh geometry and each face in one mesh geometry
+/// refers vertices with the same position and normals with the same direction,
+/// both in the same circular order, as exactly one face in the other mesh
+/// geometry.
+///
+/// The indices (order in which they are stored) of vertices, normals and faces
+/// can differ but as long as the previous conditions are met, and the count of
+/// vertices, normals and faces are identical, the mesh geometries are similar.
+///
+/// The mesh geometries are not necessarily identical in the memory but they
```
yanchith: Nit: "in the memory" -> "in memory", also in the comment below

src/mesh_analysis.rs
```diff
@@ -281,6 +314,31 @@ mod tests {
         Geometry::from_triangle_faces_with_vertices_and_normals(faces, vertices, vertex_normals)
     }
 
+    pub fn quad_not_watertight_with_normals() -> Geometry {
```
yanchith: Let's indicate which vertex/normal is extra

yanchith (dismissed):
👍 but please address comments

yanchith (approved):

PR #57, opened 2019-10-30
Add ability to draw mesh edges and add more drawing modes

yanchith:
This adds 4 drawing modes and the ability to switch between them:

- Shaded:          What we had until now
- Edges:           Draws just the edges of the mesh
- ShadedEdges:     Draws both the shaded view and the visible edges
- ShadedEdgesXray: Draws both the shaded view and all the edges, even
                   occluded

The edge drawing is done with a shader trick, which seemed like the
simplest solution initially (required the least re-architecting), but
it might not be enough for our later needs. See `viewport.frag` for
more.

The scene renderer is refactored internally to support
(very trivial and mostly incorrect) rendering of transparent objects:
`Edges` and `ShadedEdgesXray` use the new transparent render pipeline.

Both rendering of faces and edges is done by the same shader, with
flags used to enable rendering of one or the other, or both (for
`ShadedEdges`). `ShadedEdgesXray` is rendered in two passes - first
`Shaded`, then `Edges`.

**Reviewers:**

- Please also test visually on your machines (this paint job was done on an aging macOS MBP).
- Also, I silenced ~the lamb~ `clippy::too_many_arguments` on two occasions. I don't believe the lint is wrong, but GPU programming with `wgpu` is a bit verbose and listening to clippy would IMO make the code less readable in this case.


src/renderer/scene_renderer.rs
```diff
@@ -719,6 +876,115 @@ fn apply_wgpu_correction_matrix(projection_matrix: &Matrix4<f32>) -> Matrix4<f32
     wgpu_correction_matrix * projection_matrix
 }
 
+fn barycentric_sequence_iter() -> impl Iterator<Item = u32> {
```
ondrowan, line 880: This could use a docstring.

src/renderer/mod.rs
```diff
@@ -1,12 +1,12 @@
...
 use std::fmt;
 
 use nalgebra::base::Matrix4;
 
 use self::imgui_renderer::{ImguiRenderer, ImguiRendererOptions};
 use self::scene_renderer::{
-    SceneRenderer, SceneRendererAddGeometryError, SceneRendererClearFlags, SceneRendererOptions,
+    SceneRenderer, SceneRendererAddGeometryError, SceneRendererClearFlags,
+    SceneRendererDrawGeometryMode, SceneRendererGeometry, SceneRendererGeometryId,
+    SceneRendererOptions,
```
ondrowan: Not fully part of this PR and I should've pointed it out earlier, but I'm not a fan of this kind of naming. Besides `SceneRenderer` and maybe errors, none of these should contain `SceneRenderer` prefix. It should be obvious it comes from there and if it's not, users can alias it.

ondrowan (changes requested):
I've tried it on Windows machine with both vulkan and D3D12 and it seems to work well.

Unrelated to this PR, I've found out that D3D12 rendering makes the controls quite unresponsive and it seems the higher value for MSAA I choose, the worse it is. It shows I have good framerate, but if I try to move camera it responds with a lag. If MSAA is 1, it happens only for a second of two after starting the app. If it's higher, it's unusable. It happens only in debug mode though, so might not be a cause for panic.

src/renderer/mod.rs
yanchith: Hmm there is some merit to what you say. I will ponder it.

yanchith: Just looked elsewhere in the codebase, errors do not share the prefix, so I'll remove it from them as well.

yanchith:
> Unrelated to this PR, I've found out that D3D12 rendering makes the controls quite unresponsive and it seems the higher value for MSAA I choose, the worse it is.

I think we'll need to have a pass on dealing with the winit/wgpu bugs at some point, documenting all of them, trying to update and discerning what changed. Worst case we might need to get our hands dirty and fix them upstream.

ondrowan (approved):

janper (approved):
great! 
image

PR #58, opened 2019-10-30
Join two mesh geometries into one

janper:
Joins two mesh geometries into one

Concatenates vertex and normal slices, while keeping the first mesh geometry's element indices intact and second geometry's indices offset by the length of the respective elements. Reuses first mesh geometry's faces and recomputes the second mesh geometry's faces to match new indices of its elements.


src/interpreter_funcs.rs
```diff
@@ -203,6 +203,38 @@ impl Func for FuncImplSeparateIsolatedMeshes {
     }
 }
 
+pub struct FuncImplJoinMeshes;
+impl Func for FuncImplJoinMeshes {
+    fn flags(&self) -> FuncFlags {
+        FuncFlags::PURE
+    }
+    fn param_info(&self) -> &[ParamInfo] {
```
yanchith, line 212: Missing a newline here

yanchith (commented):
Nice and simple!

Don't we also want to also add a test that doesn't rely on `are_visually_similar` here? This seems exactly the simple and straightforward functionality we could cover with simple tests covering the usual edges cases 0, 1 and many.

janper:
isn't `is_visually_similar` valid when tested? i actually made it mainly for testing purposes, plus i don't know how else could we test it.

and yes, i agree we should test it on more types of geometry

yanchith:
> isn't is_visually_similar valid when tested? i actually made it mainly for testing purposes, plus i don't know how else could we test it.

It is, but my point was it is a too generic test that doesn't tell you what exactly is wrong, only that something is. The concrete logic here is so simple that more rigorous tests could easily be written.

EDIT: we can decide not to write them (or maybe not right now), but let that be a conscious choice and let's not pretend the current testing is all we need.

ondrowan:
This doesn't deal with duplicate vertices, does it? Should UI allow to use the same geometry for both its arguments?

janper:
> This doesn't deal with duplicate vertices, does it? Should UI allow to use the same geometry for both its arguments?

Good question. It opens a couple of issues somewhere else, too.

**No, this doesn't deal with duplicate vertices** as it could result in implicit (and therefore potentially unwanted) welding. 

It's hard to come up with an excuse to forbid referencing the same geometry twice, yet doing so would result in an unusable geometry in the following steps. But then, should we test every chance of creating a geometry, which might become invalid after some of the upcoming operations? So at least for now, let's **allow the UI to use the same geometry twice**.

Another question arises though - shouldn't we somehow deal with duplicate faces, too? There are no duplicate faces when joining a mesh with itself, but once you weld it, all the faces will be present twice. Should we remove them? When? During the welding or explicitly anytime? Or should we check if such a thing might happen before welding and warn?

Once we support voxels (and we will), the easiest thing to do when converting voxels into a mesh, is to generate a mesh cube for each voxel, weld them and only then remove both of the faces, which are present twice. It might not be the most efficient way, but anything else seems to be terribly complicated.

yanchith:
> let's allow the UI to use the same geometry twice.

I think so too!

> Another question arises though - shouldn't we somehow deal with duplicate faces, too?

Interesting, do we want to clean that up manually/automatically? I too don't see why we would want to keep duplicate faces around.

janper:
> > Another question arises though - shouldn't we somehow deal with duplicate faces, too?
> 
> Interesting, do we want to clean that up manually/automatically? I too don't see why we would want to keep duplicate faces around.

I don't know. This is a fundamental decision - either we are creating an end-user software or a scripting platform. The first means automating as much as possible (in this case removing duplicate faces in each operation that can possibly produce them - so far just welding and possibly also mesh constructors themselves), in the latter case we should have an explicit cleanup operation. 

In fact, I think we should be doing this automatically.

janper:
> Nice and simple!
> 
> Don't we also want to also add a test that doesn't rely on `are_visually_similar` here? This seems exactly the simple and straightforward functionality we could cover with simple tests covering the usual edges cases 0, 1 and many.

I'm actually not sure how to make these tests in this case?

yanchith:
@janper 

> I'm actually not sure how to make these tests in this case?

Test cases:

- simple geometry + empty geometry
- empty geometry + simple geometry
- simple geometry + another simple geometry

Since all we are doing is concatenating arrays an re-indexing, this should be well cover-able by tests, and tells us much more about correctness than visual similarity. Let's use the visual similarity comparison as an escape hatch where more rigorous tests are impractical or very hard to write.

yanchith:
> In fact, I think we should be doing this automatically.

I have the same the same intuition, but am undecided where this should happen. If we do it in the constructor, we'll feel no pressure to deal with things correctly and always let the constructor sort stuff out (a precedence I'd rather not set). Soooo... maybe let's do it in the operation itself?

janper:
> > In fact, I think we should be doing this automatically.
> 
> I have the same the same intuition, but am undecided where this should happen. If we do it in the constructor, we'll feel no pressure to deal with things correctly and always let the constructor sort stuff out (a precedence I'd rather not set). Soooo... maybe let's do it in the operation itself?

Let's do it in the operation. However, so far the only operation that might cause double faces is weld, which is in a different PR ;)

src/mesh_tools.rs
```diff
@@ -56,6 +59,47 @@ fn crawl_faces(
     connected_face_indices
 }
 
+/// Joins two mesh geometries into one
```
yanchith: Nit: Missing a period at the end of the sentence.

yanchith (commented):
I think this is almost there, just fix the 2 small nits (one is still left over from previous review)

src/mesh_tools.rs
janper: Yeah, I keep adding the periods when you two request them, but in fact it's not a mistake. I wasn't using them for headlines but I shall.

ondrowan (approved):

PR #59, opened 2019-10-30
Weld mesh geometry vertices and normals

janper:
Weld similar (close enough) vertices into one and reuse such vertices in connected faces

Weld is used to actually connect faces, which are connected merely visually (often resulting in a watertight mesh) or to reduce number of vertices in case when vertices are multiplied because vertices at the same coordinates are not referenced by more faces but rather each face references its own copy of the vertex.

Weld is one of the auto-fixes leading to a simplified, watertight or true-to-its-genus mesh geometries.


src/geometry.rs
```diff
@@ -1399,28 +1397,6 @@ mod tests {
...
-    fn test_unoriented_edge_hash_returns_true_because_reverted() {
-        let unoriented_edge_one_way = UnorientedEdge(OrientedEdge::new(0, 1));
-        let unoriented_edge_other_way = UnorientedEdge(OrientedEdge::new(1, 0));
-        let mut hasher_1 = DefaultHasher::new();
-        let mut hasher_2 = DefaultHasher::new();
-        unoriented_edge_one_way.hash(&mut hasher_1);
-        unoriented_edge_other_way.hash(&mut hasher_2);
-        assert_eq!(hasher_1.finish(), hasher_2.finish());
-    }
-
```
yanchith: Don't remove this 😉

src/interpreter_funcs.rs
```diff
@@ -211,6 +242,7 @@ pub const FUNC_ID_SHRINK_WRAP: FuncIdent = FuncIdent(1);
 pub const FUNC_ID_TRANSFORM: FuncIdent = FuncIdent(2);
 pub const FUNC_ID_LAPLACIAN_SMOOTHING: FuncIdent = FuncIdent(3);
 pub const FUNC_ID_SEPARATE_ISOLATED_MESHES: FuncIdent = FuncIdent(4);
+pub const FUNC_ID_WELD: FuncIdent = FuncIdent(6);
```
yanchith, line 276: You'll get a merge conflict here anyway. Keep this a 5 and resolve to the actual ID when that happens.

src/mesh_tools.rs
```diff
@@ -1,11 +1,155 @@
...
+///
+/// Weld is used to actually connect faces, which are connected merely visually
+/// (often resulting in a watertight mesh) or to reduce number of vertices in
+/// case when vertices are multiplied because vertices at the same coordinates
+/// are not referenced by more faces but rather each face references its own
+/// copy of the vertex.
+///
+/// Weld is one of the auto-fixes leading to a simplified, watertight or
+/// true-to-its-genus mesh geometries.
+pub fn weld(geometry: &Geometry, tolerance: f32) -> Geometry {
```
yanchith, line 22: Let's document how exactly tolerance affects the algorithm.

src/mesh_tools.rs
```diff
@@ -1,11 +1,155 @@
...
+    // key = rounded vertex position with a tolerance (it's expected that the
+    // same value will be shared by more close vertices)
+    // value = actual positions of close vertices
+    let mut vertex_proximity_map: HashMap<(u32, u32, u32), SmallVec<[usize; 8]>> = HashMap::new();
+    for (current_vertex_index, vertex) in geometry.vertices().iter().enumerate() {
+        let vertex_with_tolerance = (
+            (vertex.x / tolerance).floor() as u32,
+            (vertex.y / tolerance).floor() as u32,
+            (vertex.z / tolerance).floor() as u32,
```
yanchith: Why floor instead of round? Also... what about negative coordinates? Isn't a signed integer a better choice here? Also, `i32` might not be sufficient and there is a rust bug we really don't want to trigger (https://doc.rust-lang.org/reference/expressions/operator-expr.html#semantics and https://github.com/rust-lang/rust/issues/10184), so let's use `i64`.

src/mesh_tools.rs
```diff
@@ -185,4 +386,20 @@ mod tests {
             ));
         }
     }
+
+    // FIXME: test on more geometries, check with updated visual similarity
```
yanchith: > check with updated visual similarity comparators

Updated how?

src/mesh_tools.rs
```diff
@@ -1,11 +1,155 @@
...
+        close_vertices.push(current_vertex_index);
+    }
+
+    // All vertices sorted into clusters of positionally close items. These will
+    // be later averaged into a single vertex
+    let close_vertex_clusters = vertex_proximity_map.values();
+
+    // key = original vertex index value = new (averaged) vertex index It is
+    // expected that more keys will share the same value; more original vertices
+    // will be replaced by a single averaged vertex
```
yanchith: I think there's some punctuation missing here, but not sure where

src/mesh_tools.rs
```diff
@@ -1,11 +1,155 @@
...
+            .fold(Point3::origin(), |summed: Point3<f32>, old_vertex_index| {
+                summed + geometry.vertices()[*old_vertex_index].coords
+            })
+            / old_vertex_indices.len() as f32
+    });
+
+    // New faces with renumbered vertex (and normal) indices. Some faces might
+    // end up invalid (not referencing three distinct vertices). Those will be
+    // removed as they don't affect the visual appearance of the mesh geometry.
+    #[allow(clippy::unnecessary_filter_map)]
```
yanchith: Why the filter map though? Won't an analogous process also happen for quads?

src/mesh_tools.rs
```diff
@@ -1,11 +1,155 @@
...
+            .iter()
+            .fold(Point3::origin(), |summed: Point3<f32>, old_vertex_index| {
+                summed + geometry.vertices()[*old_vertex_index].coords
+            })
+            / old_vertex_indices.len() as f32
+    });
+
+    // New faces with renumbered vertex (and normal) indices. Some faces might
+    // end up invalid (not referencing three distinct vertices). Those will be
+    // removed as they don't affect the visual appearance of the mesh geometry.
```
yanchith, line 68: Is it guaranteed that there will be no holes if we filter out degenerate faces? It doesn't sound implausible, but...

src/mesh_tools.rs
```diff
@@ -1,11 +1,155 @@
...
+    // original vertices in clusters and averaging those.
+    let mut old_vertex_normals_index_map: HashMap<u32, SmallVec<[u32; 8]>> = HashMap::new();
+    for face in geometry.faces() {
+        match face {
+            Face::Triangle(f) => {
+                let vertex_indices: SmallVec<_> = SmallVec::from_buf([
+                    (f.vertices.0, f.normals.0),
+                    (f.vertices.1, f.normals.1),
+                    (f.vertices.2, f.normals.2),
+                ]);
```
yanchith: Can just be an array, no need for smallvec.

src/mesh_tools.rs
```diff
@@ -1,11 +1,155 @@
...
+                for (vertex_index, normal_index) in vertex_indices {
+                    let associated_normals = &mut old_vertex_normals_index_map
```
yanchith: Why the `&mut`? The entry already is a mutable handle.

src/mesh_tools.rs
```diff
@@ -1,11 +1,155 @@
...
+    }
+
+    // All vertices sorted into clusters of positionally close items. These will
+    // be later averaged into a single vertex
+    let close_vertex_clusters = vertex_proximity_map.values();
+
+    // key = original vertex index value = new (averaged) vertex index It is
+    // expected that more keys will share the same value; more original vertices
+    // will be replaced by a single averaged vertex
+    let mut old_new_vertex_map: HashMap<usize, usize> = HashMap::new();
```
yanchith: Later in the PR you use `u32`. Let's be consistent.

src/mesh_tools.rs
```diff
@@ -1,11 +1,155 @@
...
+                }
+            }
+        }
+    }
+
+    // Associate old normals to the new averaged vertices
+    let mut new_vertex_old_normals_index_map: Vec<SmallVec<[u32; 8]>> =
+        vec![SmallVec::new(); new_vertices.len()];
+    for (old_vertex_index, old_normals_indices) in old_vertex_normals_index_map {
+        if let Some(new_vertex_index) = old_new_vertex_map.get(&cast_usize(old_vertex_index)) {
```
yanchith: Nice of you to use `if let`, but shouldn't we `expect` this?

src/mesh_tools.rs
```diff
@@ -1,11 +1,155 @@
...
+            new_vertex_old_normals_index_map[*new_vertex_index]
+                .extend_from_slice(&old_normals_indices);
+        };
```
yanchith: Unnecessary `;`

src/mesh_tools.rs
```diff
@@ -70,6 +214,63 @@ mod tests {
         Vector3::new(x, y, z)
     }
 
+    fn tessellated_triangle_geometry_after_welding() -> Geometry {
```
yanchith, line 259: This is dependent on the welding parameter. Not sure I am a fan of creating these data functions that are only used once. Shouldn't this be inlined in the tests? Also below.

src/mesh_tools.rs
```diff
@@ -185,4 +386,20 @@ mod tests {
...
     }
+
+    // FIXME: test on more geometries, check with updated visual similarity
+    // comparators
+    #[test]
+    fn test_weld_tesselated_triangle() {
+        let geometry = tessellated_triangle_geometry_for_welding();
+        let geometry_after_welding_correct = tessellated_triangle_geometry_after_welding();
+
+        // TODO: Vertex order fails
```
yanchith: Huh?

yanchith (changes requested):
The algorithm itself seems legit, but there are some issues on the rust side of things. Also some accidental and unexplained things.

src/geometry.rs
janper: you requested it should be removed several PRs ago. it returned from beyond the grave, so i got rid of it again

src/mesh_tools.rs
janper: you're right, this is my bad assumption.

src/mesh_tools.rs
janper: already updated, so i should make more tests

src/mesh_tools.rs
janper: i don't know how to Rust this with a simple `map`. too many matches...

src/mesh_tools.rs
janper: yes, it is (i can explain visually in person)

src/geometry.rs
yanchith: 🤔 I don't presume to remember what I said exactly, but I remember I asked you to remove another test, which really was incorrect. We removed that one and that was that.

These 2 test functions seem correct. Do you remember the exact context of the conversation?

src/mesh_tools.rs
yanchith: Well, you remove the `filter_` from `filter_map`, and you remove the `Some` wraps in the lambda body, and that should work 🙂

yanchith: I was worried that this was a logic decision I didn't understand, but if this is just a rust issue, it's simple to solve.

src/mesh_tools.rs
ondrowan: Nit: There's missing period at the end of first sentence.

src/mesh_tools.rs
```diff
@@ -1,11 +1,155 @@
...
+    for (current_vertex_index, vertex) in geometry.vertices().iter().enumerate() {
+        let vertex_with_tolerance = (
+            (vertex.x / tolerance).floor() as u32,
+            (vertex.y / tolerance).floor() as u32,
+            (vertex.z / tolerance).floor() as u32,
+        );
+        let close_vertices = vertex_proximity_map
+            .entry(vertex_with_tolerance)
+            .or_insert_with(SmallVec::new);
+        close_vertices.push(current_vertex_index);
```
ondrowan: While not better, I think it looks cleaner if you use `.entry(..).and_modify(..).or_insert(..)` chain.

src/mesh_tools.rs
ondrowan: I'm not entirely against this. I think it's okay since it gives a name to something complicated, it doesn't pollute code of the test case and you can focus on more important code when looking at it.

ondrowan (changes requested):

src/geometry.rs
janper: I don't. Okay, I'll put them back.

src/interpreter_funcs.rs
janper: ok

src/mesh_tools.rs
janper: forgotten note, it doesn't fail anymore

src/mesh_tools.rs
janper: the following makes the test fail:
```
vertex_proximity_map
            .entry(vertex_with_tolerance)
            .and_modify(|close_vertices| close_vertices.push(current_vertex_index))
            .or_insert_with(SmallVec::new);
```

janper:
While implementing the requests, I found an error in `cube_sharp_var_len` and fixed it. 
image

yanchith:
> While implementing the requests, I found an error in `cube_sharp_var_len` and fixed it.

Nice, but it is not clear from the diff or the screenshot what the error was. What is it? 😂

janper:
> > While implementing the requests, I found an error in `cube_sharp_var_len` and fixed it.
> 
> Nice, but it is not clear from the diff or the screenshot what the error was. What is it? 😂

Originally i thought some of the faces were referencing wrong vertices resulting in a corrupted shape of the mesh. Eventually the overall shape was okay, only the triangulation of the sides of the cube was different than the one of the `cube_sharp_same_len`. The two meshes were then visually different, because the sides of the cube were cut by a different diagonal.

janper:
ready for another round of reviews

src/mesh_tools.rs
ondrowan: Well, you have to initialize `SmallVec` with the first value:

```rust
vertex_proximity_map
            .entry(vertex_with_tolerance)
            .and_modify(|close_vertices| close_vertices.push(current_vertex_index))
            .or_insert_with(|| smallvec![current_vertex_index]);
```

But it's potentionally good as it is. I don't think there's any performance benefit to my suggested code, so pick whichever you think is nicer and more readable.

src/mesh_tools.rs
yanchith: Nice but please go for `i64` as I suggested in the comment above.

src/mesh_tools.rs
```diff
@@ -104,20 +99,17 @@ pub fn weld(geometry: &Geometry, tolerance: f32) -> Geometry {
...
+                for (vertex_index, normal_index) in &vertex_indices {
+                    let associated_normals = old_vertex_normals_index_map
+                        .entry(*vertex_index)
                         .or_insert_with(SmallVec::new);
-                    if associated_normals
-                        .iter()
-                        .all(|value| *value != normal_index)
-                    {
-                        associated_normals.push(normal_index);
+                    if associated_normals.iter().all(|value| value != normal_index) {
```
yanchith: Please use `!associated_normals.contains(&normal_index)` here.

src/mesh_tools.rs
```diff
@@ -333,6 +334,158 @@ mod tests {
         Geometry::from_triangle_faces_with_vertices_and_normals(faces, vertices, vertex_normals)
     }
 
+    pub fn cube_sharp_same_len(position: [f32; 3], scale: f32) -> Geometry {
```
yanchith, line 388: This is misformatted. Either format properly manually, or don't use `#[rustfmt::skip]`...

Also... is this a 1:1 copy of this function in `geometry.rs`? If so, import it instead. If not, document the differences.

src/mesh_tools.rs
```diff
@@ -333,6 +334,158 @@ mod tests {
...
+    ];
+
+        Geometry::from_triangle_faces_with_vertices_and_normals(
+            faces,
+            vertex_positions,
+            vertex_normals,
+        )
+    }
+
+    pub fn cube_smooth_var_len(position: [f32; 3], scale: f32) -> Geometry {
```
yanchith: Same as above. Why duplicate this?

yanchith (changes requested):
I see you duplicated functions from `geometry.rs`, perhaps with small changes (hard to tell without comments telling the intent)? If they were incorrect in `geometry.rs`, let's fix that, but let's not duplicate without a good reason.

src/mesh_tools.rs
janper: The function from `geometry.rs` will be eventually removed because it's mainly good for testing. I think it shouldn't end up being compiled into the release, so I put a copy of it into the test scope here. Therefore the 1:1 copy.

src/mesh_tools.rs
janper: I'll rename this one because the normals are calculated differently from your function to match the welding algorithm.

src/mesh_tools.rs
yanchith: Okay, agreed!

yanchith: This is still misformatted though!

src/interpreter_funcs.rs
```diff
@@ -203,6 +203,37 @@ impl Func for FuncImplSeparateIsolatedMeshes {
     }
 }
 
+pub struct FuncImplWeld;
+impl Func for FuncImplWeld {
+    fn flags(&self) -> FuncFlags {
+        FuncFlags::empty()
+    }
+    fn param_info(&self) -> &[ParamInfo] {
```
yanchith, line 240: Missing a newline here

src/mesh_tools.rs
janper: ah, right. dumb me!

src/interpreter_funcs.rs
janper: the flags don't need to be there anyway

src/mesh_tools.rs
ondrowan: Isn't this misformatted again but in a different way?

janper: i don't think so. `cargo fmt` formats it this way

ondrowan: I just meant that the other function in `geometry.rs` uses `rustfmt::skip` and the numbers are aligned nicely into columns, but this is also fine if it's okay with both of you.

ondrowan (dismissed):

ondrowan (approved):

PR #62, opened 2019-11-02
Synchronize and revert mesh face winding

janper:
SYNCHRONIZE MESH FACE WINDING

Make sure all the faces are oriented the same way - have the same winding (vertex order).

This function crawls the mesh geometry and flips all the faces, which are not facing the same way as the previous faces in the process, starting with the first face in the list. As a result, the entre mesh can end up facing inwards (be entirely reverted). At the moment we have no tools to detect such a case automatically, so we need to rely on the user to check it and potentially revert winding of the entire mesh.

The algorithm relies on the fact that in a proper non-manifold mesh, each oriented edge has exactly one (for watertight mesh geometry) or none (for mesh patches) counter-edge oriented the opposite direction. It crawls the geometry and if a face neighboring the current one doesn't have the proper winding, it's being reverted and only then triggers checking its own neighbors.

This method, doesn't flip the normals associated with the face vertices, as there is no unambiguous way to do so automatically.

REVERT MESH FACE WINDING

Reverts vertex and normal winding of all faces in the mesh geometry and returns a reverted mesh geometry


src/mesh_tools.rs
```diff
@@ -2,15 +2,165 @@ use std::collections::{HashMap, HashSet};
...
+    Geometry::from_triangle_faces_with_vertices_and_normals(
+        reverted_faces,
+        geometry.vertices().to_vec(),
+        geometry.normals().to_vec(),
+    )
+}
 
 /// Crawls the geometry to find continuous patches of geometry.
 /// Returns a vector of new separated geometries.
 pub fn separate_isolated_meshes(geometry: &Geometry) -> Vec<Geometry> {
```
janper, line 287: ~Wasn't this already done differently? The topology was outside the function.
Nevermind, I'll do it again.~
Now I remember: I did it for the Separation operation.

src/geometry.rs
```diff
@@ -1722,4 +1772,109 @@ mod tests {
...
+        assert_eq!(face_reverted_correct, face_reverted_calculated);
+    }
+
+    #[test]
+    fn test_triangle_face_to_reverted_should_fail_because_compared_to_same() {
+        let face = TriangleFace::new_separate(1, 2, 3, 4, 5, 6);
+
+        let face_reverted_calculated = face.to_reverted();
+        assert_ne!(face, face_reverted_calculated);
+    }
```
ondrowan, line 1803: I think this test is redundant since you've proven this works in previous one.

src/geometry.rs
```diff
@@ -1722,4 +1772,109 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_triangle_face_to_reverted_should_pass_because_compared_to_reverted() {
+        let face = TriangleFace::new_separate(1, 2, 3, 4, 5, 6);
+        let face_reverted_correct = TriangleFace::new_separate(3, 2, 1, 6, 5, 4);
+
+        let face_reverted_calculated = face.to_reverted();
+        assert_eq!(face_reverted_correct, face_reverted_calculated);
+    }
```
ondrowan, line 1795: I find many of these tests problematic. `assert_eq` uses partial equality, which is defined in such a way for `TriangleFace` that it equals even if vertices and normals are shuffled. This therefore doesn't prove that `to_reverted` really works.

src/interpreter/value.rs
```diff
@@ -191,6 +191,17 @@ impl Value {
...
+    /// Get the refcounted value if geometry, otherwise panic.
+    ///
+    /// # Panics
+    /// This function panics when value is not a geometry.
+    pub fn unwrap_refcounted_geometry(&self) -> Arc<Geometry> {
+        match self {
+            Value::Geometry(geometry_ptr) => Arc::clone(geometry_ptr),
+            _ => panic!("Value not geometry"),
+        }
+    }
```
ondrowan: This is also implemented in the other PR by @yanchith, so we'll have to sync merging it.

src/interpreter_funcs.rs
```diff
@@ -274,6 +335,8 @@ pub const FUNC_ID_LAPLACIAN_SMOOTHING: FuncIdent = FuncIdent(3);
 pub const FUNC_ID_SEPARATE_ISOLATED_MESHES: FuncIdent = FuncIdent(4);
 pub const FUNC_ID_JOIN_MESHES: FuncIdent = FuncIdent(5);
 pub const FUNC_ID_WELD: FuncIdent = FuncIdent(6);
+pub const FUNC_ID_REVERT_MESH_FACES: FuncIdent = FuncIdent(7);
+pub const FUNC_ID_SYNCHRONIZE_MESH_FACES: FuncIdent = FuncIdent(8);
```
ondrowan: Just a note this will clash with Loop Subdivision PR once it rebases onto master.

src/mesh_analysis.rs
```diff
@@ -67,6 +67,11 @@ pub fn non_manifold_edges<'a>(
         })
 }
 
+// Checks if mesh contains only manifold or border edges
+pub fn is_mesh_manifold(edge_sharing: &EdgeSharingMap) -> bool {
+    non_manifold_edges(edge_sharing).next().is_none()
```
ondrowan, line 71: You should be able to remove `dead_code` from `non_manifold_edges` now.

src/mesh_analysis.rs
```diff
@@ -67,6 +67,11 @@ pub fn non_manifold_edges<'a>(
         })
 }
 
+// Checks if mesh contains only manifold or border edges
```
ondrowan: There's one more `/` missing.

src/mesh_tools.rs
```diff
@@ -5,8 +5,158 @@ use nalgebra::geometry::Point3;
...
+/// potentially revert winding of the entire mesh.
+///
+/// The algorithm relies on the fact that in a proper non-manifold mesh, each
+/// oriented edge has exactly one (for watertight mesh geometry) or none (for
+/// mesh patches) counter-edge oriented the opposite direction. It crawls the
+/// geometry and if a face neighboring the current one doesn't have the proper
+/// winding, it's being reverted and only then triggers checking its own
+/// neighbors.
+///
+/// This method, doesn't flip the normals associated with the face vertices, as
```
ondrowan: I think the first comma is a typo.

src/mesh_tools.rs
```diff
@@ -5,8 +5,158 @@ use nalgebra::geometry::Point3;
 use smallvec::{smallvec, SmallVec};
 
 use crate::convert::{cast_u32, cast_usize};
-use crate::geometry::{Face, Geometry, TriangleFace};
-use crate::mesh_topology_analysis::face_to_face_topology;
+use crate::geometry::{Face, Geometry, OrientedEdge, TriangleFace, UnorientedEdge};
+use crate::mesh_topology_analysis;
+
+/// Make sure all the faces are oriented the same way - have the same winding
```
ondrowan: Nit: "Makes"

src/mesh_tools.rs
```diff
@@ -5,8 +5,158 @@ use nalgebra::geometry::Point3;
...
+
+    // Synchronize faces in all mesh geometry triangles
+    while synchronized_triangle_faces.len() < original_triangle_faces.len() {
+        // Start with the first untreated face in the list of the original mesh
+        // geometry faces. The winding of this face also determines the winding
+        // of the rest of the mesh geometry.
+        let mut current_face_index: u32 = 0;
+        while face_treatment_pattern[cast_usize(current_face_index)] {
+            current_face_index += 1;
+        }
```
ondrowan: You can use iterator's `position` for searching for index.

src/mesh_tools.rs
```diff
@@ -5,8 +5,158 @@ use nalgebra::geometry::Point3;
...
+            // it's given by the topology generator.
+            let proper_current_face = if current_face.contains_oriented_edge(current_test_edge) {
+                current_face
+            } else {
+                current_face.to_reverted()
+            };
+            synchronized_triangle_faces.push(proper_current_face);
+
+            // Edge-to-index map for faster lookup
+            let mut unoriented_edge_index_map: HashMap<UnorientedEdge, u32> = HashMap::new();
```
ondrowan: You don't need to allocate this anew every single iteration. It can be created with size of `unoriented_edges.len()` before the while loop.

src/mesh_tools.rs
```diff
@@ -146,8 +296,8 @@ pub fn weld(geometry: &Geometry, tolerance: f32) -> Geometry {
 /// Crawls the geometry to find continuous patches of geometry.
 /// Returns a vector of new separated geometries.
 pub fn separate_isolated_meshes(geometry: &Geometry) -> Vec<Geometry> {
-    let face_to_face = face_to_face_topology(geometry);
-    let mut available_face_indices: HashSet<u32> = face_to_face.keys().copied().collect();
+    let face_to_face = mesh_topology_analysis::face_to_face_topology(geometry);
+    let mut available_face_indices: HashSet<u32> = face_to_face.keys().cloned().collect();
```
ondrowan: Why was this changed from `copied` to `cloned`?

ondrowan (dismissed):

yanchith:
@janper 

I'll review the code once @ondrowan is happy with it (unless Ondro wants to hand off the review to me), but I do have some high level questions just based on your description:

> The algorithm relies on the fact that in a proper non-manifold mesh, each oriented edge has exactly one (for watertight mesh geometry) or none (for mesh patches) counter-edge oriented the opposite direction.

Did this mean to say "manifold", or "non non-manifold" instead? Also: why can't mesh patches also have the edge facing the opposite side (when the edge in question is not a border edge)?

> This method, doesn't flip the normals associated with the face vertices, as there is no unambiguous way to do so automatically.

Why can't the (vertex) normal just be flipped in the plane defined by the vertex and the face normal? Do we sometimes want other strategies? We can decide to tackle this later and just regenerate normals though.

src/geometry.rs
```diff
@@ -486,6 +509,29 @@ impl From<(u32, u32, u32)> for TriangleFace {
     }
 }
 
+impl PartialEq for TriangleFace {
```
yanchith: I disagree with this (as I already mentioned on our slack discussion), at least until you convince me 😹 . Let's discuss offline.

src/geometry.rs
janper: i need to make sure this won't be considered equal. that is not covered by the previous one.

src/geometry.rs
janper: not shuffled! shifted. two faces with shifted lists of referenced vertices and normals are identical. the tests do prove this works, also because they are covering also inequality (see the following test).

src/interpreter/value.rs
janper: this is a copypaste of his code. but true, let's be careful when merging

src/interpreter_funcs.rs
janper: i would take the numbering into consideration (i.e. counting from 7 when i know there will be another merge using 6) but @yanchith suggested i'll have to resolve merge conflicts there anyway, so there is no need to predict the numbering.

src/mesh_analysis.rs
janper: thanks, will do. this might occur more often now. my plan is to make a PR for such maintenance tasks once in a while.

src/mesh_tools.rs
janper: true, but let's not be like this. eventually we should  proofread all the comments and docs anyway

src/mesh_tools.rs
janper: not sure. will set it back to `copied`

src/mesh_tools.rs
janper: > @janper
> 
> I'll review the code once @ondrowan is happy with it (unless Ondro wants to hand off the review to me), but I do have some high level questions just based on your description:
> 
> > The algorithm relies on the fact that in a proper non-manifold mesh, each oriented edge has exactly one (for watertight mesh geometry) or none (for mesh patches) counter-edge oriented the opposite direction.
> 
> Did this mean to say "manifold", or "non non-manifold" instead? Also: why can't mesh patches also have the edge facing the opposite side (when the edge in question is not a border edge)?

you're right in both cases. it's an error in the comment though. i'll fix it

janper:
> @janper
> 
> I'll review the code once @ondrowan is happy with it (unless Ondro wants to hand off the review to me), but I do have some high level questions just based on your description:
> 
> > The algorithm relies on the fact that in a proper non-manifold mesh, each oriented edge has exactly one (for watertight mesh geometry) or none (for mesh patches) counter-edge oriented the opposite direction.
> 
> Did this mean to say "manifold", or "non non-manifold" instead? Also: why can't mesh patches also have the edge facing the opposite side (when the edge in question is not a border edge)?

You're right in both cases. I'll fix the comment, the logic should be ok

janper:
> > This method, doesn't flip the normals associated with the face vertices, as there is no unambiguous way to do so automatically.
> 
> Why can't the (vertex) normal just be flipped in the plane defined by the vertex and the face normal? Do we sometimes want other strategies? We can decide to tackle this later and just regenerate normals though.

There can be some strategies we can consider but I'm afraid it won't be simple. The question is if we need to care about the normals anyway. We might have an operation to generate sharp or flat normals for an existing mesh instead.

src/geometry.rs
janper: I'm strongly **for** this solution and you'll have a hard time convincing me otherwise ;) . Later tonight I'll write down my reasons.

janper: 1. The face is indifferent to which vertex/normal pair is referenced first. Faces `(0, 1, 2)`, `(1, 2, 0)`, `(2, 0, 1)` are identical. Not similar, not alike, they really are identical.
2. Our data structure suffers from non-existence of a circular list in the language. Instead, we're compensating it by the `Eq` and `Hash` functions.
3. Another solution for this would be some unification of face-vertex shift. Let's say when the face is being created, the winding will be reordered so that the lowest index will come first in list. I don't like it mainly because there is no need for this.
4. Having a custom `Eq` and `Hash` function allows for direct use of language features, such as `==` or hash-based structures, which is and will be incredibly useful.
5. There is no reason for wrapping the "linear" face into a "circular" face tuplestruct because there is no use case where the linear face would be useful or even correct. This is a different case from the `OrientedEdge` and `UnorientedEdge`, where the different view makes some real difference and there are different use cases for them. The face and its wrapper would only add more complexity and will have no reason, nor use.
6. There are currently no tests that would fail with the custom `Eq` function but if there were, I prefer manually checking for field-by-field equality in the test rather than in the main software logic. It would privilege technicalities to semantics.
7. Because of points 1. and 2. I believe any logic that would consider faces `(0, 1, 2)` and `(1, 2, 0)` different would be wrong even in tests. This point makes point 6. obsolete.

ondrowan:
@yanchith You can do your review pass, I pretty much said what I wanted to say. I might do another sweep later on when current problems are fixed.

src/geometry.rs
ondrowan: What I'm trying to say is that I'd expect you to test if `to_reverted` really reverted verices and normals. In this case, because of how partial equality works, all of these tests would pass:

```
assert_eq!(
    face_reverted_calculated,
    TriangleFace::new_separate(3, 2, 1, 6, 5, 4)
);
assert_eq!(
    face_reverted_calculated,
    TriangleFace::new_separate(1, 3, 2, 4, 6, 5)
);
assert_eq!(
    face_reverted_calculated,
    TriangleFace::new_separate(2, 1, 3, 5, 4, 6)
);
```

I know it's technically true, but at the same time, I'd expect this test to pass only the first case to showcase that reverted vertices and normals are **exactly equal**. For all we know, there might be some bug reverting these values incorrectly to e.g.  `(1, 3, 2, 4, 6, 5)` and this test case wouldn't catch it.

janper: I understand. We can test the reverted face to be the exact mirrored-image of the original, but what would it be good for? Semantically `(1, 3, 2, 4, 6, 5)` is a reverted version of `(1, 2, 3, 4, 5, 6)`.

 `(1, 2, 3, 4, 5, 6)` ==  `(2, 3, 1, 5, 6, 4)` ==  `(3, 1, 2, 6, 4, 5)` <- they are the same/identical face

ondrowan: That's of course right, but since this is an unit test of function that should produce precise order of numbers, I'd lean towards exact equality.

Maybe @yanchith has something to say about this as well since he had an opinion regarding partial equality.

janper: well, if you insist, I'll check the result field by field. I still think this is an unreasonable privilege of a technicality over a true nature of the data.

ondrowan: I don't insist, it's just a gut feeling. That's why I'd like to hear another opinion.

src/geometry.rs
yanchith: Thanks for the write-up. I hear your arguments and sympathize with them - this is actually very similar to `HashSet` and `HashMap`, which also override/weaken the meaning of `Eq`. The other side is that I don't agree with the following:

> I believe any logic that would consider faces (0, 1, 2) and (1, 2, 0) different would be wrong even in tests.

I think that at least in tests, we do want to have the ability to compare chunks of memory without additional semantics. I will review the existing tests to see how this would affect them.

> There are currently no tests that would fail with the custom Eq

I am worried about the opposite case - a test falsely passing because of the weaker `Eq`.

src/geometry.rs
```diff
@@ -475,8 +475,31 @@ impl TriangleFace {
...
+            self.vertices.1,
+            self.vertices.0,
+            self.normals.2,
+            self.normals.1,
+            self.normals.0,
+        )
+    }
+
+    /// Checks if the other face is the same ith a reverted vertex and normal winding
+    pub fn is_reverted(&self, other: Self) -> bool {
```
yanchith: This shouldn't take `other` by value. Disregard if clippy suggested this.

src/geometry.rs
```diff
@@ -475,8 +475,31 @@ impl TriangleFace {
...
     /// Does the face contain the specific unoriented edge
     pub fn contains_unoriented_edge(self, unoriented_edge: UnorientedEdge) -> bool {
-        let [o_e_0, o_e_1, o_e_2] = self.to_unoriented_edges();
-        o_e_0 == unoriented_edge || o_e_1 == unoriented_edge || o_e_2 == unoriented_edge
+        let [u_e_0, u_e_1, u_e_2] = self.to_unoriented_edges();
+        u_e_0 == unoriented_edge || u_e_1 == unoriented_edge || u_e_2 == unoriented_edge
+    }
+
+    /// Does the face contain the specific oriented edge
+    pub fn contains_oriented_edge(self, oriented_edge: OrientedEdge) -> bool {
```
yanchith: This shouldn't take `self` nor `oriented_edge` by value. If clippy forced the parameter to be passed by value, at least don't pass `self` by value.

src/geometry.rs
```diff
@@ -475,8 +475,31 @@ impl TriangleFace {
 
     /// Does the face contain the specific unoriented edge
     pub fn contains_unoriented_edge(self, unoriented_edge: UnorientedEdge) -> bool {
```
yanchith: This shouldn't take `self` not `unoriented_edge` by value. If clippy forced the parameter to be passed by value, at least don't pass `self` by value.

src/geometry.rs
```diff
@@ -523,6 +569,10 @@ impl OrientedEdge {
     pub fn contains_vertex(self, vertex_index: u32) -> bool {
         self.vertices.0 == vertex_index || self.vertices.1 == vertex_index
     }
+
+    pub fn to_reverted(self) -> Self {
```
yanchith, line 594: Don't take self by value.

src/geometry.rs
```diff
@@ -1722,4 +1772,109 @@ mod tests {
...
+    fn test_triangle_face_to_reverted_should_pass_because_compared_to_reverted() {
+        let face = TriangleFace::new_separate(1, 2, 3, 4, 5, 6);
+        let face_reverted_correct = TriangleFace::new_separate(3, 2, 1, 6, 5, 4);
+
+        let face_reverted_calculated = face.to_reverted();
+        assert_eq!(face_reverted_correct, face_reverted_calculated);
+    }
+
+    #[test]
+    fn test_triangle_face_to_reverted_should_fail_because_compared_to_same() {
```
yanchith: Wouldn't it be better to assert the exact structure of the reverted face here? `assert_ne!` proves very little about about the property we actually wan to test here.

src/geometry.rs
yanchith: Reading the tests, I have to agree with @ondrowan . The weaker equality doesn't prove the properties we want to prove in some tests. We either must not use it in those tests and compare memory manually (not a fan of this, it is going to be very error-prone), or devise something else. I suggest making it an explicit method on `Face` and `TriangleFace` called `equivalent(&self, other: &Self) -> bool` and use it when we don't want to just compare memory. It's going to be less comfortable, but I suspect the relaxed `Eq` impl would lead to many tests that pass by accident.

src/geometry.rs
```diff
@@ -1722,4 +1772,109 @@ mod tests {
 
         assert!(geometry.has_no_orphan_vertices());
     }
+
+    #[test]
+    fn test_triangle_face_to_reverted_should_pass_because_compared_to_reverted() {
```
yanchith: Let's not use the word `should` in test names. Instead name them like this:  `test_[object_if_any]_[method_or_function]_[situation]`, e.g: `test_triangle_face_to_reverted_comparison_to_reverted`. (You've got this right for some tests, but sometimes you are inconsistent).

Applies to all tests in the PR.

src/geometry.rs
```diff
@@ -1722,4 +1772,109 @@ mod tests {
...
+    #[test]
+    fn test_triangle_face_to_reverted_should_fail_because_compared_to_same() {
+        let face = TriangleFace::new_separate(1, 2, 3, 4, 5, 6);
+
+        let face_reverted_calculated = face.to_reverted();
+        assert_ne!(face, face_reverted_calculated);
+    }
+
+    #[test]
+    fn test_triangle_face_to_reverted_should_pass_because_compared_to_reverted_and_shifted() {
```
yanchith: Instead of testing with `assert_eq`, test the new `TriangleFace::equivalent` function here.

src/geometry.rs
```diff
@@ -1722,4 +1772,109 @@ mod tests {
...
+        let face = TriangleFace::new_separate(1, 2, 3, 4, 5, 6);
+        let face_reverted_correct_shifted = TriangleFace::new_separate(2, 1, 3, 5, 4, 6);
+
+        let face_reverted_calculated = face.to_reverted();
+        assert_eq!(face_reverted_correct_shifted, face_reverted_calculated);
+    }
+
+    #[test]
+    fn test_triangle_face_is_reverted_should_should_return_true() {
+        let face = TriangleFace::new_separate(1, 2, 3, 4, 5, 6);
```
yanchith, line 1816: Instead of testing with `assert`, test the new `TriangleFace::is_reverted_equivalent` function here.

src/geometry.rs
```diff
@@ -1722,4 +1772,109 @@ mod tests {
...
+        let face_reverted_correct_shifted = TriangleFace::new_separate(2, 1, 3, 5, 4, 6);
+
+        assert!(face_reverted_correct_shifted.is_reverted(face));
+    }
+
+    #[test]
+    fn test_triangle_face_is_reverted_should_should_return_false() {
```
yanchith: Instead of testing with `assert_eq`, test the new `TriangleFace::is_reverted_equivalent` function here.

src/geometry.rs
```diff
@@ -1722,4 +1772,109 @@ mod tests {
 
         assert!(geometry.has_no_orphan_vertices());
     }
+
```
yanchith, line 1745: In all the created tests in this file you use `TriangleFace::new_separate(1, 2, 3, 4, 5, 6)` or similar, but you never actually use the normals. Use `TriangleFace::new` so this doesn't mislead.

src/interpreter_funcs.rs
```diff
@@ -264,6 +265,66 @@ impl Func for FuncImplWeld {
...
+        Value::Geometry(Arc::new(value))
+    }
+}
+
+pub struct FuncImplSynchronizeMeshFaces;
+impl Func for FuncImplSynchronizeMeshFaces {
+    fn flags(&self) -> FuncFlags {
+        FuncFlags::PURE
+    }
+    fn param_info(&self) -> &[ParamInfo] {
```
yanchith, line 498: Missing newline

src/interpreter_funcs.rs
```diff
@@ -264,6 +265,66 @@ impl Func for FuncImplWeld {
     }
 }
 
+pub struct FuncImplRevertMeshFaces;
+impl Func for FuncImplRevertMeshFaces {
+    fn param_info(&self) -> &[ParamInfo] {
```
yanchith, line 473: Is this not pure on purpose? Looking at the implementation I see no reason why it shouldn't be.

src/interpreter_funcs.rs
```diff
@@ -264,6 +265,66 @@ impl Func for FuncImplWeld {
...
+    }
+
+    fn call(&self, args: &[Value]) -> Value {
+        let geometry = args[0].unwrap_refcounted_geometry();
+
+        let oriented_edges: Vec<_> = geometry.oriented_edges_iter().collect();
+        let edge_sharing_map = edge_analysis::edge_sharing(&oriented_edges);
+        let unoriented_edges: Vec<_> = geometry.unoriented_edges_iter().collect();
+        let edge_to_face =
+            mesh_topology_analysis::edge_to_face_topology(&geometry, &unoriented_edges);
```
yanchith: These can be moved inside the `if` so they are not created unnecessarily.

src/mesh_analysis.rs
```diff
@@ -776,6 +780,36 @@ mod tests {
         assert_eq!(oriented_edges_non_manifold_check.len(), 3);
     }
 
+    #[test]
+    fn test_mesh_analysis_manifold_mesh_returns_false_because_non_manifold() {
```
yanchith: The function is called `is_mesh_manifold`, not `manifold_mesh`. Also below.

src/mesh_tools.rs
```diff
@@ -371,6 +524,32 @@ mod tests {
...
+
+        let vertex_normals = vec![n(0.0, 0.0, 1.0), n(0.0, 0.0, 1.0)];
+
+        let faces = vec![
+            TriangleFace::new_separate(0, 3, 1, 0, 0, 0),
+            TriangleFace::new_separate(1, 3, 4, 0, 0, 0),
+            TriangleFace::new_separate(2, 4, 1, 0, 0, 0),
+            TriangleFace::new_separate(3, 5, 4, 0, 0, 0),
+            TriangleFace::new_separate(6, 7, 8, 1, 1, 1),
+        ];
```
yanchith: Which one is flipped? Please mark in comment.

src/mesh_tools.rs
```diff
@@ -5,8 +5,161 @@ use nalgebra::geometry::Point3;
...
+/// (vertex order).
+///
+/// This function crawls the mesh geometry and flips all the faces, which are
+/// not facing the same way as the previous faces in the process, starting with
+/// the first face in the list. As a result, the entre mesh can end up facing
+/// inwards (be entirely reverted). At the moment we have no tools to detect
+/// such a case automatically, so we need to rely on the user to check it and
+/// potentially revert winding of the entire mesh.
+///
+/// The algorithm relies on the fact that in a proper non-manifold mesh, each
```
yanchith: "non-manifold" -> "manifold"?

src/mesh_tools.rs
```diff
@@ -5,8 +5,161 @@ use nalgebra::geometry::Point3;
...
+/// Reverts vertex and normal winding of all faces in the mesh geometry and
+/// returns a reverted mesh geometry
+pub fn revert_mesh_faces(geometry: &Geometry) -> Geometry {
+    let reverted_faces = geometry.faces().iter().map(|face| match face {
+        Face::Triangle(t_f) => t_f.to_reverted(),
+    });
+    Geometry::from_triangle_faces_with_vertices_and_normals(
+        reverted_faces,
+        geometry.vertices().to_vec(),
+        geometry.normals().to_vec(),
```
yanchith: Unnecessary `to_vec()` for both

src/mesh_tools.rs
```diff
@@ -5,8 +5,161 @@ use nalgebra::geometry::Point3;
...
+    // item index = face index; TRUE = the face was already checked, FALSE = the
+    // face hasn't yet been checked
+    let mut face_processed_pattern = vec![false; original_triangle_faces.len()];
+
+    // Faces to be checked for winding, determined by the orientation of the
+    // OrientedEdge of the neighbor, which triggered the check. Current face has
+    // to contain a reverted edge to have the proper winding, otherwise it has
+    // to be reverted.
+    let mut edge_face_stack: Vec<(OrientedEdge, u32)> =
+        Vec::with_capacity(original_triangle_faces.len() / 2);
```
yanchith: Use `Vec::new` unless you know the exact capacity.

src/mesh_tools.rs
```diff
@@ -5,8 +5,161 @@ use nalgebra::geometry::Point3;
...
+    // list have the same vertex winding.
+    let mut synchronized_triangle_faces: Vec<TriangleFace> =
+        Vec::with_capacity(original_triangle_faces.len());
+
+    // Edge-to-index map for faster lookup
+    let mut unoriented_edge_index_map: HashMap<UnorientedEdge, u32> =
+        HashMap::with_capacity(unoriented_edges.len());
+    for (unoriented_edge_index, unoriented_edge) in unoriented_edges.iter().enumerate() {
+        unoriented_edge_index_map.insert(*unoriented_edge, cast_u32(unoriented_edge_index));
+    }
```
yanchith: Use `FromIterator::from_iter`

src/mesh_tools.rs
```diff
@@ -5,8 +5,161 @@ use nalgebra::geometry::Point3;
...
+                }
+            }
+        }
+    }
+
+    // Rebuild the mesh geometry with the new faces and original faces and normals.
+    Geometry::from_triangle_faces_with_vertices_and_normals(
+        synchronized_triangle_faces,
+        geometry.vertices().to_vec(),
+        geometry.normals().to_vec(),
```
yanchith: Unnecessary `to_vec()`

src/mesh_tools.rs
```diff
@@ -5,8 +5,161 @@ use nalgebra::geometry::Point3;
...
+        let mut current_test_edge: OrientedEdge =
+            original_triangle_faces[cast_usize(current_face_index)].to_oriented_edges()[0];
+
+        // Put the first edge-face couple to the stack
+        edge_face_stack.push((current_test_edge, current_face_index));
+
+        // Check and revert (if needed) faces in the stack as long as there is any
+        while !edge_face_stack.is_empty() {
+            // Get ready for the next iteration.
+            let next_edge_face = edge_face_stack.pop().expect("Popping from an empty vector");
```
yanchith: `while let Some(next_face_edge) = edge_face_stack.pop() {`

src/mesh_tools.rs
```diff
@@ -5,8 +5,161 @@ use nalgebra::geometry::Point3;
...
+                face_unoriented_edges.iter().map(|unoriented_edge| {
+                    unoriented_edge_index_map
+                        .get(unoriented_edge)
+                        .expect("The current edge is not found in the edge collection")
+                });
+
+            // Convert into oriented edges so that the neighboring faces can check
+            // for correct winding
+            let face_oriented_edges: Vec<_> =
+                face_unoriented_edges.iter().map(|u_e| u_e.0).collect();
```
yanchith: This seems unnecessary. Can't the oriented edge just be extracted of the unoriented edge on demand in the loop below? You can use `Iterator::zip` make it available.

src/mesh_tools.rs
```diff
@@ -5,8 +5,161 @@ use nalgebra::geometry::Point3;
...
+            // For each face edge index
+            for (i, face_unoriented_edge_index) in face_unoriented_edge_indices.enumerate() {
+                // get the actual oriented edge
+                let face_oriented_edge = face_oriented_edges[i];
+                // and try to find it in the edge-to-face topology.
+                if let Some(edge_in_faces) = edge_to_face_topology.get(&face_unoriented_edge_index)
+                {
```
yanchith: Shouldn't we `.expect()` this?

src/mesh_tools.rs
```diff
@@ -5,8 +5,161 @@ use nalgebra::geometry::Point3;
...
+                    // If it exists, iterate the faces containing the edge
+                    for face_index in edge_in_faces {
+                        // and if it was not already added to the stack or even checked
+                        if !face_processed_pattern[cast_usize(*face_index)] {
+                            // add it to the stack with the expected edge orientation
+                            edge_face_stack.push((face_oriented_edge.to_reverted(), *face_index));
```
yanchith: It might be the lateness of the hour, but why do we revert the edge here?

yanchith (changes requested):
Nice job, but besides the rust technicalities, naming conventions and the usual, there's a few things to discuss here:

I would opt for an explicit semantic comparison method instead of weakening `Eq`. We can name it `TriangleFace::equivalent` or similar. Some of the `is_reverted` methods may need to have another version also: `is_reverted_equivalent`. The tests prove that the current strict `Eq` is still desired in some places.

Before approving, I will also need to do another review pass on the winding synchronization algorithm when my brain is not asleep.

src/geometry.rs
```diff
@@ -412,7 +412,7 @@ impl fmt::Display for Face {
 
 /// A triangular face. Contains indices to other geometry data, such
 /// as vertices and normals.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, Eq)]
```
yanchith: If we implement `PartialEq` manually, we must also implement `Hash` manually.

src/geometry.rs
janper: I can skip this test because what you describe is already covered by other tests. This one was just to make sure that the original and the reverted face are not considered equal. In fact, this is a test of equality rather than of the `to_reverted` function.

src/geometry.rs
janper: Not sure I understand. The new `PartialEq` function takes them into consideration and also  in the test I'm shifting the list of normals along with the list of vertices.

src/interpreter_funcs.rs
janper: My mistake

src/mesh_tools.rs
janper: It's possible that some of the unoriented edges will be unwrapped more than once. Maybe it's not a performance issue, but that was my original reasoning for this solution.

src/mesh_tools.rs
janper: because we are expected to find the reverted edge in the neighboring face. i'm reverting the face already here rather than when checking the neighbor to keep related data in the newly created tuple: it's the edge to be found and the face where it's supposed to be found

src/mesh_tools.rs
yanchith: kk, should actually be a perf benefit

yanchith:
@ondrowan 

Me and @janper discussed offline the issue of the custom `PartialEq` implementation and this is what we finally agreed upon (and I think it is a very good solution, because it reduces cognitive load while preserving the semantic equality):

We are going to remove the custom `PartialEq` in favor of making sure two faces with same winding and same indices will always look the same in memory. We will achieve that by finding the smallest index in the face, and starting the winding there. The downside is that face construction will be a bit more computation heavy, but `PartialEq` will mean both semantic and bitwise equality, so we won't have to think about it in the tests.

Also, we will add a couple of tests to check the abovementioned construction logic.

src/geometry.rs
yanchith: I just meant that since you don't actually need to specify the normals specially for all/most test cases here, you can just use `TriangleFace::new`.

src/geometry.rs
```diff
@@ -420,7 +420,7 @@ impl fmt::Display for Face {
 
 /// A triangular face. Contains indices to other geometry data, such
 /// as vertices and normals.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Serialize)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, serde::Serialize)]
```
ondrowan, line 433: Is this `Hash` needed?

src/geometry.rs
```diff
@@ -432,9 +432,21 @@ impl TriangleFace {
...
+        } else if i2 < i1 && i2 < i3 {
+            TriangleFace {
+                vertices: (i2, i3, i1),
+                normals: (i2, i3, i1),
+            }
+        } else {
+            TriangleFace {
+                vertices: (i3, i1, i2),
+                normals: (i3, i1, i2),
+            }
```
ondrowan, line 459: Wouldn't it be cleaner if you had array `[i1, i2, i3]`, sorted it and then moved elements 0, 1 and 2 into tuple?

src/geometry.rs
janper: Not sure right now, but the whole hassle with the face was for the possibility of deduplication and HashMaps / HashSets. Maybe we don't have any right now, but we certainly will.
Should I remove it now? I'm worried we'll forget it is possible when we need it.

src/geometry.rs
janper: Nononono (unless I misunderstood you). This is not about sorting! I'm only finding the lowest index and from there on, the order remains unchanged.

src/geometry.rs
ondrowan: Keep it if you're sure we'll use it.

src/geometry.rs
ondrowan: Carry on then.

src/geometry.rs
```diff
@@ -492,9 +517,33 @@ impl TriangleFace {
...
+            self.vertices.2,
+            self.vertices.1,
+            self.vertices.0,
+            self.normals.2,
+            self.normals.1,
+            self.normals.0,
+        )
+    }
+
+    /// Checks if the other face is the references the same vertices and normals
```
yanchith: "other face is the references" -> "other face references"

src/geometry.rs
```diff
@@ -1086,10 +1139,13 @@ mod tests {
             v(-1.0, 1.0, 0.0, [0.0, 0.0, 0.0], 1.0),
         ];
 
+        // When comparing TriangleFaces or Faces from Geometry, make sure the
+        // manually defined faces start their winding from the lowers vertex
```
yanchith: "lowers" -> "lowest", also for every comment you added for this. Btw nice comment 👍

src/interpreter_funcs.rs
```diff
@@ -361,6 +388,45 @@ impl Func for FuncImplLoopSubdivision {
...
+
+        if !mesh_analysis::is_mesh_orientable(&edge_sharing_map)
+            && mesh_analysis::is_mesh_manifold(&edge_sharing_map)
+        {
+            let unoriented_edges: Vec<_> = geometry.unoriented_edges_iter().collect();
+            let edge_to_face =
+                mesh_topology_analysis::edge_to_face_topology(&geometry, &unoriented_edges);
+
+            Ok(Value::Geometry(Arc::new(
+                mesh_tools::synchronize_mesh_winding(&geometry, &unoriented_edges, &edge_to_face),
```
yanchith: Just for readability please extract this to a variable called `value`.

src/mesh_tools.rs
```diff
@@ -1,12 +1,161 @@
...
-use crate::mesh_topology_analysis::face_to_face_topology;
+use crate::geometry::{Face, Geometry, OrientedEdge, TriangleFace, UnorientedEdge};
+use crate::mesh_topology_analysis;
+
+/// Makes sure all the faces are oriented the same way - have the same winding
+/// (vertex order).
+///
+/// This function crawls the mesh geometry and flips all the faces, which are
+/// not facing the same way as the previous faces in the process, starting with
+/// the first face in the list. As a result, the entre mesh can end up facing
```
yanchith: "entire", we are not a restaurant :))

src/mesh_tools.rs
```diff
@@ -567,6 +742,42 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_mesh_tools_revert_mesh_faces() {
+        let test_geometry = geometry::cube_sharp([0.0, 0.0, 0.0], 1.0);
```
yanchith: I'd rather test this on the smallest possible geometry and instead of the assert with `all` and `any` (which is not a sufficient test of correctness) just directly compare the faces. The smallest geometry for which this makes sense is a plane/quad, or maybe even just a single triangle.

src/mesh_tools.rs
```diff
@@ -567,6 +742,42 @@ mod tests {
...
+        assert!(calculated_geometry.faces().iter().all(|face| match face {
+            Face::Triangle(f) => test_geometry
+                .faces()
+                .iter()
+                .any(|other_face| match other_face {
+                    Face::Triangle(o_f) => o_f.is_reverted(f),
+                }),
+        }));
+    }
+
```
yanchith, line 766: I'd add two more simple tests (let's test this on cube or uv_sphere):

- reverting once should not be equal to original
- reverting twice should be equal to original

src/mesh_tools.rs
```diff
@@ -1,12 +1,161 @@
...
+        // of the rest of the mesh geometry.
+        let mut current_face_index = cast_u32(
+            face_processed_pattern
+                .iter()
+                .position(|&v| !v)
+                .expect("Face index not found"),
+        );
+
+        face_processed_pattern[cast_usize(current_face_index)] = true;
+        let mut current_test_edge: OrientedEdge =
```
yanchith: ~Why not push to the edge_face_stack for all edges, but only the first?~ Nevermind, I see, but please document that you are adding the first triggering edge with the self-face just to trigger the rest of the algorithm in a modular way.

src/mesh_tools.rs
```diff
@@ -1,12 +1,161 @@
...
+            original_triangle_faces[cast_usize(current_face_index)].to_oriented_edges()[0];
+
+        // Put the first edge-face couple to the stack
+        edge_face_stack.push((current_test_edge, current_face_index));
+
+        // Check and revert (if needed) faces in the stack as long as there is any
+        while let Some(next_edge_face) = edge_face_stack.pop() {
```
yanchith: You can destructure right here for clarity: `while let Some((next_edge, next_face)) = ...`

src/mesh_tools.rs
```diff
@@ -1,12 +1,161 @@
...
+        .collect();
+
+    // item index = face index; TRUE = the face was already checked, FALSE = the
+    // face hasn't yet been checked
+    let mut face_processed_pattern = vec![false; original_triangle_faces.len()];
+
+    // Faces to be checked for winding, determined by the orientation of the
+    // OrientedEdge of the neighbor, which triggered the check. Current face has
+    // to contain a reverted edge to have the proper winding, otherwise it has
+    // to be reverted.
```
yanchith: I think this comment is outdated. The edge this stack contains is already reverted, so the "current face" has to contain an edge equal to this one.

src/mesh_tools.rs
```diff
@@ -1,12 +1,161 @@
...
+
+            // Find the indices of edges of the current face in the list of edges,
+            // from which the topology was created
+            let face_unoriented_edges = proper_current_face.to_unoriented_edges();
+            let face_unoriented_edge_indices =
+                face_unoriented_edges.iter().map(|unoriented_edge| {
+                    unoriented_edge_index_map
+                        .get(unoriented_edge)
+                        .expect("The current edge is not found in the edge collection")
+                });
```
yanchith: Let's get the index in the loop below and iterate over `face_unoriented_edges`. Should behave the same, but be more readable.

src/mesh_tools.rs
```diff
@@ -1,12 +1,161 @@
...
+
+            // For each face edge index
+            for (i, face_unoriented_edge_index) in face_unoriented_edge_indices.enumerate() {
+                // get the actual oriented edge
+                let face_oriented_edge = face_unoriented_edges[i].0;
+                // and try to find it in the edge-to-face topology.
```
yanchith: Misleading comment: This does nothing with the `face_oriented_edge` above. Instead should be something like: "Find faces for current unoriented edge"

src/mesh_tools.rs
```diff
@@ -1,12 +1,161 @@
...
+                    .get(&face_unoriented_edge_index)
+                    .expect("Edge not found in the topology");
+                // If it exists, iterate the faces containing the edge
+                for face_index in edge_in_faces {
+                    // and if it was not already added to the stack or even checked
+                    if !face_processed_pattern[cast_usize(*face_index)] {
+                        // add it to the stack with the expected edge orientation
+                        edge_face_stack.push((face_oriented_edge.to_reverted(), *face_index));
+                        // and mark it treated.
+                        face_processed_pattern[cast_usize(*face_index)] = true;
```
yanchith: This is a bit misleading as well... Usually graph search algorithms mark something as visited only after they are done with it, but this algorithm does it pre-emptively.

I would suggest the following modifications:

- Check here if `face_index == current_face_index` and don't add if it is
- Check here if the face has already been marked processed and don't add if it has been (you already do this)
- Mark faces as processed only after they are really processed
- Don't add the face to edge_face_stack if it is already contained. For this point it is worth mentioning, that it probably should no longer be a `Vec`, but rather a `HashSet` so that this check is not too expansive. Also, it would make sense to rename edge_face_stack to edge_face_pairs_to_process while doing it.

src/mesh_tools.rs
```diff
@@ -1,12 +1,161 @@
...
+            original_triangle_faces[cast_usize(current_face_index)].to_oriented_edges()[0];
+
+        // Put the first edge-face couple to the stack
+        edge_face_stack.push((current_test_edge, current_face_index));
+
+        // Check and revert (if needed) faces in the stack as long as there is any
+        while let Some(next_edge_face) = edge_face_stack.pop() {
+            // Get ready for the next iteration.
+            current_test_edge = next_edge_face.0;
+            current_face_index = next_edge_face.1;
```
yanchith: Any particular reason for mutating these instead of declaring local variables?

yanchith (changes requested):
Almost there! There's a couple of typos that need fixing here and there, and I think the face winding sync algorithm can be simplified. I tried to express simplification suggestions in inline comments, but when in doubt, this is the underlying graph search algo:  https://en.wikipedia.org/wiki/Breadth-first_search from which we should use the terminology and approaches where applicable.

Also, as we discussed in person, we need to check the diffs for the snapshots:

- The laplacian smoothing snapshots have different vertex index winding, which is to be expected, and also some slight floating point precision changes. Those can be explained by the change in iteration (and therefore math operation) order we introduced, when we forced the face winding order. 

- Something similar is most likely be true for the loop subdivision snapshots too, but let's at least check.

It is enough to check the diffs for 1_iterations snapshots for both.

src/mesh_tools.rs
janper: but we should be!

src/mesh_topology_analysis.rs
```diff
@@ -3,13 +3,13 @@ use std::collections::HashMap;
...
-use crate::geometry::{Face, Geometry, UnorientedEdge};
+use crate::geometry::{Face, Geometry, OrientedEdge, UnorientedEdge};
 
 /// Calculates topological relations (neighborhood) of mesh face -> faces.
 /// Returns a Map (key: face index, value: list of its neighboring faces indices)
 #[allow(dead_code)]
-pub fn face_to_face_topology(geometry: &Geometry) -> HashMap<u32, SmallVec<[u32; 8]>> {
-    let mut f2f: HashMap<u32, SmallVec<[u32; 8]>> = HashMap::new();
+pub fn face_to_face_topology(geometry: &Geometry) -> HashMap<u32, SmallVec<[u32; 1]>> {
+    let mut f2f: HashMap<u32, SmallVec<[u32; 1]>> = HashMap::new();
```
yanchith: I think you meant 3, not 1.

Also, since we are now no longer using a flat constant 8 everywhere (which was easy to remember and review), we should attack this more systematically so that we don't trip over it all the time.

Let's define some terminology (we may come up with other terms for this later):

- manifold maximum of relations: the maximum number of relations that applies to only manifold meshes,
- hard maximum of relations: the maximum number of relations that applies to all meshes.

For topologies where we know the hard maximum, let's use `ArrayVec` instead of `SmallVec`, because it panics if you try to push beyond its inline capacity.

For the current topologies, that is:

- `face_to_face` has manifold maximum of 3 (or 4 once we add quads) => `SmallVec<[u32; 3]>` with a FIXME for quads
- `edge_to_face` has manifold maximum of 2 => `SmallVec<[u32; 2]>`
- the newly created `face_to_oriented_edge` has hard maximum of 3 (or 4 once we add quads) => `ArrayVec<[u32; 3]>` with a FIXME for quads
- the newly created `unoriented_edge_to_face` has manifold maximum of 2 => `SmallVec<[u32; 2]>` (btw isn't this function exactly the same as edge_to_face?)

We don't know the exact maximum for `vertex_to_*` topologies, and there we should keep using `SmallVec<[u32, 8]>` for now, at least until we benchmark.

I'd like to fix this in 2 steps, for every topology:

- Create a type alias for each topology, e.g.: `pub type FaceToFaceTopology = HashMap<u32, SmallVec<[u32; 3]>>` and make the creating functions return it.
- Where we know the hard maximum of relations, use `ArrayVec` instead of  `SmallVec`

src/mesh_topology_analysis.rs
```diff
@@ -3,13 +3,13 @@ use std::collections::HashMap;
 use smallvec::SmallVec;
 
 use crate::convert::cast_u32;
-use crate::geometry::{Face, Geometry, UnorientedEdge};
+use crate::geometry::{Face, Geometry, OrientedEdge, UnorientedEdge};
 
 /// Calculates topological relations (neighborhood) of mesh face -> faces.
```
yanchith: Let's add to the doc comment what it means when 2 faces are related - that is, they are related when they share an edge (and not just a vertex)

src/mesh_topology_analysis.rs
```diff
@@ -77,6 +77,58 @@ pub fn edge_to_face_topology(
...
+                            .position(|e| e == o_e)
+                            .expect("The edge not found in the list"),
+                    )
+                })
+                .collect(),
+        ),
+    })
+}
+
+/// Calculates topological relations (containment) of mesh oriented edge ->
```
yanchith: Shouldn't this be "unoriented edge" in the comment? Also, what is the difference between this and the already existing `edge_to_face`, that also uses unoriented edges? If they are semantically the same, let's only keep that one (it is already tested) and rename it to `unoriented_edge_to_face_topology`

src/interpreter_funcs.rs
```diff
@@ -415,12 +415,25 @@ impl Func for FuncImplSynchronizeMeshFaces {
             && mesh_analysis::is_mesh_manifold(&edge_sharing_map)
         {
             let unoriented_edges: Vec<_> = geometry.unoriented_edges_iter().collect();
```
yanchith: I missed this last time, but shouldn't this be de-duplicated? Or is it ok if it is not? Also for every other usage of this in the PR. This is an open-ended question and depending on the answer a lot of what I say in this review doesn't apply, and we have a much deeper problem.

src/mesh_tools.rs
```diff
@@ -28,117 +23,161 @@ use crate::mesh_topology_analysis;
 ///
 /// This method doesn't flip the normals associated with the face vertices, as
 /// there is no unambiguous way to do so automatically.
-
+///
+/// https://gist.github.com/vTurbine/16fbb99225ad4c0ac80b24855dd61a7c
```
yanchith: Let's not add this link.

src/mesh_tools.rs
```diff
@@ -764,18 +803,87 @@ mod tests {
     }
 
     #[test]
-    fn test_mesh_tools_synchronize_mesh_winding() {
+    fn test_mesh_tools_synchronize_mesh_winding_bfs() {
```
yanchith: No need to suffix "bfs". It's just an implementation detail. Also for all newly added tests.

src/mesh_tools.rs
```diff
@@ -764,18 +803,87 @@ mod tests {
...
+            &face_to_oriented_edge_topo,
+            &unoriented_edge_to_face_topo,
+        );
 
-        let calculated_geometry =
-            synchronize_mesh_winding(&test_geometry, &unoriented_edges, &edge_to_face);
+        assert!(mesh_analysis::are_similar(
+            &test_geometry_correct,
+            &calculated_geometry
+        ));
```
yanchith: This changed assert_eq! to assert!(are_similar). Why? These tests can know about implementation details of the algorithm, e.g. which face it picks as it's first one, so we should be able to assert exactly.

src/mesh_tools.rs
```diff
@@ -764,18 +803,87 @@ mod tests {
...
+        );
+
+        assert!(!mesh_analysis::are_similar(
+            &test_geometry_correct,
+            &test_geometry
+        ));
+        assert!(mesh_analysis::are_similar(
+            &test_geometry_correct,
+            &calculated_geometry
+        ));
```
yanchith: Isn't assert_eq! enough? These tests can theoretically know about implementation details of the algorithm, e.g. which face it picks as it's first one.

src/mesh_topology_analysis.rs
```diff
@@ -77,6 +77,58 @@ pub fn edge_to_face_topology(
...
+/// Calculates topological relations (containment) of mesh face -> its oriented
+/// edges.
+///
+/// Returns an iterator (index: face index, value: list of indices of oriented
+/// edges it contains
+#[allow(dead_code)]
+pub fn face_to_oriented_edge_topology<'a>(
+    geometry: &'a Geometry,
+    oriented_edges: &'a [OrientedEdge],
+) -> impl Iterator<Item = SmallVec<[u32; 3]>> + 'a {
```
yanchith: What is the added benefit of returning iterator? To able to use the the topology, it has to be collected to a Vec anyway. Same for `unoriented_edge_to_face_topology`.

src/geometry.rs
```diff
@@ -306,6 +306,10 @@ impl Geometry {
         &self.normals
     }
 
+    pub fn faces_len(&self) -> usize {
+        self.faces.len()
+    }
```
yanchith: Let's not add this helper.

src/mesh_tools.rs
```diff
@@ -1,12 +1,200 @@
...
+    let oriented_to_unoriented_edge_index: Vec<_> = oriented_edges
+        .iter()
+        .map(|o_e| {
+            let to_unoriented = o_e.to_unoriented();
+            unoriented_edges
+                .iter()
+                .position(|u_e| *u_e == to_unoriented)
+                .expect("No unoriented edge matches the oriented")
+        })
+        .collect();
```
yanchith: If I understand this correctly, unoriented faces and oriented faces are the same length in here, and they share the same indices, becuase we didn't deduplicate. Meaning this should not be unnecessary.

Also, it is quadratic and will take a long time to compute.

src/mesh_tools.rs
```diff
@@ -1,12 +1,200 @@
...
+
+    let original_triangle_faces: Vec<_> = geometry
+        .faces()
+        .iter()
+        .map(|f| match f {
+            Face::Triangle(t_f) => *t_f,
+        })
+        .collect();
```
yanchith: Isn't this an allocation just for a convenience's sake? You can get to the original triangle face through the original geometry, but will have to pattern match.

src/mesh_tools.rs
```diff
@@ -1,12 +1,200 @@
...
+
+    // For each island in the mesh geometry
+    while proper_faces.len() < original_triangle_faces.len() {
+        // find first unvisited face
+        let current_face_index = cast_u32(
```
yanchith: You are casting to u32 here only to cast to usize later on. Remove this cast and many things should be simplified.

src/mesh_tools.rs
```diff
@@ -1,12 +1,200 @@
...
+    geometry: &Geometry,
+    oriented_edges: &[OrientedEdge],
+    unoriented_edges: &[UnorientedEdge],
+    face_to_oriented_edge_topo: &[SmallVec<[u32; 3]>],
+    unoriented_edge_to_face_topo: &[SmallVec<[u32; 2]>],
+) -> Geometry {
+    // Processing queue: indices of faces sharing edges with the current face,
+    // zipped with the oriented edge they should contain if they have a proper
+    // winding.
+    let mut face_indices_with_desired_oriented_edge_queue: VecDeque<(u32, OrientedEdge)> =
```
yanchith: Let's just call it `queue_to_process` or something simple.

src/mesh_tools.rs
```diff
@@ -1,12 +1,200 @@
...
+        // add it to the processing queue.
+        face_indices_with_desired_oriented_edge_queue.push_front((
+            // the face
+            current_face_index,
+            // one of the edges it should contain
+            oriented_edges
+                [cast_usize(face_to_oriented_edge_topo[cast_usize(current_face_index)][0])],
+        ));
+        // Mark the face already inserted into the queue.
+        visited[cast_usize(current_face_index)] = true;
```
yanchith: You should not mark it as visited here, but only as discovered. These are 2 different things.

You need an additional data structure, guard against duplication in the processing queue. This new datastructure should be a `HashSet`, because it will most likely never grow to be the full size of faces, unlike `visited`, which is correctly a `Vec`.

src/mesh_tools.rs
```diff
@@ -1,12 +1,200 @@
...
+            let is_properly_winded = face.contains_oriented_edge(desired_oriented_edge);
+
+            // Put the original or reverted face into the stack of processed
+            // faces.
+            if is_properly_winded {
+                proper_faces.push(face);
+            } else {
+                proper_faces.push(face.to_reverted());
+            }
+
```
yanchith: Here would be the place to finally mark the face as visited, and potentially remove it from discovered (to keep the discovered collection small), because we already processed it.

src/mesh_tools.rs
```diff
@@ -1,12 +1,200 @@
...
+                // get the indices of faces containing the edge. For a proper
+                // watertight mesh, this should contain exactly two face
+                // indices, including the index of the current face. For a face
+                // at the naked edge a mesh patch, this only contains one value:
+                // the index of the current face. For non-manifold edges this
+                // contains more than 2 values and the result of the entire
+                // method is unpredictable.
+                //
+                // FIXME: Throw a warning if the mesh is non-manifold. The
+                // result might be wrong.
```
yanchith: I think we can remove this FIXME. The function warns about its unpredictable behaviour for non-manifold meshes the documentation. Keeping it garbage-in, garbage-out is ok.

src/mesh_tools.rs
```diff
@@ -28,117 +23,161 @@ use crate::mesh_topology_analysis;
...
 // FIXME: Flip also vertex normals if the visual/practical tests prove it's
 // needed
+#[allow(dead_code)]
 pub fn synchronize_mesh_winding(
     geometry: &Geometry,
+    oriented_edges: &[OrientedEdge],
     unoriented_edges: &[UnorientedEdge],
-    edge_to_face_topology: &HashMap<u32, SmallVec<[u32; 8]>>,
+    face_to_oriented_edge_topo: &[SmallVec<[u32; 3]>],
+    unoriented_edge_to_face_topo: &[SmallVec<[u32; 2]>],
```
yanchith: To be honest, it is not very clear to me why we need both oriented and unoriented edges now.

src/mesh_tools.rs
```diff
@@ -1,12 +1,200 @@
...
+/// topologies.
+///
+// FIXME: Flip also vertex normals if the visual/practical tests prove it's
+// needed
+#[allow(dead_code)]
+pub fn synchronize_mesh_winding(
+    geometry: &Geometry,
+    oriented_edges: &[OrientedEdge],
+    unoriented_edges: &[UnorientedEdge],
+    face_to_oriented_edge_topo: &[SmallVec<[u32; 3]>],
```
yanchith: What is the purpose of this topology at all? You can get both oriented and unoriented faces directly from the face. No need to have a map store it for you. I tried to go over all usages and I think mostly you just introduce additional indirection by using an index where you could use the the value directly. I have inline comments in those places.

src/mesh_tools.rs
```diff
@@ -1,12 +1,200 @@
...
+                .position(|v| !v)
+                .expect("All faces already visited"),
+        );
+
+        // add it to the processing queue.
+        face_indices_with_desired_oriented_edge_queue.push_front((
+            // the face
+            current_face_index,
+            // one of the edges it should contain
+            oriented_edges
```
yanchith: Can't you just extract the oriented edge directly from the face?

yanchith (changes requested):
This is just a preliminary review. Too much changed and I ran out of time. Summary:

- The algorithm itself is now much more readable and understandable, cool :+1: 

- The algorithm needs a notion of "discovered" faces so we can can guard against adding a face for processing multiple times (see inline comment for more) 

- If we are going to start changing sizes of smallvecs arbitrarily, we should document why that is ok for each case (see my inline comment for in detail suggestions)

- The newly added `unoriented_edge_to_face_topology` seems like a duplicate of the already existing `edge_to_face` topology, or at least the difference is so small that I missed it, in which case it requires better documentation (also see inline comment for suggestions)

- I believe, but am not 100% sure the `face_to_oriented_edge` topology is not needed at all and all cases that use it could just extract the data directly from the face, at least in the code as presented by this PR.

I think I missed a lot of things and will have to go through this again another day. Please try to fix the most obvious errors meanwhile.

src/mesh_topology_analysis.rs
janper: Agree, however I'm not sure there is any hard maximum anywhere so far. Once i come along such a case, I'll use the ArrayVec

src/mesh_topology_analysis.rs
janper: The output of the new function is more suitable for the algorithm which it uses. Semantically they are similar, the new one is more uptodate and during the next refactor we'll change the rest of the topology calculators to match this one. I prefer keeping the new one and adjusting all the functions and test to use it.

src/interpreter_funcs.rs
janper: You're right, they should be deduplicated. Nevertheless, nothing fails if they are not, just the calculations might be less efficient. 
I prefer some sort of deduplication already in the `unoriented_edges_iter` method. Once we get rid of the custom `Eq` on the `UnorientedEdge`, this will be also more efficient.
When changing the `UnorientedEdge` I'd like to add a common getter that would return `OrientedEdge`s, `UnorientedEdge`s and some `oriented_edge_index_to_unoriented_edge_index_map` at the same time. They are and will be useful together and they are also quite efficient to be calculated at the same time.

src/mesh_tools.rs
janper: The algorithm crawls the geometry and returns the faces in an arbitrary order. If not using the the similarity check, we need to use snapshot tests. I can do that if you think it's better.

src/mesh_tools.rs
janper: The first one yes, the rest not. Snapshot?

src/mesh_topology_analysis.rs
janper: I can imagine requesting the iterator much sooner (when the geometry is created) and only using it when/if needed. Returning a Vec is ok too, if it makes more sense.

src/mesh_tools.rs
janper: True and not deduplicating `UnorientedEdge`s is a mistake. Should be fixed. It is quadratic but the calculation would be needed anyway - either collected into a data structure like this or calculated as a logic where needed (and every item of the list will be needed).
As I mentioned above, it would be way more efficient if calculated together with `OrientedEdge`s and `UnorientedEdge`s. Then we could define a new topology type containing all three of these structures.

src/mesh_tools.rs
janper: Yes. The memory use is probably higher higher but aside of that, is there any drawback? I have to access each item anyway.

src/mesh_tools.rs
janper: I thought we are always using u32 for indices. Does this only apply to the public things?

src/mesh_tools.rs
janper: Should I then give a name to the type it contains? It's not self-explanatory.

src/mesh_tools.rs
janper: Maybe a better name for `visited` would be `added_to_queue` but I don't see how can the queue contain duplicate items? Maybe I'm missing something, but I'm pretty sure it never happens. See line 168.

src/mesh_tools.rs
janper: Why? The discovered faces are stored in `face_indices_with_desired_oriented_edge_queue`, they are marked in `visited` and processed are stored in `proper_faces`. I agree to change the variable names, if it makes more sense.

src/mesh_tools.rs
janper: We need `UnorientedEdge`s to find face's neighbors and identify the edge they share. It's also the edge that should be reverted in each of the faces. We might get rid of `oriented_edges` here somehow though: for example there can be a method of the `TriangleFace` that would take an `UnorientedEdge` as an argument and return a respective `OrientedEdge`. It might be actually quite efficient but it's a different approach.

src/mesh_tools.rs
janper: Exactly, it contains indices, so that you don't need to use the actual data. The variable name should be more explanatory.

src/mesh_tools.rs
janper: Yes, in this case I can but later in the function I do it this way, so trying to keep consistent.

src/mesh_tools.rs
yanchith: > Does this only apply to the public things?

In practices yes

src/mesh_tools.rs
yanchith: What you did with the comment is fine

yanchith:
@janper, as agreed offline, I implemented my own review requests in the most minimal way. Please have a look. I couldn't add you as a reviewer because you are the PR creator, but you should be able to review and leave comments regardless.

Especially please review the diff of my new changes against yours: https://github.com/subdgtl/HURBAN-Selector/pull/62/files/28f772d0807f6894b682aa93c9c4af0dc311b33f..4806810f5e95d353533e26742044e939355e2497

yanchith (dismissed):

src/mesh_analysis.rs
```diff
@@ -776,6 +780,36 @@ mod tests {
...
+    fn test_mesh_analysis_is_mesh_manifold_returns_false_because_non_manifold() {
+        let (faces, vertices) = non_manifold_shape();
+        let geometry = Geometry::from_triangle_faces_with_vertices_and_computed_normals(
+            faces.clone(),
+            vertices.clone(),
+            NormalStrategy::Sharp,
+        );
+
+        let oriented_edges: Vec<OrientedEdge> = geometry.oriented_edges_iter().collect();
+        let edge_sharing_map = edge_analysis::edge_sharing(&oriented_edges);
```
janper, line 793: For functions like this, when you need to request some other data (here `oriented_edges_iter`) first to be able to request the data you actually need (here `edge_analysis::edge_sharing`), shouldn't we also have an extra function doing this directly from `Geometry`? The current functions will remain and will be used for optimization purposes (reuse of the data) when it makes sense, but in most cases we won't need them.

src/mesh_tools.rs
```diff
@@ -1,12 +1,159 @@
...
+                                    break;
+                                }
+                            }
+                        }
+                    }
+
+                    assert!(
+                        found,
+                        "Face to face topology and face disagree on 2 faces being neighbors",
+                    );
```
janper: This can only happen if the `face_to_face_topology` generator contains an error. Shouldn't we make sure it works well rather than asserting here? 
Also, there is no way of making sure the topology really belongs to the given geometry and this doesn't seem like a reliable way for testing that.

src/mesh_topology_analysis.rs
```diff
@@ -3,12 +3,25 @@ use std::collections::HashMap;
...
-pub fn face_to_face_topology(geometry: &Geometry) -> HashMap<u32, SmallVec<[u32; 8]>> {
+// FIXME: @Optimization Explore whether the topologies wouldn't be
+// better served by being backed by a `Vec` instead of
+// `HashMap`. Ideally, we'd also create a wrapper struct that casts
+// the indices to/from u32 as necessary.
+
+// FIXME: @Optimization Analyze where this threshold is overly
+// benevolent and define different thresholds for different
+// topologies.
+const NEIGHBORS_ALLOCATION_THRESHOLD: usize = 8;
```
janper: To me this name isn't very explanatory. I'd prefer a more elaborate version of something like `MAX_NEIGHBORS_IN_PROPER_MESH`.

src/mesh_topology_analysis.rs
```diff
@@ -3,12 +3,25 @@ use std::collections::HashMap;
...
+
+/// A topology containing neighborhood relations between faces. Two
+/// faces are neighbors if and only if they they share an unoriented
+/// edge.
+pub type FaceToFaceTopology = HashMap<u32, SmallVec<[u32; NEIGHBORS_ALLOCATION_THRESHOLD]>>;
+
+/// Computes vertex to vertex topology for geometry.
```
janper: face to face

src/mesh_topology_analysis.rs
```diff
@@ -44,91 +57,13 @@ pub fn face_to_face_topology(geometry: &Geometry) -> HashMap<u32, SmallVec<[u32;
...
-                }
-            }
-        }
-
-        e2f.insert(from_edge_index_u32, neighbors);
-    }
+/// A topology containing neighborhood relations between vertices. Two
+/// vertices are neighbors if and only if they they share an
+/// unoriented or oriented edge.
+pub type VertexToVertexTopology = HashMap<u32, SmallVec<[u32; NEIGHBORS_ALLOCATION_THRESHOLD]>>;
```
janper: are end points of an edge

janper (commented):
I can't request changes, nor approve, so I'm just leaving inline comments.

src/mesh_topology_analysis.rs
janper: Or just comment

src/mesh_tools.rs
yanchith: You are right. This was not correct to assert.

src/mesh_analysis.rs
yanchith: We can make a task to identify when this makes sense and do it. Or just do it 😸

yanchith: Feel free to submit small fixes like this any time. **In a separate PR**

yanchith (approved):
Good job 😉

PR #63, opened 2019-11-02
Implement Loop Subdivision

yanchith:
This adds the loop subdivision functionality, also as an interpreter
func. Functionality was implemented based on
https://graphics.stanford.edu/~mdfisher/subdivision.html

This is a draft, because I still want to resolve the remaining TODO, which is to add a plane/quad/triangle or some other simple primitive as a interpreter func so this can be visually tested on geometry containing naked edges. I do consider the functionality final though.

This can be tested by running the hardcoded program in `lib.rs` (run and press `R`). After review we can discard the modifications this does to `lib.rs`.

@ondrowan to save your time and priorities, don't review this yet, let @janper review the geometry logic first.


src/geometry.rs
```diff
@@ -396,6 +396,14 @@ pub enum Face {
     Triangle(TriangleFace),
 }
 
+impl Face {
+    pub fn contains_vertex(&self, vertex_index: u32) -> bool {
```
janper, line 400: This is ok but let's then look-up all the uses of the `TriangleFace` equivalent of this (and other functions too) and think about replacing all of them with methods on this level. Generally, let's use `TriangleFace` only where other types of faces wouldn't work.

src/interpreter_funcs.rs
```diff
@@ -199,7 +199,58 @@ impl Func for FuncImplSeparateIsolatedMeshes {
...
+        if iterations == 0 {
+            return Value::Geometry(geometry);
+        }
+
+        let mut v2v_topology = mesh_topology_analysis::vertex_to_vertex_topology(&geometry);
+        let mut f2f_topology = mesh_topology_analysis::face_to_face_topology(&geometry);
+        let mut current_geometry =
+            mesh_smoothing::loop_subdivision(&geometry, &v2v_topology, &f2f_topology);
+
+        for _ in 1..iterations {
```
janper, line 312: Please, make sure no more than 3 iterations are requested in one operation because anything more can and will result in an enormous mesh and slow calculation times. 3 iterations are an industry standard and even that usually takes a long time (seconds) to calculate. If it's is not enough, it's always possible to stack another smoothing operation into the pipeline.

src/interpreter_funcs.rs
```diff
@@ -199,7 +199,58 @@ impl Func for FuncImplSeparateIsolatedMeshes {
...
+    }
+
+    fn return_ty(&self) -> Ty {
+        Ty::Geometry
+    }
+
+    fn call(&self, args: &[Value]) -> Value {
+        let geometry = args[0].unwrap_refcounted_geometry();
+        let iterations = args[1].unwrap_uint();
+        if iterations == 0 {
```
janper, line 303: Wouldn't it help to check if the geometry is triangulated? We already have a method for that.

src/mesh_smoothing.rs
```diff
@@ -91,6 +94,267 @@ pub fn laplacian_smoothing(
...
+            cmp::min(self.0, self.1).hash(state);
+            cmp::max(self.0, self.1).hash(state);
+        }
+    }
+
+    let mut vertices: Vec<Point3<f32>> = geometry.vertices().iter().copied().collect();
+
+    // Relocate existing vertices first
+    for (i, vertex) in vertices.iter_mut().enumerate() {
+        let neighbors = &vertex_to_vertex_topology[&cast_u32(i)];
```
janper, line 141: Is this an equivalent of `&vertex_to_vertex_topology.get(&cast_u32(i))`? Also, you can't be 100% sure the `vertex_to_vertex_topology` belongs to the `geometry` because it comes from outside.

src/mesh_smoothing.rs
```diff
@@ -91,6 +94,267 @@ pub fn laplacian_smoothing(
...
+                let v1 = geometry.vertices()[vi1];
+                let v2 = geometry.vertices()[vi2];
+
+                *vertex = Point3::origin()
+                    + v0.coords * 3.0 / 4.0
+                    + v1.coords * 1.0 / 8.0
+                    + v2.coords * 1.0 / 8.0;
+            }
+            3 => {
+                // N == 3 means the vertex is not a part of a naked edge.
```
janper: Why not? It might.

src/mesh_smoothing.rs
```diff
@@ -91,6 +94,267 @@ pub fn laplacian_smoothing(
...
+                let v3 = geometry.vertices()[vi3];
+
+                *vertex = Point3::origin()
+                    + vertex.coords * (1.0 - N * BETA)
+                    + v1.coords * BETA
+                    + v2.coords * BETA
+                    + v3.coords * BETA;
+            }
+            n => {
+                // N >= 3 also means the vertex is not a part of a naked edge.
```
janper: Still seems like it might. We already have a `get_naked_vertices` function.

src/mesh_smoothing.rs
```diff
@@ -91,6 +94,267 @@ pub fn laplacian_smoothing(
...
+    // The key is an unordered pair of faces that share the mid-edge
+    // vertex. The value is the index of the vertex they share.
+    let mut created_mid_vertex_indices: HashMap<UnorderedPair, u32> = HashMap::new();
+
+    for (face_index, face) in geometry.faces().iter().enumerate() {
+        let face_index_u32 = cast_u32(face_index);
+        match face {
+            Face::Triangle(triangle_face) => {
+                let (vi1, vi2, vi3) = triangle_face.vertices;
+                let face_neighbors = &face_to_face_topology[&face_index_u32];
```
janper, line 216: ~Also here you can't be sure a proper `face_to_face_topology` is passed as an argument.~

src/geometry.rs
yanchith: I did. This is the first use of the any of the `contains_vertex` functions 😂 (at least of what is merged to master)

src/interpreter_funcs.rs
yanchith: Okay. I'll add a clamp for now, and we can provide a range for the UI later.

src/interpreter_funcs.rs
yanchith: The question is, what do we do if it is not? Skip or triangulate? The geometry I am testing on is basically 90% triangulated quads. The subdivision could even potentially triangulate internally for other faces.

The more I think of this, the less I am eager to add quads 😸

src/mesh_smoothing.rs
yanchith: > Is this an equivalent of &vertex_to_vertex_topology.get(&cast_u32(i))

Yes, followed by an unwrap.

> Also, you can't be 100% sure the vertex_to_vertex_topology belongs to the geometry because it comes from outside.

Then you get a panic resulting from an incorrect function use - a programmer error. A panic is what you want in this case. We could potentially ignore the vertex if it is not present in the topology, but that would cover up the underlying issue, not a fan of that solution.

I can change one thing though: since the topologies can not be re-used between the iterations (at least not in a simple way), we can just generate them in the function itself and this should be less of an issue (although, there still can be a bug in the topology itself - but once again, we would want that to surface as a panic while we are still developing).

src/mesh_smoothing.rs
yanchith: Damn... my source article did not cover this case or provide a relocation scheme for it. Do we want handle that case somewhat specially, or should I just update the comment?

src/mesh_smoothing.rs
```diff
@@ -91,6 +94,267 @@ pub fn laplacian_smoothing(
...
+
+                let vi1 = cast_usize(neighbors[0]);
+                let vi2 = cast_usize(neighbors[1]);
+
+                let v0 = geometry.vertices()[i];
+                let v1 = geometry.vertices()[vi1];
+                let v2 = geometry.vertices()[vi2];
+
+                *vertex = Point3::origin()
+                    + v0.coords * 3.0 / 4.0
```
yanchith: Change `v0.coords` to `vertex.coords` and get rid of `v0`

src/mesh_smoothing.rs
janper: sure, let's keep the getter as is. if it properly panics, then it's good.
i'd keep the topology outside: 1. there is a chance of overusing an existing one; 2. we'll have it that way in other functions, too

src/mesh_smoothing.rs
janper: If your algorithm cares about edge valency, then it's good as it is and I suggest changing just the comments. But if you actually need to identify vertices on the borders, then please use one of the existing functions doing exactly that. 

A side note: it might be useful to have the option to preserve the edges during smoothing. It's not that much extra work and we have the same functionality for the Laplacian smoothing.

src/mesh_smoothing.rs
yanchith: > 1. there is a change of overusing an existing one;

Good point, I'll keep it a parameter.

src/mesh_smoothing.rs
yanchith: > If your algorithm cares about edge valency, then it's good as it is and I suggest changing just the comments.

Naked edges were not specified in sources I read. It only cares about valency.

yanchith: > A side note: it might be useful to have the option to preserve the edges during smoothing. It's not that much extra work and we have the same functionality for the Laplacian smoothing.

I think this should be do-able. It'll just skip the `N == 2` branch if this is enabled. I'll add it.

yanchith: Hmmm, wait. What exactly do you mean by that? Preserve naked edges (not relocate vertices that are part of a naked edge)? Or all edges?

src/interpreter_funcs.rs
janper: i think we might need quads for morphing later on. let's keep the possibility for now. 

in this case i suggest triangulating prior to smoothing

src/mesh_smoothing.rs
janper: > > A side note: it might be useful to have the option to preserve the edges during smoothing. It's not that much extra work and we have the same functionality for the Laplacian smoothing.
> 
> I think this should be do-able. It'll just skip the `N == 2` branch if this is enabled. I'll add it.

that's not enough - it only preserves "corners". you'll need to check for real border vertices. but we can keep it for later

janper: > Hmmm, wait. What exactly do you mean by that? Preserve naked edges (not relocate vertices that are part of a naked edge)? Or all edges?

preserve naked edges, including the newly created vertices in the middle of the existing naked edges.
but as i said, this might be noted as a new feature and implemented later

yanchith: Ok, I'd keep it for later then.

src/interpreter_funcs.rs
yanchith: Okay, but we don't have triangulation yet. So... we'll have to implement it as soon as we add quads. I'll add an `is_triangulated` assert to the `loop_subdivision` function, so that the caller is responsible for triangulating. The operation will have to do the graceful handling, because the user can input anything.

src/geometry.rs
janper: In one of my upcoming PRs I'll do the same for other methods

ondrowan (changes requested):
At least some basic test could be written for now.

yanchith:
@ondrowan 

> At least some basic test could be written for now.

What kind of test do you have in mind exactly?

I believe 2 things could be of use here:

- This could really use snapshot testing to test various behaviors on various meshes, but in our slack discussion we agreed to postpone it.
- I can imagine writing the unit tests for the average case and the edge cases, but in earlier PR discussion we decided to postpone the edge case handling too, because for the average case this works as advertised.

Do you want me to just add the most basic possible test for the average case?

ondrowan:
Yes.

yanchith:
@ondrowan, as discussed on slack, let's take care of #67 first, and then this once again just becomes about adding a simple test.

ondrowan (approved):

PR #64, opened 2019-11-03
Adds UI representation of operations

ondrowan:


src/operation_manager.rs
```diff
@@ -0,0 +1,343 @@
...
+
+    /// Geometries produced by operations and their metadata. Order matters
+    /// for `geometry_stack` to be able to properly work with geometries.
+    geometry_metadata: Vec<GeometryMetadata>,
+
+    /// Since operations can produce multiple geometries, this stack tracks
+    /// number of geometries returned from each operation. This is useful
+    /// for removing operations and tracking what geometries should be
+    /// available for which operations in the pipeline.
+    geometry_stack: Vec<usize>,
```
yanchith, line 135: Smart!

src/operation_manager.rs
```diff
@@ -0,0 +1,343 @@
...
+
+        let mut args: Vec<ast::Expr> = vec![];
+
+        for param in &operation.params {
+            let expr = match param.repr {
+                OpParamUiRepr::Dropdown(_) => match param.value {
+                    ast::LitExpr::Uint(uint) => {
+                        ast::Expr::Var(ast::VarExpr::new(VarIdent(u64::from(uint))))
+                    }
+                    _ => ast::Expr::Var(ast::VarExpr::new(VarIdent(0 as u64))),
```
yanchith: Can use `LitExpr::Nil` here possibly?

src/operation_manager.rs
```diff
@@ -0,0 +1,343 @@
...
+                },
+                _ => ast::Expr::Lit(param.value.clone()),
+            };
+            args.push(expr);
+        }
+
+        self.interpreter_server
```
yanchith, line 165: Will need to track this in-flight request in some kind of map, so you can match the response later, similarly below. Lol, have we build our own micro networking thing?

src/operation_manager.rs
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    /// Submits program to interpreter and sets status of last operation
+    /// to "running".
+    pub fn submit_program(&mut self) {
+        let last_op = self
+            .selected_ops
+            .last_mut()
+            .expect("Failed to get last selected operation");
+        last_op.status = OpStatus::Running;
```
yanchith: Shouldn't they all be running, officially? (Interpreter does not expose any cache-ing it internally does).

src/operation_manager.rs
```diff
@@ -0,0 +1,343 @@
...
+#[derive(Debug)]
+pub struct SelectedOp {
+    pub op: Op,
+    pub status: OpStatus,
+}
+
+#[derive(Debug)]
+pub struct GeometryMetadata {
+    name: String,
+    pub geometry: Geometry,
```
yanchith, line 111: Would probably be better if this were an `Arc` that was returned, if we want to clone it around. Once you rebase on top of master, you'll find `Value::unwrap_refcounted_geometry` that gives you the whole `Arc`.

src/operation_manager.rs
```diff
@@ -0,0 +1,343 @@
...
+                            name: format!("Geometry #{} from {}", i + 1, &op_name),
+                            geometry: value.unwrap_geometry().clone(),
+                            var_ident: var_ident.0,
+                            geometry_id: None,
+                            used: false,
+                        };
+
+                        let geometry_id = geometry_handler(&mut geometry_metadata);
+
+                        geometry_metadata.geometry_id = Some(geometry_id);
```
yanchith: Gruesome, but if it works...

src/lib.rs
```diff
@@ -368,6 +279,30 @@ pub fn init_and_run(options: Options) -> ! {
                 #[cfg(debug_assertions)]
                 ui_frame.draw_fps_window();
 
+                operation_ui.prepare_ops();
```
yanchith: Can this happen when the manager is polled for interpreter response?

src/ui.rs
```diff
@@ -144,4 +150,173 @@ impl<'a> UiFrame<'a> {
...
+                            OpParamUiRepr::Dropdown(ref choices) => {
+                                // Imgui's combo box requires usize, but our Uint
+                                // values are u32, so we need to cast them before
+                                // and after.
+                                let mut usize_value: usize = *value as usize;
+                                imgui::ComboBox::new(&label).build_simple(
+                                    ui,
+                                    &mut usize_value,
+                                    choices,
+                                    &|item| std::borrow::Cow::from(imgui::im_str!("{}", item.1)),
```
yanchith: Holy Cow! You beat me in using it!

yanchith (commented):
I like what I see overall and agree with the overall direction. Added a few high-level suggestions.

src/operation_manager.rs
ondrowan: This is actually unreachable at the moment so will implement it as such.

src/operation_manager.rs
ondrowan: This is still in Loop Subdivision PR, so will have to wait for that.

yanchith:
Had a quick look at the UI, no code yet. This is a non-exhaustive list of things that I found confusing somehow. I don't think all or even most will need to be solved within this PR, but they should be eventually:

- It should be possible to add more than one operation without running the pipeline
- The buttons for adding an operation should be below the list of operations (so that their location is physically close to the newly created operation). There's more to this, but let's discuss online. 
- We are missing search for operations to add them or a more ergonomic way to select them
- If an operation has an optional parameter, the UI should not prefill it. Not sure we *really* want the UI to know about `Value::Nil` though - maybe we should go for default parameter values instead of optional parameters.

These are missing features:

- The title of the operation should not be missing
- Default parameters (so that the parameters in the initial operation UI are valid)
- Number ranges (so that the parameters in the initial operation UI are valid)
- Error reporting from name resolution, type checking, running

These are probably bugs:

- All geometries produced from a certain operation are named "Geometry 1 from ..." and the sequence is not incremented
- If interpreter panics, sending it a message panics the main thread as well (the interpreter **should not** panic though).

Also, we will need to work on the styles, padding, and automatic resizing.

src/lib.rs
```diff
@@ -468,6 +259,42 @@ pub fn init_and_run(options: Options) -> ! {
 
                 ui_frame.draw_renderer_settings_window(&mut renderer_draw_geometry_mode);
 
+                operation_ui.prepare_ops();
+
+                let (run_button_clicked, remove_button_clicked) =
+                    ui_frame.draw_operations_window(&ops_reprs, &mut operation_ui);
+
+                if run_button_clicked {
+                    let geometries_to_invalidate = operation_ui.submit_program();
```
yanchith, line 264: After reading the implementation of `submit_program()` I think what this should return is the list of IDs of values/geometries that should be purged. See main review comment.

src/lib.rs
```diff
@@ -468,6 +259,42 @@ pub fn init_and_run(options: Options) -> ! {
...
+                            if let Some(index) = index {
+                                scene_renderer_geometry_ids.remove(index);
+                                renderer.remove_scene_geometry(geometry_id);
+
+                                scene_geometries
+                                    .pop()
+                                    .expect("Failed to pop geometry from scene");
+                            }
+                        }
+                    }
```
yanchith: Isn't it a bit weird that we use `scene_geometries.pop()` and `scene_renderer_geometry_ids.remove()`? Shouldn't they both be one or the other?

src/ui.rs
```diff
@@ -138,9 +143,15 @@ impl<'a> UiFrame<'a> {
...
         let ui = &self.imgui_ui;
+        let window_width = 250.0;
+        let window_position = [
+            self.imgui_ui.io().display_size[0] - window_width - 50.0,
+            50.0,
+        ];
 
         imgui::Window::new(imgui::im_str!("Renderer Settings"))
-            .size([250.0, 200.0], imgui::Condition::Once)
+            .position(window_position, imgui::Condition::Always)
```
yanchith, line 156: FYI, this is really meh on my small display. Please keep it move-able.

src/operation_manager.rs
```diff
@@ -0,0 +1,530 @@
...
+                OpUiParam {
+                    name: "Parallels".to_string(),
+                    repr: OpParamUiRepr::IntInput,
+                    value: Value::new(ast::LitExpr::Uint(2)),
+                },
+                OpUiParam {
+                    name: "Meridians".to_string(),
+                    repr: OpParamUiRepr::IntInput,
+                    value: Value::new(ast::LitExpr::Uint(3)),
+                },
```
yanchith, line 457: This looks like it's going to be fragile (double maintenance with `interpreter_funcs::global_definitions`). I initially thought that we'll be defining much more stuff here, but if this is it, having it in `func::ParamInfo` makes more sense. For the interpreter tests I'd create a stub constructor function that pre-fills default repr, empty name, etc... `ParamInfo` can have methods to quickly extract important bits, like `ty()` and `optional()`.

The argument that we shouldn't mix UI and behavior still stands, but we need to weight it against the double maintenance we will be up against if we do it the way it currently is in this PR.

I didn't think about it very concretely yet, but I think it could be similar to something like this

```rust
fn params(&self) -> &[ParamInfo] {
        &[
            ParamInfo {
                param_ty: ParamTy::FloatSlider("Foo", 0.5), // will later also contain min and max
                optional: false, // TBD, but we might remove optional parameters in favor of default values just because it's easier on the UI
            },
            ParamInfo {
                param_ty: ParamTy::UintInput("Bar", 1, 1, 10),
                optional: true,
            },
            ParamInfo {
                param_ty: ParamTy::UintSlider("Bar", 1, 1, 10),
                optional: true,
            },
        ]
}
```

src/operation_manager.rs
```diff
@@ -0,0 +1,530 @@
...
+    IntSlider,
+    FloatInput,
+    Float3Input,
+    #[allow(unused)]
+    FloatSlider,
+    #[allow(unused)]
+    Checkbox,
+    #[allow(unused)]
+    Radio,
+    GeometryDropdown(Vec<(u64, String)>),
```
yanchith: This is mixing definitions and data. Shouldn't the "repr" as you call it only contain the rules on how to draw the input, and the data can come from somewhere else? This would also clean up the `ref mut` you have to use to set this later.

src/operation_manager.rs
```diff
@@ -0,0 +1,530 @@
...
+}
+
+/// Value aware of its current and previous state.
+///
+/// This is useful for determining what changed and if operation containing
+/// this value should be marked as dirty for purposes of interpreter program
+/// invalidation.
+#[derive(Debug, Clone)]
+pub struct Value {
```
yanchith, line 39: To not overload terms too much, maybe call this `Literal` or `ParameterValue`?

src/operation_manager.rs
```diff
@@ -0,0 +1,530 @@
...
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct GeometryMetadata {
+    name: String,
+    geometry: Geometry,
+    var_ident: u64,
+    geometry_id: Option<GpuGeometryId>,
```
yanchith: This doesn't seem too clean. Does this have to be here for setting later? Can't there be a cleaner way to subscribe for notifications about geometry removal so that we can also remove it from the renderer? There could be a `HashMap` sitting outside of the manager doing the conversion, we'll just need to generate an id here in the manager from a sequence.

src/operation_manager.rs
```diff
@@ -0,0 +1,530 @@
...
+fn build_operation_var_ident(operation: &Op, var_ident_id: u64) -> ast::Stmt {
+    let var_ident = ast::VarIdent(var_ident_id);
+    let mut args = vec![];
+
+    for param in &operation.params {
+        let expr = match param.repr {
+            OpParamUiRepr::GeometryDropdown(_) => match param.value.get() {
+                ast::LitExpr::Uint(uint) => {
+                    ast::Expr::Var(ast::VarExpr::new(VarIdent(u64::from(*uint))))
+                }
```
yanchith: Ugh! :eyes:

See my main review comment. I don't see the reason why vars these are faked as literals of uint.

src/operation_manager.rs
```diff
@@ -0,0 +1,530 @@
...
+    pub fn selected_ops_iter_mut(&mut self) -> std::slice::IterMut<SelectedOp> {
+        self.selected_ops.iter_mut()
+    }
+
+    pub fn runnable(&self) -> bool {
+        !self.selected_ops.is_empty()
+    }
+}
+
+fn build_operation_var_ident(operation: &Op, var_ident_id: u64) -> ast::Stmt {
```
yanchith, line 409: This actually builds a stmt, not a var ident... I was wondering why there was so much logic here :joy_cat:

src/operation_manager.rs
```diff
@@ -0,0 +1,530 @@
...
+        // Having an operation doesn't mean it also produced any geometries.
+        // Those are cleaned up only if this operation finished properly.
+        if removed_op.status == OpStatus::Finished {
+            if let Some(geometries_len) = self.geometry_stack.pop() {
+                let first_index = self.geometry_metadata.len() - geometries_len;
+                let geometry_ids = self
+                    .geometry_metadata
+                    .drain(first_index..)
+                    .filter_map(|geometry_metadata| geometry_metadata.geometry_id)
+                    .collect();
```
yanchith: If you just generate ids for the values returned by the interpreter in the manager, you can use those to notify the outside that a value/geometry is now dead (and should be removed from the renderer, e.g.). No need to store the renderer ids in the manager.

src/operation_manager.rs
```diff
@@ -0,0 +1,530 @@
...
+    /// case nothing changed, interpreter doesn't run.
+    pub fn submit_program(&mut self) -> usize {
+        let mut invalidate_statements = false;
+        let mut results_to_invalidate = 0;
+        let mut should_interpret = false;
+
+        // Find out which operations should be invalidated.
+
+        for (var_ident_id, selected_op) in self.selected_ops.iter_mut().enumerate() {
+            // First dirty operation marks all subsequent ones as invalidated.
```
yanchith, line 219: Note that this is correct from the perspective of being outside the interpreter. Internally, the interpreter will try to re-use values, if possible. To allow that you need to make sure the `ast::Call` you provide is equal to the one provided the last time. That is, unless the parameter really did not change.

src/operation_manager.rs
```diff
@@ -0,0 +1,530 @@
...
+
+    /// Polls for respone from interpreter server and runs the handler function
+    /// for each unused geometry in case a result is returned.
+    ///
+    /// Since unused geometries are the only ones to be shown in scene, the
+    /// handler function runs only for them since the geometry ID needs to be
+    /// assigned.
+    pub fn poll_interpreter_response<F>(&mut self, mut geometry_handler: F)
+    where
+        F: FnMut(&mut GeometryMetadata) -> GpuGeometryId,
```
yanchith: As noted in the main review comment and also elsewhere, it would seem better if instead of using this to inject renderer ids, we would generate our own value ids here and publish them from this function. The caller code could then store the mapping between values and renderer geometries.

src/operation_manager.rs
```diff
@@ -0,0 +1,530 @@
...
+
+                    // We care only about results with var idents larger than
+                    // the geometry stack length. This means that only new
+                    // geometries are going to be processed.
+                    let last_used_var_ident = self.geometry_stack.len();
+                    let mut new_geometry_metadata = HashMap::new();
+                    let mut var_idents = HashSet::new();
+
+                    for (var_ident, value) in &value_set.used_values {
+                        if var_ident.0 + 1 > last_used_var_ident as u64 {
```
yanchith: The `+1` seems weird here. Is there an off-by-one error somewhere?

src/operation_manager.rs
```diff
@@ -0,0 +1,530 @@
...
+                            let geometry_metadata = GeometryMetadata {
+                                name: format!("Geometry #{} from {}", 1, op_name),
```
yanchith: Aaaah, so this is why every geometry is "1" :eyes:

src/operation_manager.rs
```diff
@@ -0,0 +1,530 @@
...
+
+                    // Geometry metadata are pushed onto geometry stacks in order.
+                    // Geometry handler runs for each unused geometry so it can
+                    // be added to scene.
+                    for var_ident in var_idents {
+                        let mut geometry_metadata = new_geometry_metadata
+                            .remove(&var_ident)
+                            .expect("Failed to remove new geometry");
+
+                        if !geometry_metadata.used {
```
yanchith, line 361: Well, there's 2 options here:

- We will render just the unused geometry, in which case we don't need to keep track of used geometry at all,
- We will render used geometry differently than unused, in which case we will also want to add it to the renderer.

src/operation_manager.rs
```diff
@@ -0,0 +1,530 @@
...
+        }
+    }
+
+    /// Populates geometry dropdowns with geometries produced in previous
+    /// operations.
+    pub fn prepare_ops(&mut self) {
+        for (i, selected_op) in self.selected_ops.iter_mut().enumerate() {
+            if i == 0 {
+                continue;
+            }
```
yanchith: Why special case 0? If there's no vars, the dropdown will be empty. You later sum the `num_available_geos` anyway.

yanchith (commented):
This time around I had a closer look at the general architecture. I am ignoring nits and technicalities for now.

I think there's potential here, but am worried about the coupling of the operations manager with the renderer. Currently renderer geometry ids are passed in just so we know when to remove the geometry from the renderer. This is a weird half-coupled integration. I suggest one of 2 things:

1) Generate ids for values returned from the interpreter in the manager and give those to the outside, along with a way to subscribe to notifications when those values should be purged. We can then have a mapping from the id produced by the manager to the id produced by the renderer to know what to remove. I prefer this!

2) Integrate the renderer into the manager so that the manager just does what is needed. I don't like this as much, because we would need to republish most of the renderer's methods from the manager.

There's also some concerns inline. One notable is about using `LitExpr::Uint` to fake a geometry reference. It would be better if the code knew about vars and stored the var/dropdown data elsewhere.

PR #65, opened 2019-11-04
Optimize code and performance of separate_isolated_meshes

janper:
The function was split into two, which was unnecessary. While refactoring, I realized there are some redundant parts:
```
let mut index_stack = vec![start_face_index];
index_stack.push(start_face_index);
```
and
```
let mut connected_face_indices = HashSet::new();
connected_face_indices.shrink_to_fit();
```

There were also some mistakes in the logic, when some items could get into the into the stack more than once (see the changes regarding `available_face_indices`).

_Cry for help:_ The line `159 available_face_indices.remove(&current_face_index);` is there only to remove the very first index of the main loop. I'd like to remove this line and add `156 available_face_indices.remove(&start_face_index);` instead, but I'm getting some nasty `immutable borrow` error. Please?!

### DIFF
--- a/src/mesh_tools.rs
+++ b/src/mesh_tools.rs
@@ -150,47 +150,34 @@ pub fn separate_isolated_meshes(geometry: &Geometry) -> Vec<Geometry> {
     let mut available_face_indices: HashSet<u32> = face_to_face.keys().copied().collect();
     let mut patches: Vec<Geometry> = Vec::new();
 
-    while let Some(first_face_index) = available_face_indices.iter().next() {
-        let connected_indices = crawl_faces(*first_face_index, &face_to_face);
+    while let Some(start_face_index) = available_face_indices.iter().next() {
+        let mut index_stack = vec![*start_face_index];
+        let mut connected_face_indices = HashSet::new();
+
+        while let Some(current_face_index) = index_stack.pop() {
+            if connected_face_indices.insert(current_face_index) {
+                available_face_indices.remove(&current_face_index);
+                for neighbor_index in &face_to_face[&current_face_index] {
+                    if available_face_indices.contains(neighbor_index) {
+                        index_stack.push(*neighbor_index);
+                        available_face_indices.remove(neighbor_index);
+                    }
+                }
+            }
+        }
 
         patches.push(
             Geometry::from_faces_with_vertices_and_normals_remove_orphans(
-                connected_indices
+                connected_face_indices
                     .iter()
                     .map(|face_index| geometry.faces()[cast_usize(*face_index)]),
                 geometry.vertices().to_vec(),
                 geometry.normals().to_vec(),
             ),
         );
-
-        for c in &connected_indices {
-            available_face_indices.remove(c);
-        }
     }
 
     patches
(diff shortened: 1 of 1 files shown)


src/mesh_tools.rs
```diff
@@ -150,49 +150,36 @@ pub fn separate_isolated_meshes(geometry: &Geometry) -> Vec<Geometry> {
     let mut available_face_indices: HashSet<u32> = face_to_face.keys().copied().collect();
     let mut patches: Vec<Geometry> = Vec::new();
 
-    while let Some(first_face_index) = available_face_indices.iter().next() {
-        let connected_indices = crawl_faces(*first_face_index, &face_to_face);
+    while let Some(start_face_index) = available_face_indices.iter().next() {
+        let mut index_stack = vec![*start_face_index];
+        let mut connected_face_indices = HashSet::new();
```
ondrowan: Unless I misunderstand something, I'm pretty positive we can assume this is gonna have the same length as the original geometry. We'll also need `connected_face_indices.shrink_to_fit();` at the end as in previous iteration of the code.

src/mesh_tools.rs
```diff
@@ -150,49 +150,36 @@ pub fn separate_isolated_meshes(geometry: &Geometry) -> Vec<Geometry> {
...
+
+        while let Some(current_face_index) = index_stack.pop() {
+            if connected_face_indices.insert(current_face_index) {
+                available_face_indices.remove(&current_face_index);
```
ondrowan: As for your cry for help, you can use `available_face_indices.iter().cloned().next()` to please the borrow checker. This shouldn't even add any significant overhead in case there's just couple of different meshes.

ondrowan (changes requested):

src/mesh_tools.rs
janper: great, that works! thanks

ondrowan (approved):

src/mesh_tools.rs
```diff
@@ -150,49 +150,36 @@ pub fn separate_isolated_meshes(geometry: &Geometry) -> Vec<Geometry> {
     let mut available_face_indices: HashSet<u32> = face_to_face.keys().copied().collect();
     let mut patches: Vec<Geometry> = Vec::new();
 
-    while let Some(first_face_index) = available_face_indices.iter().next() {
-        let connected_indices = crawl_faces(*first_face_index, &face_to_face);
+    while let Some(start_face_index) = available_face_indices.iter().cloned().next() {
```
yanchith, line 153: `.copied()` would be better

src/mesh_tools.rs
```diff
@@ -150,49 +150,36 @@ pub fn separate_isolated_meshes(geometry: &Geometry) -> Vec<Geometry> {
...
+        available_face_indices.remove(&start_face_index);
+        let mut index_stack = vec![start_face_index];
+        let mut connected_face_indices = HashSet::new();
```
yanchith, line 156: Let's not recreate these for each iteration, rather have just one instance of each and clean it

PR #66, opened 2019-11-08
OBSOLETE Pull point to mesh

janper:
**PULL POINT TO MESH**
Pulls arbitrary point to the closest point of a mesh geometry.
Cast a ray from the point perpendicular to each mesh face and if there is an intersection, measure the distance. Also measure a distance to the closest point on each mesh geometry edge. Pick the closest point of them all as the pulled point.

**PULL POINT TO LINE**
Pulls arbitrary point to an arbitrary line.
Returns both, a clamped point on the line adn an unclamped points on an endless ray.
https://stackoverflow.com/questions/3120357/get-closest-point-to-a-line

**RAY INTERSECTS TRIANGLE**
The Möller–Trumbore ray-triangle intersection algorithm is a fast method for calculating the intersection of a ray and a triangle in three dimensions without needing precomputation of the plane equation of the plane containing the triangle.
https://en.wikipedia.org/wiki/Möller–Trumbore_intersection_algorithm
https://www.scratchapixel.com/lessons/3d-basic-rendering/ray-tracing-rendering-a-triangle/moller-trumbore-ray-triangle-intersection

**CALCULATE TRIANGLE FACE NORMAL**
Calculate a normalized vector perpendicular to the face pointing outwards following the right-hand rule.


src/face_tools.rs
```diff
@@ -0,0 +1,95 @@
+use nalgebra::base::Vector3;
+
+use crate::convert::cast_usize;
+use crate::geometry::{Geometry, TriangleFace};
+
+#[allow(dead_code)]
+pub fn calculate_triangle_face_normal(
```
yanchith: This function already exists, but is not `pub` yet (and does not normalize the vector). `geometry.rs`, line 1168 in master. Let's use that (and add the normalization to it).

Also, @ondrowan will probably disagree, but I don't think this function should have these tests. We need to choose our battles in what to test and how to do it. The correct way to write tests for this function would be property testing (which is pretty low on our backlog). Meanwhile I would prefer no tests to the tests provided here, since the function is pretty small.

src/vertex_tools.rs
```diff
@@ -0,0 +1,398 @@
...
+    };
+
+    PointOnLine {
+        clamped: closest_point_on_line_clamped,
+        unclamped: closest_point_on_line_unclamped,
+    }
+}
+
+#[allow(dead_code)]
+pub struct PulledPoint {
```
yanchith, line 101: Let's name this `PointOnMesh` to be consistent? Also, let's at least derive `Debug, Clone, Copy, PartialEq` for it.

src/vertex_tools.rs
```diff
@@ -0,0 +1,398 @@
...
+#[allow(dead_code)]
+pub fn ray_intersects_triangle(
+    ray_origin: &Point3<f32>,
+    ray_vector: &Vector3<f32>,
+    triangle_vertices: (&Point3<f32>, &Point3<f32>, &Point3<f32>),
+) -> Option<Point3<f32>> {
+    let edge_1_vector = triangle_vertices.1 - triangle_vertices.0;
+    let edge_2_vector = triangle_vertices.2 - triangle_vertices.0;
+    let perpendicular_vector = ray_vector.cross(&edge_2_vector);
+    let determinant = edge_1_vector.dot(&perpendicular_vector);
```
yanchith: I know this is a bit stupid, but can we please comment this a bit more and explain the relation of `determinant` to the ray being parallel to the plane defined by the triangle. Also let's explain why when the angle between `perpendicular_vector` and `edge1` is perpendicular, it means that the ray is parallel to the plane.

src/vertex_tools.rs
```diff
@@ -0,0 +1,398 @@
...
+use crate::face_tools;
+use crate::geometry::{Face, Geometry, UnorientedEdge};
+
+/// The Möller–Trumbore ray-triangle intersection algorithm is a fast method for
+/// calculating the intersection of a ray and a triangle in three dimensions
+/// without needing precomputation of the plane equation of the plane containing
+/// the triangle.
+///
+/// https://en.wikipedia.org/wiki/Möller–Trumbore_intersection_algorithm
+/// https://www.scratchapixel.com/lessons/3d-basic-rendering/ray-tracing-rendering-a-triangle/moller-trumbore-ray-triangle-intersection
```
yanchith: This scratchpixel article is actually not very good at explaining what's happening here. Let's remove the reference and instead let's link to the original paper, which at least contains some schematic pictures: http://webserver2.tecgraf.puc-rio.br/~mgattass/cg/trbRR/Fast%20MinimumStorage%20RayTriangle%20Intersection.pdf

src/vertex_tools.rs
```diff
@@ -0,0 +1,398 @@
...
+
+/// Pulls arbitrary point to an arbitrary line.
+///
+/// Returns both, a clamped point on the line adn an unclamped points on an
+/// endless ray.
+///
+/// https://stackoverflow.com/questions/3120357/get-closest-point-to-a-line
+pub fn pull_point_to_line(
+    point: &Point3<f32>,
+    line_points: (&Point3<f32>, &Point3<f32>),
```
yanchith: Wouldn't this be a bit more readable without the tuple and with two params (`line_start`, `line_end`) instead?

src/vertex_tools.rs
```diff
@@ -0,0 +1,398 @@
...
+
+#[allow(dead_code)]
+pub struct PointOnLine {
+    clamped: Point3<f32>,
+    unclamped: Point3<f32>,
+}
+
+/// Pulls arbitrary point to an arbitrary line.
+///
+/// Returns both, a clamped point on the line adn an unclamped points on an
```
yanchith: "adn" -> "and"

src/vertex_tools.rs
```diff
@@ -0,0 +1,398 @@
...
+    // Ray-triangle intersection
+    if t_parameter > f32::EPSILON && t_parameter < 1.0 / f32::EPSILON {
+        Some(ray_origin + ray_vector * t_parameter)
+    } else {
+        None
+    }
+}
+
+#[allow(dead_code)]
+pub struct PointOnLine {
```
yanchith, line 61: Also, let's at least derive `Debug, Clone, Copy, PartialEq` for this.

src/vertex_tools.rs
```diff
@@ -0,0 +1,398 @@
...
+            (
+                &vertices[cast_usize(t_f.vertices.0)],
+                &vertices[cast_usize(t_f.vertices.1)],
+                &vertices[cast_usize(t_f.vertices.2)],
+            ),
+            face_tools::calculate_triangle_face_normal(t_f, geometry),
+        )
+    });
+    let points_pulled_to_faces =
+        all_mesh_faces_with_normals.filter_map(|(unwrapped_face, face_normal)| {
```
yanchith: `unwrapped_face` -> `face_vertices`?

src/vertex_tools.rs
```diff
@@ -0,0 +1,398 @@
...
+            ray_intersects_triangle(point, &(-1.0 * face_normal), unwrapped_face)
+        });
+    let unwrapped_edges = unoriented_edges.iter().map(|u_e| {
```
yanchith: `let face_edges = ...`?

src/vertex_tools.rs
```diff
@@ -0,0 +1,398 @@
...
+            let current_dist_sq = nalgebra::distance_squared(point, &current_pt);
+            if current_dist_sq < cls_dist_sq {
+                (current_dist_sq, Some(current_pt))
+            } else {
+                (cls_dist_sq, cls_pt_opt)
+            }
+        });
+
+    PulledPoint {
+        closest_point: closest_point_option.expect("Invalid point on mesh"),
```
yanchith: Let's specify in this expect message that `Geometry` can't be empty (contain no faces). There is a trello task to assert that in `Geometry` constructors, but we don't have to implement that in this PR.

src/vertex_tools.rs
```diff
@@ -0,0 +1,398 @@
...
+            point_on_line_calculated.clamped
+        );
+        assert_eq!(
+            point_on_line_unclamped_correct,
+            point_on_line_calculated.unclamped
+        );
+    }
+
+    #[test]
+    fn test_vertex_tools_pull_point_to_line_quarter() {
```
yanchith: Isn't this test a bit redundant? It does not cover an edge case, and there already is a test for the average success case above.

src/vertex_tools.rs
```diff
@@ -0,0 +1,398 @@
...
+        let point_on_mesh_correct = Point3::new(1.0, 1.0, 1.0);
+        let pulled_point_on_mesh_calculated =
+            pull_point_to_mesh(&test_point, &cube_geometry, &unoriented_edges);
+
+        assert_eq!(
+            point_on_mesh_correct,
+            pulled_point_on_mesh_calculated.closest_point
+        );
+        assert_eq!(
+            ((3.0 * 0.25 * 0.25) as f32).sqrt(),
```
yanchith: Let's use nalgebra's distance function to compute this since we know both points. It will convey meaning better. Also for the test below.

yanchith (changes requested):
Very nice and very dense PR! Every inline request is basically just a technicality and should be easily addressable.

src/face_tools.rs
ondrowan: These tests are useful, already written and aren't interfering with anything. I don't get how no tests would be more benefitial.

janper: I've removed the redundant function, however I kept the already existing tests - mainly because I wanted to make sure the replacement function works exactly the same like the removed one.

src/vertex_tools.rs
janper: I disagree with the naming part of the request. This can be any pulled point, also there is a convention for a  *point on mesh*, which contains absolute coordinates and geometry's local coordinates (in our case they would be the face index and barycentric coords).

src/vertex_tools.rs
janper: Sure, I just want to say I don't like this sort of parameter unwrapping. Semantically, the three parameters are not hierarchically equal. I have the same bad feeling about `TriangleFace::new_separate`.

src/vertex_tools.rs
janper: resolved differently - the variable is not needed now

src/vertex_tools.rs
janper: I've described all the parts that are straight-forward understandable to me. The rest is some calculation of barycentric coordinates in a tetrahedron and to understand the math I'd have to spend too much time drawing the cases so that i can describe it. 
Also, do you think such descriptions are crucial for our project?

src/geometry.rs
```diff
@@ -1612,4 +1616,82 @@ mod tests {
...
+    fn test_face_tools_compute_triangle_normal_returns_z_vector_for_horizontal_triangle() {
+        let faces = vec![(0, 1, 2)];
+        let vertices = vec![
+            Point3::new(0.0, 1.0, 0.0),
+            Point3::new(-0.866025, -0.5, 0.0),
+            Point3::new(0.866025, -0.5, 0.0),
+        ];
+        let geometry = Geometry::from_triangle_faces_with_vertices_and_computed_normals(
+            faces,
+            vertices.clone(),
```
ondrowan, line 1630: This does not need to be cloned since it's not used further. There are more cases of this in subsequent tests.

ondrowan (changes requested):

PR #67, opened 2019-11-09
Add ability to make snapshot tests

yanchith:
Snapshot tests allow to detect changes in output of complex data, in
our case geometry. When an output of a tested function changes (or the
functionality is new and there is no snapshot to compare against yet),
the data has to be manually reviewed and approved. For us, that will
most likely mean comparing with other 3d software, or at least
visually checking in the HS editor.

Snapshot testing is provided by the `insta` library. Snapshot review
can also be done manually, but using the cargo plugin `cargo-insta`
simplifies snapshot review - `cargo insta review` interactively lists
changed snapshots and asks the reviewer to accept or reject them.

This removes a quite verbose laplacian smoothing test and replaces it
with snapshots. I also noticed one spot, where topology is passed by
value and unnecessarily cloned all over the place, so I fixed that.


janper:
I'm not sure if I get this feature right. So you calculate an output from some function once and mark it correct and then write a test that compares the current results to the correct ones?

yanchith:
Yes, but it happens automatically. The test tries to compare current data to an existing snapshot. If the snapshot does not exist, or is different from current data, a new snapshot is written to disk and the test fails. You can then use `cargo insta review` to replace the old snapshot with the new - after you manually verify the new data is correct.

ondrowan (changes requested):
This seems to work! However, there's no mention of snapshot testing or  `cargo-insta` in readme. Please fix that first, so we can ship this.

ondrowan (approved):

janper:
I don't feel qualified to approve this. 
Nevertheless, I'm looking forward to this feature!

janper:
One question though: how do you make sure the tested function generates a valid output the first time when the snapshot is being created? Do you need to test it visually or make some other test first?

yanchith:
Yes. Manual verification by whatever means. Good point though - the PR adding new tests should describe how the data wss verified. In this case, the data surplanted an existing test.

PR #70, opened 2019-11-14
Assert geometry is always nonempty (has faces) and fix tests

yanchith:


src/importer.rs
```diff
@@ -370,6 +370,14 @@ mod tests {
         }
     }
 
+    fn triangle() -> tobj::Model {
+        create_tobj_model(
+            vec![0, 1, 2],
```
janper, line 379: Is this parameter called `indices` due to a convention in the importer? If not, let's call it somewhat more informative, like `face_vertex_indices` or so.

janper (dismissed):
Yup, this is a no-brainer. Approved!
Had one question, deal with it your way ;)

src/importer.rs
yanchith: yeah, that's how it's named in the obj library we are using, but I will rename it on our side

PR #71, opened 2019-11-20
Pull vertex

janper:
~!!!!!!! IT SEEMS THE DIFF WITH MASTER CAUSED SOME OF THE NEW STUFF COMING FROM MASTER TO GET REJECTED AND REPLACED BY THE ORIGINAL STUFF. THEREFORE THIS PR IS USELESS. COULD YOU PLEASE SUGGEST A CORRECT WAY OF PREPARING A SUITABLE COMMIT, RESPECTING ALL THE CHANGES MADE INTO MASTER SINCE THE CURRENT BRANCH WAS CREATED ?????? thanks~

**I have resolved the erroneous merge manually. Ready for review.**

The pull functions are inspired by their equivalents in Grasshopper - they find the closest point on the respective geometry to current point. This is a non-trivial logic, because the behavior is different for various target geometries. 

**Pull to point** would return the target point. It is not implemented as a separate function but is a part of the following algorithms.
**Pull to line** returns a point on the line, which is a perpendicular projection of the current point onto the line or one of the line ends in case the perpendicular projection doesn't fall into the domain of the line.
**Pull to plane** returns a perpendicular projection of the current point onto the plane.
**Pull to face** returns a perpendicular projection of the current point onto a triangular face or a perpendicular projection of the current point onto a face edge (in case the perpendicular projection onto the triangle plane doesn't fall into the are of the triangle) or one of the triangle corners (in case the perpendicular projection of the current point doesn't fall into a domain of any of the triangle face edges).
**Pull to mesh** returns the closest of the points pulled to each mesh face.

The pull functions return the pulled (closest) point and a distance between it and the original point.

The functionality is vast and brings along many supplemental functions as well as data types. The most important is a Plane, which is a custom orthogonal coordinate system with arbitrary origin and rotation (normal) but original scale and no spatial distortion.

As for the future optimizations, there might be more efficient algorithms for finding perpendicular projections of point onto planes or lines than generic intersections or arbitrary rays. For now the current algorithms seem to work well and there already are incorporated some performance optimizations.


src/geometry.rs
```diff
@@ -1,5 +1,6 @@
 use std::cmp;
 use std::collections::HashSet;
+use std::f32;
```
ondrowan: This import seems to be redundant.

src/geometry.rs
```diff
@@ -785,7 +786,87 @@ fn remove_orphan_vertices_and_normals(
...
+            x_vector_normalized != -1.0 * y_vector_hint_normalized,
+            "The X and Y vectors defining a plane can't form a 180 degree angle."
+        );
+
+        // Make sure the Y vector is perpendicular to the leading X vector
+        let plane_normal = x_vector_normalized.cross(&y_vector_hint_normalized);
+        let y_vector_normalized = plane_normal.cross(&x_vector_normalized).normalize();
+
+        Plane {
+            origin: *origin,
```
ondrowan, line 884: Would it make more sense to take `origin` by value?

src/geometry.rs
```diff
@@ -785,7 +786,87 @@ fn remove_orphan_vertices_and_normals(
...
+/// The plane is endless, has an origin, orientation defined by mutually
+/// perpendicular X and Y direction vectors. Such plane doesn't allow for
+/// distortions or scaling.
+///
+/// The X vector is leading and its direction is kept in the plane unchanged.
+/// The Y vector is only a hint and is being recalculated to be perpendicular to
+/// the leading X vector.
+///
+/// # Panics
+/// Panics if the X and Y vectors are parallel (identical or reverted).
```
ondrowan: Double nit here:

* It'd be probably better if you moved comments for struct fields to fields themselves.
* The structure doesn't panic, `Plane::new` panics, so that comment should be probably moved there.

src/geometry.rs
```diff
@@ -1671,4 +1745,93 @@ mod tests {
...
+
+    #[test]
+    fn test_geometry_plane_new_calculate_perpendicular_y() {
+        let test_plane = Plane::new(
+            &Point3::new(0.0, 0.0, 0.0),
+            &Vector3::new(1.0, 0.0, 0.0),
+            &Vector3::new(1.0, 1.0, 0.0),
+        );
+
+        assert_eq!(test_plane.y_vector, Vector3::new(0.0, 1.0, 0.0));
```
ondrowan, line 1885: I don't know what agreement you have with @yanchith regarding testing, but wouldn't it make sense to test all the scenarios for construction of `Plane`?

src/interpreter/func.rs
```diff
@@ -3,7 +3,7 @@ use bitflags::bitflags;
 use super::{FuncError, Ty, Value};
 
 bitflags! {
-    /// Information about the function behaviour.
```
ondrowan, line 14: Are we 'murican now? 🍔

src/plane_tools.rs
```diff
@@ -0,0 +1,168 @@
+use std::f32;
```
ondrowan: This import seems to be redundant.

src/plane_tools.rs
```diff
@@ -0,0 +1,168 @@
...
+        assert_eq!(plane_calculated.normal(), normal_correct);
+    }
+
+    /// The test compares calculated values to the result of a similar function
+    /// in Grasshopper. The internal logic of the Grasshopper function is
+    /// unknown, therefore the resulting plane is flipped or rotated around its
+    /// normal differently. The origin of a plane fitted in Grasshopper is also
+    /// located randomly.
+    #[test]
+    fn test_plane_tools_plane_fit_random_seed_12() {
```
ondrowan, line 136: Is 12 a typo or some special seed number?

src/plane_tools.rs
```diff
@@ -0,0 +1,168 @@
...
+            Point3::new(9.717592, 4.613409, 1.807314),
+            Point3::new(0.23115, 5.332513, 0.465347),
+            Point3::new(6.13154, 3.183633, 0.808695),
+            Point3::new(5.983811, 0.062063, 1.120022),
+            Point3::new(2.671959, 3.858621, 0.598954),
+            Point3::new(4.155326, 9.551214, 0.09266),
+            Point3::new(6.821242, 9.279189, 1.344678),
+            Point3::new(8.870291, 2.748636, 0.067714),
+            Point3::new(2.329386, 7.446438, 1.592039),
+            Point3::new(0.037465, 2.681375, 0.790804),
```
ondrowan, line 118: I don't know much about insta, but wouldn't it be possible to move this to snapshot?

src/vertex_tools.rs
```diff
@@ -0,0 +1,1043 @@
...
+
+        let ray_origin = Point3::new(0.25, 0.25, 0.25);
+        let ray_vector = Vector3::new(1.0, 0.0, 0.0);
+
+        let point_on_triangle_calculated =
+            ray_intersects_triangle(&ray_origin, &ray_vector, face_points);
+
+        // There is no intersection of the triangle and ray
+        assert_eq!(None, point_on_triangle_calculated);
+    }
```
ondrowan, line 800: Aren't the previous 2 tests identical in their meaning? Or is the reason of miss significant?

src/vertex_tools.rs
```diff
@@ -0,0 +1,1043 @@
...
+        assert_eq!(0.75, pulled_point_on_mesh_calculated.distance);
+    }
+
+    #[test]
+    fn test_vertex_tools_pull_point_to_mesh_cube_point_inside_top_front_right() {
+        let cube_geometry = geometry::cube_sharp_geometry([0.0, 0.0, 0.0], 1.0);
+        let test_point = Point3::new(0.25, 0.25, 0.25);
+        let unoriented_edges: Vec<_> = cube_geometry.unoriented_edges_iter().collect();
+
+        // any fo the following points on mesh would be correct
```
ondrowan: fo -> of

ondrowan (dismissed):
I noticed there are couple of expects / asserts that panic the app if some prerequisite isn't met. I cannot say if it's ok or `Result` would be preferable since those are in `dead_code` functions / methods.

src/geometry.rs
yanchith: I am torn here as well, but my latest way of thinking is that references are a bit more clear. `rustc` usually does the right thing anyway with optimization, so I wouldn't worry about perf (there was an article on this some time ago on reddit). Wouldn't hurt if we were consistent in the entire codebase about passing these nalgebra types, which we probably aren't.

src/geometry.rs
yanchith: We have no new agreement beyond what you already know. Skimming through the changes in this PR, this would be a good candidate for property testing, which we de-prioritized. The next best thing is to test all edge-cases and one average case (I think that's the same as what you suggest)

src/interpreter/mod.rs
```diff
@@ -393,7 +393,7 @@ impl Interpreter {
     }
 
     /// Interprets the whole currently set program and returns the
-    /// used/unused values after the last statment.
+    /// used/unused values after the last statement.
```
yanchith, line 399: Thank you for fixing typos, but this inflates the PR scope. Would be better to not change code you don't need to change, even if there are typos 😉

yanchith:
I'll hopefully get to review this on Monday, stay tuned.

src/geometry.rs
janper: nalgebra itself is not consistent in this. Generally for analytic purposes it takes a reference... and actually I’m not sure if it ever expects a value. So maybe it’s not inconsistent after all.

Anyway, what do you suggest here?

src/geometry.rs
janper: There are too many edge cases here. I decided to test some and hope for the rest.

src/interpreter/func.rs
janper: Yup

src/plane_tools.rs
janper: In grasshopper number 2 is the default seed. 12 was just another random seed i chose

src/plane_tools.rs
janper: Doesn’t insta compare the results?

src/vertex_tools.rs
janper: Yes, it’s a different case. When the ray misses the triangle, it still hits the triangle plane. If there was a mistake in the logic, this might report as a hit.

src/interpreter/mod.rs
janper: Does this apply to the missing periods too?

src/vertex_tools.rs
janper: Yes, it’s a different case. When the ray misses the triangle, it still hits the triangle plane. If there was a mistake in the logic, this might report as a hit.

src/geometry.rs
yanchith: That's fine, but let's test those that you explicitly assert in the plane constructor - it means you have already thought about those 😉

src/geometry.rs
yanchith: Always be semantic in generic contexts: take a reference when you don't need to take a value. When you know the concrete type and the type is `Copy`, decide based on type's size - pass it by value only if it is smaller or equally large as our target platform's pointer type. Clippy has a lint for this.

In this concrete case that means take a reference.

src/plane_tools.rs
ondrowan: That's why I'm asking if this is maybe somehow possible. @yanchith?

src/geometry.rs
janper:
```
pub struct Plane {
    pub origin: &Point3<f32>,
    pub x_vector: &Vector3<f32>,
    pub y_vector: &Vector3<f32>,
}
```

`expected lifetime parameter`

In this case, a copy seems to make the most sense to me.

src/interpreter/mod.rs
yanchith: You are testing me... 🐱

src/geometry.rs
yanchith: I was talking about parameters only. I didn't mean you should be creating a struct containing borrowed values.

To explain better: as you might have found out, structs containing references a bit of a pain to work with, meaning you only want to create them when it's the semantically correct thing to do. e.g. when you want to extend the lifetime for which something is borrowed and lock-down the original borrowed data while the view exists. A good example of this in our codebase is `RenderPass` in `renderer/mod.rs` or `UiFrame` in `ui.rs`.

So just to be clear: in this case take the parameter by reference and trigger the copy internally.

tests/fixtures/valid_2.obj
```diff
@@ -1,7 +1,5 @@
 o Triangle
```
yanchith: Please discard changes to this file

benches/shrink_wrap.rs
```diff
@@ -6,7 +6,7 @@ use criterion::{black_box, criterion_group, criterion_main, BenchmarkId, Criteri
 
 fn bench_shrink_wrap(c: &mut Criterion) {
     let mut group = c.benchmark_group("shrink_wrap");
-    let geometry = geometry::uv_sphere([0.0, 0.0, 0.0], 1.0, 3, 3);
+    let geometry = geometry::uv_sphere_geometry([0.0, 0.0, 0.0], 1.0, 3, 3);
```
yanchith: Please don't rename the primitives. We have a task to refactor the primitive generators to their own modules. Rather, if the names conflict, move the new functionality to a new module.

src/geometry.rs
```diff
@@ -785,7 +785,85 @@ fn remove_orphan_vertices_and_normals(
...
+/// The plane is endless, has an origin, orientation defined by mutually
+/// perpendicular X and Y direction vectors. Such plane doesn't allow for
+/// distortions or scaling.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct Plane {
+    pub origin: Point3<f32>,
+    /// The X vector is leading and its direction is kept in the plane unchanged.
+    pub x_vector: Vector3<f32>,
+    /// The Y vector is only a hint and is being recalculated to be perpendicular to
+    /// the leading X vector.
```
yanchith: Both these comments should be in the constructor documentation, not here, because here the logic you describe already happened and the Y vector is recomputed.

src/geometry.rs
```diff
@@ -785,7 +785,85 @@ fn remove_orphan_vertices_and_normals(
...
+    pub y_vector: Vector3<f32>,
+}
+
+impl Plane {
+    /// # Panics
```
yanchith, line 864: Nit: missing title, should not start with "# Panics"

src/geometry.rs
```diff
@@ -785,7 +785,85 @@ fn remove_orphan_vertices_and_normals(
...
+            y_vector: y_vector_normalized,
+        }
+    }
+
+    /// The rotation of the X and Y vector around the normal is random.
+    ///
+    /// # Panics
+    /// Panics if the normal vector is a zero vector.
+    pub fn from_origin_and_normal(origin: &Point3<f32>, normal: &Vector3<f32>) -> Plane {
+        assert_ne!(*normal, Vector3::zeros());
```
yanchith: Add a message to this assert.

src/geometry.rs
```diff
@@ -785,7 +785,85 @@ fn remove_orphan_vertices_and_normals(
...
+        point_on_x: &Point3<f32>,
+        point_on_y: &Point3<f32>,
+    ) -> Plane {
+        let x_vector = point_on_x - origin;
+        let y_vector = point_on_y - origin;
+        Plane::new(origin, &x_vector, &y_vector)
+    }
+
+    pub fn normal(&self) -> Vector3<f32> {
+        self.x_vector.cross(&self.y_vector).normalize()
```
yanchith: X and Y are already normalized and perpendicular, This normalization is redundant.

src/geometry.rs
```diff
@@ -785,7 +785,85 @@ fn remove_orphan_vertices_and_normals(
...
 }
 
-pub fn plane(position: [f32; 3], scale: f32) -> Geometry {
+/// Plane defining euclidean orthogonal unit space origin and orientation.
+///
+/// The plane is endless, has an origin, orientation defined by mutually
+/// perpendicular X and Y direction vectors. Such plane doesn't allow for
+/// distortions or scaling.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct Plane {
```
yanchith, line 851: In some comments I tell you to skip some normalizations, because the x a y vectors are already normalized. That doesn't hold in one case: if you mutate the fields directly. To that end, please make the fields not `pub`, and instead create accessor methods.

src/geometry.rs
```diff
@@ -785,7 +785,85 @@ fn remove_orphan_vertices_and_normals(
...
+        let y_vector_hint_normalized = y_vector_hint.normalize();
+
+        assert!(
+            x_vector_normalized != y_vector_hint_normalized,
+            "The X and Y vectors defining a plane must be different"
+        );
+        assert!(
+            x_vector_normalized != -1.0 * y_vector_hint_normalized,
+            "The X and Y vectors defining a plane can't form a 180 degree angle"
+        );
```
yanchith: As we discussed offline, these two normalizations are redundant. If these vectors lie on the same line, the cross product will be a zero vector. We can check for that and report just a single assert error for both of the cases reported here.

src/geometry.rs
```diff
@@ -1671,4 +1742,113 @@ mod tests {
 
         assert!(geometry.has_no_orphan_vertices());
     }
+
+    #[test]
+    fn test_geometry_compute_triangle_normal_returns_z_vector_for_horizontal_triangle() {
```
yanchith, line 1839: No need to create the actual geometry in this test, the points are enough. Also in the tests below :point_down:

src/plane_tools.rs
```diff
@@ -0,0 +1,166 @@
...
+        yz += r.y * r.z;
+        zz += r.z * r.z;
+    }
+
+    xx /= n as f32;
+    xy /= n as f32;
+    xz /= n as f32;
+    yy /= n as f32;
+    yz /= n as f32;
+    zz /= n as f32;
```
yanchith: Let's cast n to f32 just once and assign it to a variable.

src/plane_tools.rs
```diff
@@ -0,0 +1,166 @@
...
+use crate::geometry::Plane;
+
+/// Fit a plane to a collection of points. Fast, and accurate to within a few
+/// degrees. Returns None if the points do not span a plane.
+///
+/// The resulting normal is flipped randomly as well as the rotation of the X
+/// and Y vector around the normal is random. The result is, nevertheless,
+/// correct.
+///
+/// https://www.ilikebigbits.com/2017_09_25_plane_from_points_2.html
```
yanchith, line 13: Lol this is actually rust code 😄 We are too mainstream now. Also - this algorithm actually seems to do a kind of principal component analysis. Very cool!

src/plane_tools.rs
yanchith: As @janper says, snapshots are for comparing results. I don't know where to move this test fixture tbh.

src/vertex_tools.rs
```diff
@@ -0,0 +1,1043 @@
...
+    if f32::abs(ortho.z) < 1.0 {
+        None
+    } else {
+        Some(Point3::new(
+            1.0 - (ortho.x + ortho.y) / ortho.z,
+            ortho.y / ortho.z,
+            ortho.x / ortho.z,
+        ))
+    }
+}
```
yanchith: Nice code 😉 , but let's call it `compute_barycentric_coords` or similar in this context. Don't forget to rename tests.

src/vertex_tools.rs
```diff
@@ -0,0 +1,1043 @@
...
+
+/// Test if an arbitrary point lies on a plane.
+///
+/// https://stackoverflow.com/questions/17227149/using-dot-product-to-determine-if-point-lies-on-a-plane
+fn is_point_on_plane(
+    point: &Point3<f32>,
+    point_on_plane: &Point3<f32>,
+    plane_normal: &Vector3<f32>,
```
yanchith: Shouldn't this take `plane: &Plane` as a parameter instead of these two?

src/vertex_tools.rs
```diff
@@ -0,0 +1,1043 @@
...
+        triangle_vertices.0,
+        triangle_vertices.1,
+        triangle_vertices.2,
+    );
+
+    if is_point_on_plane(point, &plane.origin, &plane.normal()) {
+        let a = Point2::origin();
+        let b = Point2::from((triangle_vertices.1 - triangle_vertices.0).xy());
+        let c = Point2::from((triangle_vertices.2 - triangle_vertices.0).xy());
+        let p = Point2::from((point - triangle_vertices.0).xy());
```
yanchith: As we talked offline, projecting onto the plane is not enough. Instead we need to orient both the triangle and the point on the XY plane, so we can safely disregard the Z coords. Doing this also removes the necessity of manually checking `is_point_on_plane`, because that will be answered by asking `point.z == plane.origin.z` after orientation.

src/vertex_tools.rs
```diff
@@ -0,0 +1,1043 @@
...
+        let barycentric_point = to_barycentric(a, b, c, p).expect("The triangle is degenerate");
```
yanchith: This `expect` can very well fire under current conditions, AFAIK. See my above comment. Should be ok after we orient the point and plane, instead of projecting.

src/vertex_tools.rs
```diff
@@ -0,0 +1,1043 @@
...
+    ray_origin: &Point3<f32>,
+    ray_vector: &Vector3<f32>,
+    triangle_vertices: (&Point3<f32>, &Point3<f32>, &Point3<f32>),
+) -> Option<PulledPointWithDistance> {
+    if is_point_in_triangle(ray_origin, triangle_vertices) {
+        return Some(PulledPointWithDistance {
+            point: *ray_origin,
+            distance: 0.0,
+        });
+    }
```
yanchith: As we talked offline, it would be good to check this outside of this function and just add a warning to this function's docs.

src/vertex_tools.rs
```diff
@@ -0,0 +1,1043 @@
...
+    };
+
+    PointOnLine {
+        clamped: closest_point_on_line_clamped,
+        unclamped: closest_point_on_line_unclamped,
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct PulledPointWithDistance {
```
yanchith: Define this struct above any functions that return it.

src/vertex_tools.rs
```diff
@@ -0,0 +1,1043 @@
...
+    ray_vector: &Vector3<f32>,
+    plane: &Plane,
+) -> Option<PulledPointWithDistance> {
+    let plane_normal = plane.normal();
+    if is_point_on_plane(ray_origin, &plane.origin, &plane_normal) {
+        return Some(PulledPointWithDistance {
+            point: *ray_origin,
+            distance: 0.0,
+        });
+    }
```
yanchith: Analogous to ray_intersects_triangle... shouldn't this be checked outside and the condition be just documented here for symmertry?

src/vertex_tools.rs
```diff
@@ -0,0 +1,1043 @@
...
+    fn test_vertex_tools_ray_intersects_triangle_for_horizontal_triangle_and_point_inside_returns_identical(
+    ) {
+        let face_points = (
+            &Point3::new(0.0, 1.0, 0.0),
+            &Point3::new(-0.866025, -0.5, 0.0),
+            &Point3::new(0.866025, -0.5, 0.0),
+        );
+
+        let ray_origin = Point3::new(0.0, 0.0, 0.0);
+        let ray_vector = Vector3::new(0.0, 0.0, 0.0);
```
yanchith: Shouldn't this be a nonzero vector so that we isolate the failing condition?

yanchith (changes requested):
Good job! I only have small nits and technicalities, and one Icorrectness error in `is_point_in_triangle`. See inline.

Also, can you confirm that the snapshots changed because we started normalizing the normals?

benches/shrink_wrap.rs
janper: it now follows the convention I've started before that a generator returning a geometry is suffixed with `_geometry`. the reason to do it now was a new data type `Plane`, which was in conflict with the original `plane` generator.
leaving as is, the refactor will follow

src/geometry.rs
janper: ...when you hover over the field in an IDE, it shows you this description. i find it useful here, but ok.

src/vertex_tools.rs
janper: the best piece of code ever!

src/vertex_tools.rs
janper: it should

benches/shrink_wrap.rs
yanchith: Yes, I know your reasoning, just disagree with it :joy: 

What I was suggesting is that because you wanted to move the `Plane` and related logic to it's own file (IIRC), there would no longer be a naming conflict and no reason to rename them. I don't think we should rename the primitives, because the subsequent cleanup PR will move them to their own file and we will be calling them throughout the project this way: `mesh::primitive::sphere(...)`, so the geometry suffix won't make sense anymore.

yanchith: If you really want to, we can do that work in the cleanup PR and not now.

janper: oh, a disagreement! good.
i don't think this is only about the naming conflict. as i mentioned above, there are some generators that return `Geometry` and some that only return geometry components and you need to create the `Geometry` yourself. therefore this remains relevant even after the refactor.
nevertheless, shouldn't the refactor be a separate task that happens some time in the future and if we decide the suffix is not needed anymore, we remove it there? although I thin we should keep it even there

yanchith: Ok, I don't have very strong feelings about this, so let's not spend more time on this now.

janper: no vidíš ;)

janper:
Not sure why did the snapshots change again, but the results look visually correct. Tested in the real use.

benches/shrink_wrap.rs
```diff
@@ -0,0 +1,45 @@
+use hurban_selector::geometry;
```
yanchith: This file was deleted in master and you added it back. Keep it deleted.

src/operations/shrink_wrap.rs
```diff
@@ -0,0 +1,27 @@
+use std::slice;
```
yanchith: This file was deleted in master. Keep it deleted.

src/plane_tools.rs
```diff
@@ -0,0 +1,168 @@
+use nalgebra::base::Vector3;
```
yanchith, line 1: TBH I don't see the added benefit of having all the `*_tools` modules. I would put this in a module called `plane`, together with the new `Plane` structure. Don't change anything now. I will do so in the followup cleanup.

src/vertex_tools.rs
```diff
@@ -0,0 +1,1272 @@
...
+    v0_normalized == v1_normalized && v0_normalized == v2_normalized
+}
+
+/// Checks if a point lies in a triangle.
+///
+/// #Panics
+/// Panics if triangle is colinear
+///
+/// https://math.stackexchange.com/questions/4322/check-whether-a-point-is-within-a-3d-triangle
+// TODO: Orient the triangle and point to XY plane and then test
```
yanchith: Forgot to remove this todo? I believe this is done :+1:

src/vertex_tools.rs
```diff
@@ -0,0 +1,1272 @@
...
+        triangle_vertices.2,
+    );
+
+    // If the point is not on the triangle plane, it's also not on the triangle.
+    if !is_point_on_plane(point, &plane) {
+        return false;
+    }
+
+    // In case the triangle isn't horizontal
+    if let Some(rotation_to_horizontal) =
```
yanchith: Let's merge this `if` branch with early return with the code that follows it. They are almost exact copy-paste.

The overall structure can be something like:

```rust
// pseudocode

let (rotated_triangle, rotated_point) = if check_horizontal(triangle) {
    (triangle, point)
} else {
    if let Some(rotation_to_horizontal) = rotation_between(...) {
        (apply_rotation(triangle, rotation_to_horizontal), apply_rotation(point, rotation_to_horizontal))
    } else {
        // This probably won't happen, but we shouldn't crash in case it does :)
        // https://docs.rs/nalgebra/0.19.0/src/nalgebra/geometry/rotation_specialization.rs.html#663
        (triangle, point)
    }
}

// ... the rest of the code that's the same for both branches, will operate on `rotated_triangle` and `rotated_point`
```

src/vertex_tools.rs
```diff
@@ -32,25 +31,44 @@ pub fn to_barycentric(
     }
 }
 
+/// Checks if all three points of a triangle lay on the same line.
+pub fn are_triangle_vertices_colinear(
```
yanchith: Nit: `are_vertices_colinear`? Also update comment if you change the name

src/geometry.rs
```diff
@@ -859,7 +919,19 @@ impl Plane {
     }
 
     pub fn normal(&self) -> Vector3<f32> {
```
yanchith, line 921: In case of `normal()` you return a `Vector3` and in case of `x_vector`, `y_vector` and `origin` you return a `&Vector3`. Let's be consistent. It is not super clear what the correct answer is, let's pick the owned version.

yanchith (changes requested):
Almost there! We just need to clean up some incorrect merges and refactor a small piece of code and this can land.

yanchith:
> Not sure why did the snapshots change again, but the results look visually correct. Tested in the real use.

I checked and it is just the normals, because we inserted a normalization in function that computes them, and that function is used every time we recompute sharp normals for a mesh.

src/plane_tools.rs
janper: exactly

src/vertex_tools.rs
janper: just an explanation: the input argument is a tuple `(pt, pt, pt)`, which is used in triangle-related functions

src/geometry.rs
janper: because the referenced ones are just extractions of some existing value, the normal is calculated (created anew). nevertheless, i'm doing as you say

src/vertex_tools.rs
janper: I need to apologize. You were right. This covers all the cases but the last else statement shouldn't yield the original coords. It should never occur, as you correctly expect, but if it does, it means we're not able to calculate any valid result. In fact we should return None, but that's not helpful for the caller. Therefore i decided to early return false instead. The main reason for the early return is, however, that the barycentric coords coming from this branch won't be valid and therefore the calculations won't yield any valid results either.

yanchith: Ok, let's keep it as is.

yanchith (approved):
:+1:

PR #72, opened 2019-11-21
Fix erroneous advancing of PC when setting one prog stmt

yanchith:
This is one small part extracted from the work on UI, so that that PR can be smaller when it lands.

### DIFF
--- a/src/interpreter/mod.rs
+++ b/src/interpreter/mod.rs
@@ -1499,7 +1499,7 @@ mod tests {
         assert_eq!(interpreter.pc, 0);
 
         interpreter.set_prog_stmt_at(
-            0,
+            1,
             ast::Stmt::VarDecl(ast::VarDeclStmt::new(
                 VarIdent(0),
                 ast::CallExpr::new(func_id, vec![]),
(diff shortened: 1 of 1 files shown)


src/interpreter/mod.rs
```diff
@@ -323,8 +323,11 @@ impl Interpreter {
     }
 
     pub fn set_prog_stmt_at(&mut self, index: usize, stmt: ast::Stmt) {
-        // Mark the stmt and everything following it for re-execution
-        self.pc = index;
+        // Mark the stmt and everything following it for re-execution,
+        // if we already executed that far
+        if self.pc > index {
+            self.pc = index;
+        }
```
janper, line 330: not that i'd think this is wrong, but is this better than `self.pc = min(self.pc, index);`?

janper (commented):
I only have one question because i don;t understand the rest ;)

src/interpreter/mod.rs
yanchith: Just a matter of style

ondrowan (changes requested):
The new test actually passes even with the old behaviour.

yanchith:
@ondrowan oops, that was a copy paste error. Now it doesn't.

ondrowan (approved):

PR #73, opened 2019-11-25
Implement import obj func

yanchith:
The function is implemented as a impure interpreter func. It can
return two possible errors (importer error and empty geometry
error). Internally, it uses the `Importer` and its cache.

This commit also removes the importer worker, as that is now fully
superceeded by the interpreter server.

This is another piece of functionality split off of the gui work.
The UI itself will be the next and final in this series.


src/interpreter_funcs.rs
```diff
@@ -394,6 +404,62 @@ impl Func for FuncImplCreatePlane {
...
+        Ty::Geometry
+    }
+
+    fn call(&mut self, values: &[Value]) -> Result<Value, FuncError> {
+        let path = values[0].unwrap_string();
+
+        let result = self.importer.import_obj(path);
+        match result {
+            Ok(models) => {
+                // FIXME: @Correctness Return all geometries once we support all of them
```
janper: Rather `Join all geometries into one mesh`. We need to implement such function first (it's pretty high in the backlog), so this is good enough for now.

janper (dismissed):
I only understood the importer part. Seems good enough for now. Please, be more specific with the `FIXME`.

src/interpreter_funcs.rs
```diff
@@ -447,7 +447,9 @@ impl<C: ObjCache> Func for FuncImplImportObjMesh<C> {
         let result = self.importer.import_obj(path);
         match result {
             Ok(models) => {
-                // FIXME: @Correctness Return all geometries once we support all of them
+                // FIXME: @Correctness Join all meshes into one once
+                // we have join implemented for more than just 2
+                // meshes
```
janper, line 452: *thumbs-up*

ondrowan (approved):

PR #74, opened 2019-11-26
Add editor session and its UI

yanchith:
The editor session keeps track of the current pipeline program,
notifies the interpreter of changes, and can also trigger running the
pipeline program in the interpreter. It can poll for responses from
the interpreter, and update its state accordingly.

The UI communicates directly with the session. UI actions modify the
program in the session, and the session transmits the changes to the
interpreter. Running the pipeline can be triggered in the UI.

The UI for editing (push/pop/set) operations is disabled whenever the
interpreter is busy, so that we can later add error reporting to
either the operation console, or as a tooltip over an operation. The
interpreter is considered busy when there still exists at least one
request in flight with the interpreter server.

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -41,28 +41,21 @@ pub struct Session {
     /// (exclusive), e.g. 0th stmt can not see any vars, 1st stmt can
     /// see vars produced by the 0th stmt, etc.
     var_visibility: Vec<VarIdent>,
-    var_count: u64,
 
     function_table: BTreeMap<FuncIdent, Box<dyn Func>>,
 }
 
 impl Session {
     pub fn new() -> Self {
-        let prog = Prog::new(Vec::new());
-        let var_visibility = Vec::new();
-
-        assert_eq!(prog.stmts().len(), var_visibility.len());
-
         Self {
             interpreter_server: InterpreterServer::new(),
             interpreter_requests_in_flight: HashSet::new(),
 
-            prog,
+            prog: Prog::new(Vec::new()),
 
             unused_values: HashMap::new(),
 
-            var_visibility,
-            var_count: 0,
+            var_visibility: Vec::new(),
 
             // FIXME: @Correctness this is a hack, there should be
             // just one instance of the table. Fortunately we don't
@@ -87,8 +80,6 @@ impl Session {
         self.prog.push_stmt(stmt.clone());
         self.submit(InterpreterRequest::PushProgStmt(stmt));
 
-        self.var_count += 1;
-
         self.recompute_var_visibility();
     }
 
@@ -103,8 +94,6 @@ impl Session {
 
         self.prog.pop_stmt();
         self.submit(InterpreterRequest::PopProgStmt);
-
-        self.var_count -= 1;
 
         self.recompute_var_visibility();
     }
@@ -138,7 +127,7 @@ impl Session {
     /// Returns the next free variable identifier for the current
     /// program definition.
     pub fn next_free_var_ident(&self) -> VarIdent {
-        VarIdent(self.var_count)
+        VarIdent(self.prog.stmts().len() as u64)
     }
 
     /// Returns human readable variable name for a variable identifier
--- a/src/session.rs
+++ b/src/session.rs
@@ -1,0 +1,375 @@
+use std::collections::hash_map::{Entry, HashMap};
+use std::collections::{BTreeMap, HashSet};
+use std::sync::Arc;
+
+use crate::geometry::Geometry;
+use crate::interpreter::ast::{FuncIdent, Prog, Stmt, VarIdent};
+use crate::interpreter::Func;
+use crate::interpreter_funcs;
+use crate::interpreter_server::{
+    InterpreterRequest, InterpreterResponse, InterpreterServer, PollResponseError, RequestId,
+};
+
+/// A notification from the session to the surrounding environment
+/// about what values have been added since the last poll, and what
+/// values have been removed are no longer required.
+pub enum PollInterpreterResponseNotification {
+    Add(VarIdent, Arc<Geometry>),
+    Remove(VarIdent),
+}
+
+/// An editing session.
+///
+/// Contains the current definition of the pipeline program and can
+/// execute it via the interpreter running in a separete thread.
+///
+/// All program mutations are non-blocking (the session does not wait
+/// for the interpreter to acknowledge them), and polling the session
+/// for interpreter results is also non-blocking - if there is no
+/// response, nothing happens.
+pub struct Session {
+    interpreter_server: InterpreterServer,
+    interpreter_requests_in_flight: HashSet<RequestId>,
+
+    prog: Prog,
+
+    unused_values: HashMap<VarIdent, Arc<Geometry>>,
+
+    /// Auxiliary side-array for prog. Determines the vars visible
+    /// from the statement. The value is read by producing a slice
+    /// from the begining of the array to the current stmt's index
+    /// (exclusive), e.g. 0th stmt can not see any vars, 1st stmt can
+    /// see vars produced by the 0th stmt, etc.
+    var_visibility: Vec<VarIdent>,
+    var_count: u64,
+
+    function_table: BTreeMap<FuncIdent, Box<dyn Func>>,
+}
+
+impl Session {
+    pub fn new() -> Self {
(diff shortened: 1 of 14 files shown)


src/session.rs
```diff
@@ -0,0 +1,375 @@
...
+/// values have been removed are no longer required.
+pub enum PollInterpreterResponseNotification {
+    Add(VarIdent, Arc<Geometry>),
+    Remove(VarIdent),
+}
+
+/// An editing session.
+///
+/// Contains the current definition of the pipeline program and can
+/// execute it via the interpreter running in a separete thread.
```
ondrowan: Nit: `separete` -> `separate`

src/session.rs
```diff
@@ -0,0 +1,375 @@
...
+
+    function_table: BTreeMap<FuncIdent, Box<dyn Func>>,
+}
+
+impl Session {
+    pub fn new() -> Self {
+        let prog = Prog::new(Vec::new());
+        let var_visibility = Vec::new();
+
+        assert_eq!(prog.stmts().len(), var_visibility.len());
```
ondrowan: It's not possible to fail this, is it?

src/session.rs
```diff
@@ -0,0 +1,375 @@
...
+
+    unused_values: HashMap<VarIdent, Arc<Geometry>>,
+
+    /// Auxiliary side-array for prog. Determines the vars visible
+    /// from the statement. The value is read by producing a slice
+    /// from the begining of the array to the current stmt's index
+    /// (exclusive), e.g. 0th stmt can not see any vars, 1st stmt can
+    /// see vars produced by the 0th stmt, etc.
+    var_visibility: Vec<VarIdent>,
+    var_count: u64,
```
ondrowan: Wouldn't it be possible to always assume next available var ident ID to be number of prog statements + 1? Just to get rid of tiny bit of complexity?

src/session.rs
```diff
@@ -0,0 +1,375 @@
...
+            var_count: 0,
+
+            // FIXME: @Correctness this is a hack, there should be
+            // just one instance of the table. Fortunately we don't
+            // ever use it mutable from here. Find a way to split out
+            // the mutable part of the table (the funcs' state) and
+            // make sure that one is only instantiated once, while the
+            // descriptors can be instantiated multiple times and
+            // shared.
+            function_table: interpreter_funcs::create_function_table(),
```
ondrowan, line 76: I don't understand purpose of this or the comment. What mutable state is there for functions?

src/session.rs
```diff
@@ -0,0 +1,375 @@
...
+    }
+
+    /// Starts the interpreter on the current program.
+    pub fn interpret(&mut self) {
+        // This is because the current session could want to report
+        // errors and we would like to show them somewhere
+        assert!(
+            !self.interpreter_busy(),
+            "Can't submit a request while the interpreter is already interpreting",
+        );
```
ondrowan, line 194: Why is this an assert and not just a plain condition?

src/session.rs
```diff
@@ -0,0 +1,375 @@
...
+                                    //   the outside of the change
+                                    // - Detect which values should be
+                                    //   added and `Add` them
+
+                                    // FIXME: Currently, we only call the
+                                    // add/remove handlers with when a
+                                    // value enters or leaves the
+                                    // unused_values set. This can be
+                                    // easily copy-paste extended to
+                                    // handle used_values the same way.
```
ondrowan: These kinds of comments aren't readable. Don't enforce wrapping at 80th char and just use those 60-80 chars for comment width.

src/ui.rs
```diff
@@ -155,10 +194,506 @@ impl<'a> UiFrame<'a> {
...
+            )
+            .build(ui, || {
+                ui.columns(2, imgui::im_str!("Controls columns"), false);
+
+                let running_tokens = if running_enabled {
+                    None
+                } else {
+                    Some(push_disabled_style(ui))
+                };
+                if ui.button(imgui::im_str!("Run pipeline"), [-f32::MIN_POSITIVE, 25.0])
```
ondrowan, line 569: How is `f32::MIN_POSITIVE` different from `0.0` in this context?

src/ui.rs
```diff
@@ -155,10 +194,506 @@ impl<'a> UiFrame<'a> {
...
+
+        imgui::Window::new(imgui::im_str!("Pipeline"))
+            .movable(false)
+            .resizable(false)
+            .size([PIPELINE_WINDOW_WIDTH, pipeline_window_height], imgui::Condition::Always)
+            .position([MARGIN, MARGIN], imgui::Condition::Always)
+            .build(ui, || {
+                for (stmt_index, stmt) in session.stmts().iter().enumerate() {
+                    match stmt {
+                        ast::Stmt::VarDecl(var_decl) => {
```
ondrowan, line 231: Will we ever care about anything but `VarDecl` in this case?

ondrowan (commented):
This is just a partial review since it's a huge PR.

I don't know if you wanna do it in scope of this branch, but we talked about enabling just operations that produce geometries before you can add anything that consumes them.

There's some `<nil>` geometry in dropdowns. Is it intended as "default" choice? If so, it's not correctly sorted and shown as the first one.

yanchith:
Thanks for the review! I'll also address the inline comments throughout the day.

> I don't know if you wanna do it in scope of this branch, but we talked about enabling just operations that produce geometries before you can add anything that consumes them.

I am not sure I want to do it at all. I tried to attack this problem from the other direction, by using the `<Nil>` variable references. Ultimately, we should experiment which of these methods provides better UX and pick. Or defer the decision to @janper  😉 I agree that the solution I tried here may be too nerdy 😂 

> There's some `<nil>` geometry in dropdowns. Is it intended as "default" choice? If so, it's not correctly sorted and shown as the first one.

It is only supposed to be default if there is no other geometry present. Otherwise the default is to pick first of the real, non-nil variables. There is a task (https://trello.com/c/r8vKCRBk/213-the-last-created-geometry-should-be-the-default-pick-into-the-newly-added-operation) to change the default to last, but we had to stop somewhere, so I left it for later.

src/session.rs
yanchith: Good point! I think so! At least for now. For later it might not be true, but that might be miles of, so this is premature.

src/session.rs
yanchith: Only by programmer error. But this is ridiculous, I agree. Must be a remnant of some other code I had.

src/session.rs
yanchith: Currently just the importer cache. Later, who knows... maybe rng state, etc... Let's hope there is not a lot of that. This should be thought through better for sure. Ideally the function descriptions wouldn't contain the state, but implementing it this way required less changes and the PR was already big, so I left if for the fixme.

src/session.rs
yanchith: Good question. I don't think it *really* needs to be, but I was worried it could lead to some unpredictable behavior if the UI didn't uphold this, so it was mostly a debugging tool to make sure I didn't make a mistake in the UI code. I am not against removing, if that's what you suggest.

src/session.rs
yanchith: I'd enforce wrapping at 100th char, if anything.

src/ui.rs
yanchith: Very much so! `MIN_POSITIVE` means autosize to full width of the container and `0.0` means autosize to text content... Yeah, I know 😂

src/ui.rs
yanchith: We most likely will soon. There will be operations that don't produce variables, only do side-effects (e.g. export obj). The possible description of those might be a statement that is not a `VarDecl`. By the way adding those things will invalidate a lot of assumptions made in this PR, for example the current simplistic determining of variable visibility.

src/session.rs
yanchith: Except the +1. The next var ident should be exactly the current number of statements.

src/session.rs
ondrowan: Then at least improve the comment. I spent quite a lot of time trying to find out what's happening here.

src/session.rs
ondrowan: It can stay as a reminder to fix this later on if you don't wanna do anything about it right now.

ondrowan (dismissed):

PR #75, opened 2019-11-28
Fix implementation of are_visually_similar

yanchith:
The previous implementation was too lenient, and didn't perform its check
bidirectionally, which would be ok, if the data was deduplicated, but
that may not be (and usually isn't) the case.


janper (approved):
Nice a clean. Better than the HashSet solution.

PR #77, opened 2019-12-01
Log to file only for dist builds

yanchith:
This allows us to make a release build that still can do terminal
logging for debugging purposes.

The feature `dist` can be used to distinguish whether we are building
an internal or dist build.


src/logger.rs
```diff
@@ -1,20 +1,45 @@
...
+    }
+}
 
 /// Initializes logger for current environment.
 ///
-/// In case of any error, basic logger, that does nothing, is returned. Errors
-/// can possibly happen in case of file permission errors in release build.
-pub fn init() {
+/// In case of any error, basic logger that does nothing, is returned. Errors
+/// can possibly happen in case of file fs permission errors in dist build.
```
ondrowan: Typo: "file fs"

ondrowan (dismissed):

PR #78, opened 2019-12-01
Split funcs into separate submodules

yanchith:
This PR almost doesn't change any code, only restructures the `interpreter_funcs` module into submodules. The only changed code is:

- Rename `FuncImpl*` -> `Func*`
- The `operations` module was removed and the implementations contained there (transform and shrinkwrap) were inlined into their respective funcs.


src/interpreter_funcs/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+/// Returns the global set of function definitions available to the
+/// editor.
+///
+/// Note that since funcs can have internal state such as a cache or
+/// random state, two instances of the function table are not always
+/// equivalent.
+pub fn create_function_table() -> BTreeMap<FuncIdent, Box<dyn Func>> {
+    let mut funcs: BTreeMap<FuncIdent, Box<dyn Func>> = BTreeMap::new();
+
+    // Special funcs
```
janper: Not sure if this should be called *special*. What about *manipulation* for now?

janper (dismissed):
Yes, this is cleaner.
I'm wondering if you are planning to group the functions in the UI as well.

yanchith:
> I'm wondering if you are planning to group the functions in the UI as well.

I started thinking about it. There's certainly something we can do. Let's discuss the design of it first.

src/interpreter_funcs/mod.rs
yanchith: Sure. All of it is up for discussion.

PR #79, opened 2019-12-01
Track edit and interpret requests separately

yanchith:
This allows multiple concurrent edit requests in flight and also
doesn't mark the interpreter busy when only the program is being
edited, and no interpreting is actually happening.


janper (approved):
Okay

PR #80, opened 2019-12-05
Mesh groups

yanchith:
This adds mesh groups... or `Value::GeometryArray`. In the interpreter, they are just another value. In the UI and viewport, we need to know about them specially to display them in the combo boxes and the viewport. Current user-visible behavior in viewport is identical to other geometries - there is no way to tell those 2 apart (probably will improve in at some point). In the UI combo boxes, each combo box only receives suggestions for the type it expects - one can not accidentally pass a `GeometryArray` where `Geometry` is expected and vice versa.

Two existing funcs were changed to return `GeometryArray`: `FuncImportObj`, and `FuncSeparateIsolatedMeshes`. To get a geometry from an array, use the newly added `FuncExtract`.

There seems to be functionality duplication growing between the session and the interpreter (mostly because they can only communicate asynchronously and the session doesn't like waiting for answers). It is not too terrible yet, but it would be a good target for some future cleanup effort. There is a new fixme for that.

@ondrowan If you have the capacity, this is something I'd like you to check. I know you'll probably object to me upping the rust version inside this PR. :wink: if you want me to split it out. It is required for a change that happened here though.


src/interpreter/ast.rs
```diff
@@ -133,8 +133,6 @@ pub enum Expr {
     Lit(LitExpr),
     #[allow(dead_code)]
     Var(VarExpr),
-    #[allow(dead_code)]
-    Index(IndexExpr),
```
yanchith, line 137: This was never implemented and is now superseded by `FuncExtract`

src/ui.rs
```diff
@@ -681,6 +635,90 @@ impl<'a> UiFrame<'a> {
...
+
+    // FIXME: Show used var idents differently from unused,
+    // e.g. grayed-out
+
+    // FIXME: find a way to make combo boxes read-only
+    let mut combo = imgui::ComboBox::new(input_label);
+
+    let mut combo_changed = false;
+    let preview_value = selected_var_index
+        .map(|index| visible_vars_iter.clone().nth(index).unwrap())
```
ondrowan: `expect` this instead of unwrapping

src/ui.rs
```diff
@@ -681,6 +635,90 @@ impl<'a> UiFrame<'a> {
     }
 }
 
+fn draw_var_combo_box(
```
ondrowan: Does Rust give you hell if you move this into `UiFrame`?

ondrowan (dismissed):

src/interpreter_funcs/extract.rs
```diff
@@ -0,0 +1,78 @@
...
+        let geometry_array = values[0].unwrap_geometry_array();
+        let index = values[1].unwrap_uint();
+
+        if geometry_array.is_empty() {
+            return Err(FuncError::new(FuncExtractError::Empty));
+        }
+
+        // FIXME: @Correctness The wrapping index is just a temporary
+        // crutch until we can report errors to the user.
+        let wrapping_index = index % geometry_array.len();
```
janper, line 71: I'm worried the user will get errors way too often due to strict error reporting in the final version.
~~We should at least consider treating all the cases when the min-max domain is dynamic this way (wrapping around) as a principle. We can then report a warning instead of an error.~~
More generally, we should try to autofix each case when a parameter depends on the actual pipeline calculation and report a verbose warning suggesting a meaningful manual fix (i.e. suggesting a min-max range).

src/interpreter_funcs/import_obj_mesh.rs
```diff
@@ -64,15 +64,16 @@ impl<C: ObjCache> Func for FuncImportObjMesh<C> {
...
+                if models.is_empty() {
                     Err(FuncError::new(FuncImportObjMeshError::Empty))
+                } else {
+                    let geometries: Vec<_> = models
+                        .into_iter()
+                        .map(|model| Arc::new(model.geometry))
+                        .collect();
+
+                    let value = GeometryArrayValue::new(geometries);
+                    Ok(Value::GeometryArray(Arc::new(value)))
```
janper, line 76: If you want to test it on some real-world files, we've got these: https://onedrive.live.com/?id=A0980D900E318435%211885424&cid=A0980D900E318435
They originally come from students of VSVU, collected by Maros, converted by Kristyna.

src/interpreter_funcs/separate_isolated_meshes.rs
```diff
@@ -28,19 +28,15 @@ impl Func for FuncSeparateIsolatedMeshes {
...
     fn return_ty(&self) -> Ty {
-        Ty::Geometry
+        Ty::GeometryArray
     }
 
     fn call(&mut self, args: &[Value]) -> Result<Value, FuncError> {
         let geometry = args[0].unwrap_geometry();
 
-        let values = mesh_tools::separate_isolated_meshes(&geometry);
+        let geometries = mesh_tools::separate_isolated_meshes(&geometry);
```
janper, line 37: off topic: we should consider renaming this to `disjoint_mesh` during the geometry cleanup

src/lib.rs
```diff
@@ -54,6 +55,18 @@ pub struct Options {
     pub lib_log_level: Option<logger::LogLevel>,
 }
 
+/// A unique identifier assigned to a value or subvalue for purposes
+/// of displaying in the viewport.
+///
+/// Since we support value arrays, there can be multiple geometries
+/// contained in a single value that all need to treated separately
```
janper: need to **be** treated

src/ui.rs
```diff
@@ -370,107 +370,39 @@ impl<'a> UiFrame<'a> {
...
+                                            }
+                                        }
+                                        ParamRefinement::GeometryArray => {
+                                            let changed_expr = draw_var_combo_box(
+                                                ui,
+                                                session,
+                                                stmt_index,
+                                                arg,
+                                                Ty::GeometryArray,
+                                                &input_label,
```
janper, line 395: Not sure if this is the right way and place, but what about wrapping the group name into square brackets here to distinguish the groups from meshes? Something like `&fmt!("[{}]", input_label)`

janper (dismissed):
Thank you, this is a very important feature!

src/ui.rs
yanchith: I didn't even try honestly. It should work afaik, it's just that for some reason I default to free functions instead of associated functions when the function is going to be private anyway. I can make it an associated function if you want, no strong feelings.

src/interpreter_funcs/extract.rs
yanchith: The programmer in me agrees with your second suggestion. We should always give the user as much information as possible so that they can make good decisions and fix problems. I am just not sure what non-programmer people will expect :joy:

src/interpreter_funcs/import_obj_mesh.rs
yanchith: I'll test, just not for this PR. If it doesn't work, it most likely isn't the fault of this the changes introduced here.

src/interpreter_funcs/extract.rs
janper: The programmer and the user meet half way.
In this case the autofix will be the wrapping or clamping + a warning tells the user what to do to remove it.

src/interpreter_funcs/separate_isolated_meshes.rs
yanchith: I think I've seen a new trello issue for that. Please prioritize it in the backlog accordingly, or add it to the geometry cleanup task.

src/ui.rs
yanchith: How nerdy of you. I like it! :+1:

PR #81, opened 2019-12-08
Modulageddon

yanchith:
This is just plain terrible. I think we've put off this cleanup for far too long.

I am still very dissatisfied with `pull.rs` (90% former `vertex_tools.rs`) and quite dissatisfied with `mesh/analysis.rs` (90% former `mesh_analysis.rs`), but I had to stop somewhere. We can clean those up better later.

I still need to review the diff after myself and write a summary of what changed (it's mostly what the trello issue says + doc cleanup + redistribution of functions between modules), but if you want to get a head start @janper, feel free to start reviewing. The best way to review is not to view the whole diff from the start, but rather see changes commit by commit - the commit messages describe what happened when.

UPDATE: now I consider this complete.

High level overview of the changes:

- renamed `Geometry`to `Mesh` and `geometry.rs` to `mesh/mod.rs`
- renamed `Value::Geometry` to `Value::Mesh`
- renamed `Value::GeometryArray` to `Value::MeshArray`
- removed helper types like `pub type Vertices = Vec<Point3<f32>>`
- removed totally misused helper functions like `v()` or `n()`
- renamed `separate_isolated_meshes` to `disjoint_mesh` (also renamed the `Func` and its filename)
- moved `mesh_topology_analysis.rs` to `mesh/topology.rs`
- moved `mesh_smoothing.rs` to `mesh/smoothing.rs`
- moved `mesh_tools.rs` to `mesh/tools.rs`
- moved primitive generators from `mesh/mod.rs` to `mesh/primitive.rs`
- moved `Plane` to its own file (`plane.rs`)
- merged `plane_tools,rs` with `plane.rs`.
- made `plane_fit` a method on `Plane` (`Plane::fit`)
- made `is_point_on_plane` a method on `Plane` (`Plane::contains_point`)
- repurposed `geometry.rs` to host miscellaneous geometry functions (like computing triangle normals, etc)
- moved `compute_barycentric_coords` from `vertex_tools.rs` to `geometry.rs`
- moved `are_points_colinear` from `vertex_tools.rs` to `geometry.rs`
- moved some old analysis functions (`compute_centroid` and friends) from `mesh/mod.rs` to `mesh/analysis.rs`
- merged `edge_analysis.rs` and `mesh_analysis.rs` into `mesh/analysis.rs`
- renamed `vertex_tools.rs` to `pull.rs`, because all functionality in that file is related to pulling after I moved the misc functions to `geometry.rs`
- renamed text occurrences of geometry to mesh in documentation and comments
- fixed documentation style everywhere I updated it
- removed `<_module_name_>` substring from tests everywhere I touched
- added an optimization fixme for edge sharing
- did not delete any functionality, only moved it around
- and probably a lot more :joy:


src/geometry.rs
```diff
@@ -1,1842 +1,57 @@
...
-        assert!(geometry.has_no_orphan_vertices());
-    }
-
-    #[test]
-    fn test_geometry_from_triangle_faces_with_vertices_and_normals_remove_orphans() {
-        let (faces, vertices, normals) = quad_with_normals();
-        let extra_vertex = vec![v(0.0, 0.0, 0.0, [0.0, 0.0, 0.0], 1.0)];
-        let vertices_extended = [&extra_vertex[..], &vertices[..]].concat();
-        let extra_normal = vec![n(0.0, 0.0, 0.0)];
-        le
```
janper: /// Checks if all three points lay on the same line.

src/interpreter_funcs/create_plane.rs
```diff
@@ -50,14 +50,14 @@ impl Func for FuncCreatePlane {
...
+        Ty::Mesh
     }
 
     fn call(&mut self, values: &[Value]) -> Result<Value, FuncError> {
         let position = values[0].get_float3().unwrap_or([0.0; 3]);
         let scale = values[1].get_float().unwrap_or(1.0);
 
-        let value = geometry::plane_geometry(position, scale);
-        Ok(Value::Geometry(Arc::new(value)))
+        let value = primitive::create_plane(position, scale);
```
janper: _Can't comment on line 14, so I'm commenting here instead:_ I still don't like that this is called a "plane" because we already have something else called that. Shouldn't we distinguish those two somehow? I don't care which one: this could be a "Mesh plane" or the other one can be a "Construction plane"...

src/interpreter_funcs/extract.rs
```diff
@@ -55,24 +55,24 @@ impl Func for FuncExtract {
...
+        if mesh_array.is_empty() {
             return Err(FuncError::new(FuncExtractError::Empty));
         }
 
         // FIXME: @Correctness The wrapping index is just a temporary
         // crutch until we can report errors to the user.
-        let wrapping_index = index % geometry_array.len();
-        let value = geometry_array
+        let wrapping_index = index % mesh_array.len();
+        let value = mesh_array
```
janper, line 72: _Can't comment on line 29, so commenting here:_ After I've been using this operation for some time, it seems it needs a better name in the UI. I suggest something like "Mesh from group" or "Extract from group". The internal name is OK.

src/interpreter_funcs/import_obj_mesh.rs
```diff
@@ -67,13 +67,13 @@ impl<C: ObjCache> Func for FuncImportObjMesh<C> {
...
-                    let geometries: Vec<_> = models
+                    let meshes: Vec<_> = models
                         .into_iter()
-                        .map(|model| Arc::new(model.geometry))
+                        .map(|model| Arc::new(model.mesh))
                         .collect();
 
-                    let value = GeometryArrayValue::new(geometries);
-                    Ok(Value::GeometryArray(Arc::new(value)))
+                    let value = MeshArrayValue::new(meshes);
```
janper, line 75: _Can't comment on line 40, so commenting here:_ It's a bit confusing that the import gets you a group, which you then need to extract. I suggest giving this operation a more descriptive UI name, such as "Import OBJ as group" or so.

src/interpreter_funcs/join_meshes.rs
```diff
@@ -23,26 +23,26 @@ impl Func for FuncJoinMeshes {
         &[
             ParamInfo {
                 name: "Mesh 1",
-                refinement: ParamRefinement::Geometry,
+                refinement: ParamRefinement::Mesh,
```
janper, line 26: _Food for thought and maybe a task for a far future:_ I know we don't want to promote the groups too much, but what about joining groups instead of two meshes? It makes more sense after import. i can imagine importing an OBJ with hundreds of objects and this would be the only way to work with them. Maybe for starters we can have two versions of the join operation - one for meshes, another one for groups.

src/interpreter_funcs/shrink_wrap.rs
```diff
@@ -1,11 +1,11 @@
 use std::iter;
```
janper, line 1: I'm not sure if we should keep this operation at all. It's not very useful the way it works now. We already have most of the functionality to make it useful though, so it's up to you whether we should keep it and rework or delete it now and make it from scratch. AFAIK, we're now only missing a geodesic sphere function to make this useful.

src/mesh/analysis.rs
```diff
@@ -1,13 +1,125 @@
...
+use nalgebra::{Point3, Vector3};
 
 use crate::convert::{cast_i32, cast_usize};
-use crate::edge_analysis::EdgeSharingMap;
-use crate::geometry::{Face, Geometry, OrientedEdge, UnorientedEdge};
 
-/// Finds edges with a certain valency in a mesh edge collection
+use super::{Face, Mesh, OrientedEdge, UnorientedEdge};
+
+pub fn compute_bounding_sphere<'a, I>(meshes: I) -> (Point3<f32>, f32)
```
janper, line 10: _I  know it's not related to this PR but let's keep this as a note and maybe let's make a Trello task:_ This is not a very basic function. I think we might need a more generic version of this - operating just on a single mesh or even better on a list of points. 
Also we'll need a bounding box quite soon. I'm wondering if we shouldn't turn the bounding box and bounding sphere into structs.

src/mesh/analysis.rs
```diff
@@ -1,13 +1,125 @@
...
+            if distance_squared > max_distance_squared {
+                max_distance_squared = distance_squared;
+            }
+        }
+    }
+
+    (centroid, max_distance_squared.sqrt())
+}
+
+pub fn compute_centroid<'a, I>(meshes: I) -> Point3<f32>
```
janper, line 29: _I  know it's not related to this PR but let's keep this as a note and maybe let's make a Trello task:_ This algorithm tends to put the centroid towards the areas with the highest vertex density, which is not useful for our purposes. While technically this is a centroid according to some definitions, I believe it would be more useful to compute a bounding box first and then finding its center. 
Also we'll need a more general version of this, operating on a single mesh or even better on a list of vertices.

src/mesh/analysis.rs
```diff
@@ -1,13 +1,125 @@
...
+        for vertex in &mesh.vertices {
+            let v = vertex - Point3::origin();
+            centroid += v;
+        }
+    }
+
+    centroid / (vertex_count as f32)
+}
+
+// FIXME: deprecate and remove, superseded by pull point to mesh
```
janper: _I know it's not related to this PR but it's just a matter of changing this FIXME note:_ I wouldn't remove this, rather just rename and change the arguments to `point: Point3, cloud: [Point3]`. I believe it still might be useful.

src/mesh/analysis.rs
```diff
@@ -1,13 +1,125 @@
...
+    }
+
+    Some(closest)
+}
+
+/// The edges sharing the same vertex indices.
+/// ascending_edges contains edges oriented from lower index to higher
+/// descending_edges contains edges oriented from higher index to lower
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SharedEdges {
```
janper, line 70: Not sure, perhaps you had a reason for keeping it here, but it seems to me this semantically belongs to `topology.rs`. It applies to all the other edge sharing functions.

src/mesh/analysis.rs
```diff
@@ -116,12 +227,13 @@ pub fn border_edge_loops(edge_sharing: &EdgeSharingMap) -> Vec<Vec<UnorientedEdg
...
-/// single counterpart in a reverted oriented edge. In an open orientable mesh
-/// each internal edge has its counterpart in a single reverted oriented edge
-/// and the border edges don't have any counterpart.
+/// In a proper watertight orientable mesh each oriented edge should
+/// have a single counterpart in a reverted oriented edge. In an open
+/// orientable mesh each internal edge has its counterpart in a single
+/// reverted oriented edge and the border edges don't have any
+/// counterpart.
 #[allow(dead_code)]
 pub fn is_mesh_orientable(edge_sharing: &EdgeSharingMap) -> bool {
```
janper, line 238: This again belongs here. Personally, I'd move lines 66-228 to `topology.rs`.

src/mesh/mod.rs
```diff
@@ -0,0 +1,1481 @@
+use std::cmp;
```
janper, line 1: _Just a question:_ wouldn't it be cleaner to keep the logic in `src/mesh/mesh.rs` and pub import it here?

src/mesh/smoothing.rs
```diff
@@ -3,59 +3,48 @@ use std::collections::hash_map::{Entry, HashMap};
...
-///   iterations
-///
-/// returns (smooth_geometry: Geometry, executed_iterations: u32, stable: bool)
+/// Returns `(smooth_mesh: Mesh, executed_iterations: u32, stable: bool)`.
 pub fn laplacian_smoothing(
-    geometry: &Geometry,
+    mesh: &Mesh,
     vertex_to_vertex_topology: &HashMap<u32, SmallVec<[u32; 8]>>,
-    iterations: u32,
+    max_iterations: u32,
```
janper, line 30: _Just an explanation:_ I named this `iterations` rather than `max_iterations` because in case the `stop_when_stable` flag is off, this will be the actual number of iterations.

src/pull.rs
```diff
@@ -1,42 +1,173 @@
...
+        *line_start
+    } else if line_parameter > 1.0 {
+        *line_end
     } else {
-        Some(Point3::new(
-            1.0 - (ortho.x + ortho.y) / ortho.z,
-            ortho.y / ortho.z,
-            ortho.x / ortho.z,
-        ))
+        closest_point_on_line_unclamped
```
janper, line 54: right

janper (dismissed):
Thank you, this is incredibly useful!

There are just two real question regarding this PR (see analysis vs. topology and mesh inside mod) but even there I don't have any strong feelings. I like it the way it is now and the new structure will be much easier to work with it.

As I went through the code, I left a lot of notes unrelated to this PR. Please, go through them and mark those you agree with. I'll make Trello tasks out of them. But please, don't hesitate to merge because of the notes ;)

src/geometry.rs
janper: I don't know why does it show the comment here. It belongs to line 41.

src/interpreter_funcs/create_plane.rs
yanchith: Sure! I'd keep `Plane` as is, but I can rename the primitive to `mesh_plane`

src/interpreter_funcs/extract.rs
yanchith: ~Ok, I'll create a task~ Nevermind, I implemented it here.

src/interpreter_funcs/import_obj_mesh.rs
yanchith: ~Will create a task~ Nevermind, renaming now.

src/interpreter_funcs/join_meshes.rs
yanchith: I was already thinking about this as well. I think we'll have to start using the software to know for sure, but I am not against two versions.

src/interpreter_funcs/shrink_wrap.rs
yanchith: I'll keep it for now, but I know we will rework it soon enough.

src/mesh/analysis.rs
yanchith: It takes an iterator becuase that's most generic (and also required by the way it's actually used). I wouldn't be opposed to having points version of the function.

I'll create a task.

> I'm wondering if we shouldn't turn the bounding box and bounding sphere into structs.

Sounds good. We can do so when we add the bounding box.

src/mesh/analysis.rs
yanchith: Yes, I was already thinking that we might want to have another strategy for centroid.

> Also we'll need a more general version of this, operating on a single mesh or even better on a list of vertices.

Iterator of meshes is more general than one mesh (and also required for the current use), but I am not opposed to having a points version.

I'll create a task.

src/mesh/analysis.rs
yanchith: Hmmm, yes, it is a bit similar. I can try moving it there to see if any dependencies break.

src/mesh/mod.rs
yanchith: Sure, we can do that.

src/pull.rs
yanchith: No, this is just a very confusing diff. The deleted section you see is from another function.

src/mesh/analysis.rs
yanchith: Attempted this, but the result was not very convincing - there's a lot of dependencies on the edge sharing in analysis. I'd leave it be for now and we can revisit later.

src/mesh/mod.rs
yanchith: Haha, turns out there is a clippy lint against just this :smile: Let's leave it as it then. We do the same with renderer and interpreter.

https://rust-lang.github.io/rust-clippy/master/index.html#module_inception

src/mesh/analysis.rs
janper: So what do you insert as a parameter for a single mesh? `vec![my_mesh]`?

PR #84, opened 2019-12-08
Add Bounding box and remove Bounding sphere with centroid

janper:
World-aligned bounding box is a virtual box-shaped geometry that contains the entire input geometry and defines an envelope aligned to the world (euclidean) coordinate system. it is a simplified representation of a complex mesh geometry.

### DIFF
--- a/src/mesh/analysis.rs
+++ b/src/mesh/analysis.rs
@@ -16,56 +16,41 @@ pub struct BoundingBox {
 }
 
 impl BoundingBox {
-    pub fn new(min_x: f32, min_y: f32, min_z: f32, max_x: f32, max_y: f32, max_z: f32) -> Self {
-        let (minimum_x, maximum_x) = if min_x < max_x {
-            (min_x, max_x)
-        } else {
-            (max_x, min_x)
-        };
-        let (minimum_y, maximum_y) = if min_y < max_y {
-            (min_y, max_y)
-        } else {
-            (max_y, min_y)
-        };
-        let (minimum_z, maximum_z) = if min_z < max_z {
-            (min_z, max_z)
-        } else {
-            (max_z, min_z)
-        };
+    #[allow(dead_code)]
+    pub fn new_separate(
+        min_x: f32,
+        min_y: f32,
+        min_z: f32,
+        max_x: f32,
+        max_y: f32,
+        max_z: f32,
+    ) -> Self {
         BoundingBox {
-            minimum_point: Point3::new(minimum_x, minimum_y, minimum_z),
-            maximum_point: Point3::new(maximum_x, maximum_y, maximum_z),
+            minimum_point: Point3::new(min_x, min_y, min_z),
+            maximum_point: Point3::new(max_x, max_y, max_z),
         }
     }
 
     #[allow(dead_code)]
-    pub fn from_extremes(minimum_point: Point3<f32>, maximum_point: Point3<f32>) -> Self {
-        BoundingBox::new(
-            minimum_point.x,
-            minimum_point.y,
-            minimum_point.z,
-            maximum_point.x,
-            maximum_point.y,
-            maximum_point.z,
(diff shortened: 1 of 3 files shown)


src/mesh/analysis.rs
```diff
@@ -7,40 +8,113 @@ use crate::convert::{cast_i32, cast_usize};
...
+    }
+
+    #[allow(dead_code)]
+    pub fn maximum_point(&self) -> Point3<f32> {
+        self.maximum_point
     }
 
-    centroid / (vertex_count as f32)
+    pub fn center(&self) -> Point3<f32> {
+        Point3::from((self.minimum_point.coords + self.maximum_point.coords) / 2.0)
```
yanchith: Nit: https://docs.rs/nalgebra/0.19.0/nalgebra/fn.center.html

src/mesh/analysis.rs
```diff
@@ -7,40 +8,113 @@ use crate::convert::{cast_i32, cast_usize};
...
-    let mut vertex_count = 0;
-    let mut centroid = Point3::origin();
-    for mesh in meshes {
-        vertex_count += mesh.vertices.len();
-        for vertex in &mesh.vertices {
-            let v = vertex - Point3::origin();
-            centroid += v;
+    pub fn from_points<'a, I>(points: I) -> Self
+    where
+        I: IntoIterator<Item = &'a Point3<f32>> + Clone,
```
yanchith: This `+ Clone` bound is not necessary here. It was present in the previous code because the iterator had to be iterated multiple times there.

src/mesh/analysis.rs
```diff
@@ -7,40 +8,113 @@ use crate::convert::{cast_i32, cast_usize};
...
+        };
+        let (minimum_z, maximum_z) = if min_z < max_z {
+            (min_z, max_z)
+        } else {
+            (max_z, min_z)
+        };
+        BoundingBox {
+            minimum_point: Point3::new(minimum_x, minimum_y, minimum_z),
+            maximum_point: Point3::new(maximum_x, maximum_y, maximum_z),
         }
```
yanchith: Unless you are adamant on having this, I'd remove `BoundingBox::new` and inline the logic in `BoundingBox::from_extremes` (also, I'd rename `from_extremes` to `new`). `BoundingBox::from_meshes` does not require these checks to happen.

Or do you imagine we'll be using this constructor?

src/mesh/analysis.rs
```diff
@@ -7,40 +8,113 @@ use crate::convert::{cast_i32, cast_usize};
...
+            maximum_point.y,
+            maximum_point.z,
+        )
+    }
+
+    pub fn from_meshes<'a, I>(meshes: I) -> Self
+    where
+        I: IntoIterator<Item = &'a Mesh> + Clone,
+    {
+        // TODO: don't collect here
```
yanchith: I'll debug this at home, if still relevant after removing the `Clone` bound.

yanchith (commented):
Nice! I'll debug the TODOed issue at home and then merge. Try removing the `Clone` bound though...

src/mesh/analysis.rs
```diff
@@ -7,40 +8,113 @@ use crate::convert::{cast_i32, cast_usize};
...
+            minimum_point.z,
+            maximum_point.x,
+            maximum_point.y,
+            maximum_point.z,
+        )
+    }
+
+    pub fn from_meshes<'a, I>(meshes: I) -> Self
+    where
+        I: IntoIterator<Item = &'a Mesh> + Clone,
```
yanchith: This `Clone` bound is also redundant

src/mesh/analysis.rs
janper: yup, it's not needed anymore ;)

yanchith (approved):
:+1:

PR #85, opened 2019-12-08
Join multiple meshes at once

janper:
Replace the original `join_meshes` function with `join_multiple_meshes`, which will also get an operation. This is useful after importing an OBJ with many separate meshes and is crucial for voxelization.

There is still an error, though, and I'm too inexperienced (stupid) to fix it. Could you please have a look?

`src/interpreter_funcs/join_group.rs` lines 35 to 39

```
let mesh_arc_array = args[0].unwrap_mesh_array();

let meshes = mesh_arc_array.iter();
let value = tools::join_multiple_meshes(meshes);
```

```
type mismatch resolving `<impl std::iter::Iterator as std::iter::IntoIterator>::Item == &mesh::Mesh`
expected type `std::sync::Arc<mesh::Mesh>`
   found type `&mesh::Mesh`rustc(E0271)
```

### DIFF
--- a/src/interpreter_funcs/join_group.rs
+++ b/src/interpreter_funcs/join_group.rs
@@ -34,7 +34,6 @@ impl Func for FuncJoinGroup {
     fn call(&mut self, args: &[Value]) -> Result<Value, FuncError> {
         let mesh_arc_array = args[0].unwrap_mesh_array();
 
-        // TODO: this doesn't work :(
         let meshes = mesh_arc_array.iter();
         let value = tools::join_multiple_meshes(meshes);
         Ok(Value::Mesh(Arc::new(value)))
--- a/src/interpreter_funcs/join_group.rs
+++ b/src/interpreter_funcs/join_group.rs
@@ -1,0 +1,42 @@
+use std::sync::Arc;
+
+use crate::interpreter::{
+    Func, FuncError, FuncFlags, FuncInfo, ParamInfo, ParamRefinement, Ty, Value,
+};
+use crate::mesh::tools;
+
+pub struct FuncJoinGroup;
+
+impl Func for FuncJoinGroup {
+    fn info(&self) -> &FuncInfo {
+        &FuncInfo {
+            name: "Join Group",
+            return_value_name: "Joined Meshes",
+        }
+    }
+
+    fn flags(&self) -> FuncFlags {
+        FuncFlags::PURE
+    }
+
+    fn param_info(&self) -> &[ParamInfo] {
+        &[ParamInfo {
+            name: "Group",
+            refinement: ParamRefinement::MeshArray,
+            optional: false,
+        }]
+    }
+
+    fn return_ty(&self) -> Ty {
+        Ty::Mesh
+    }
+
+    fn call(&mut self, args: &[Value]) -> Result<Value, FuncError> {
+        let mesh_arc_array = args[0].unwrap_mesh_array();
+
+        // TODO: this doesn't work :(
+        let meshes = mesh_arc_array.iter();
+        let value = tools::join_multiple_meshes(meshes);
+        Ok(Value::Mesh(Arc::new(value)))
+    }
+}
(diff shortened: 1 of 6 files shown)


ondrowan:
@janper It's saying that `mesh_arc_array.iter()` returns iterator of `Arc<Mesh>` while you expect `&Mesh`. I think changing `I` into `IntoIterator<Item = std::sync::Arc<Mesh>>,` could do the trick (if that's how you want it to be passed).

src/interpreter_funcs/join_group.rs
```diff
@@ -0,0 +1,42 @@
...
+};
+use crate::mesh::tools;
+
+pub struct FuncJoinGroup;
+
+impl Func for FuncJoinGroup {
+    fn info(&self) -> &FuncInfo {
+        &FuncInfo {
+            name: "Join Group",
+            return_value_name: "Joined Meshes",
```
yanchith: "Joined Mesh"?

src/interpreter_funcs/join_group.rs
```diff
@@ -0,0 +1,42 @@
...
+
+    fn return_ty(&self) -> Ty {
+        Ty::Mesh
+    }
+
+    fn call(&mut self, args: &[Value]) -> Result<Value, FuncError> {
+        let mesh_arc_array = args[0].unwrap_mesh_array();
+
+        // TODO: this doesn't work :(
+        let meshes = mesh_arc_array.iter();
```
yanchith, line 37: You need to `Deref` the `Arc` to a `&`. It's just one `.map()` away :wink:

https://doc.rust-lang.org/std/ops/trait.Deref.html 

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&code=use%20std%3A%3Aops%3A%3ADeref%3B%0Ause%20std%3A%3Async%3A%3AArc%3B%0A%0Astruct%20Thing%3B%0A%0Afn%20take%3C%27a%2C%20I%3E(things%3A%20I)%20where%20I%3A%20IntoIterator%3CItem%3D%26%27a%20Thing%3E%20%7B%7D%0A%0Afn%20main()%20%7B%0A%20%20%20%20let%20things%20%3D%20vec!%5BArc%3A%3Anew(Thing)%2C%20Arc%3A%3Anew(Thing)%5D%3B%0A%20%20%20%20take(things.iter().map(%7Cthing%7C%20Arc%3A%3Aderef(%26thing)))%3B%0A%7D

src/interpreter_funcs/join_meshes.rs
```diff
@@ -42,7 +42,7 @@ impl Func for FuncJoinMeshes {
         let first_mesh = args[0].unwrap_mesh();
         let second_mesh = args[1].unwrap_mesh();
 
-        let value = tools::join_meshes(first_mesh, second_mesh);
+        let value = tools::join_multiple_meshes(vec![first_mesh, second_mesh]);
```
yanchith: This is a bit low effort, isn't it? :D

You can make an iterator without allocating here for sure... Try `&[]` instead of `vec![]`

yanchith (commented):
Good job! The error can be fixed by mapping and derefing to a reference.

src/interpreter_funcs/join_group.rs
yanchith: Hmm, now that I am thinking about it, it would be good to add that derefed iterator to `MeshArrayValue` directly... Maybe rename the current iterator to `iter_refcounted` and make the new `iter` do what you want here.

src/interpreter_funcs/join_meshes.rs
```diff
@@ -39,10 +39,9 @@ impl Func for FuncJoinMeshes {
     }
 
     fn call(&mut self, args: &[Value]) -> Result<Value, FuncError> {
-        let first_mesh = args[0].unwrap_mesh();
-        let second_mesh = args[1].unwrap_mesh();
+        let meshes = args.iter().map(|a| a.unwrap_mesh());
```
yanchith, line 42: Just a note that we don't actually support varargs and this seems a bit courageous to do (will crash as soon as we add a nonmesh parameter here), but will work with the current version, so we can leave it for now.

yanchith (approved):

src/interpreter_funcs/join_meshes.rs
janper: I'm aware of it and that was also what I was thinking - once we add another arg (if we ever have a reason to do so), we'll have to rework this anyway.

PR #86, opened 2019-12-09
Create world-aligned sharp box mesh

janper:


src/mesh/primitive.rs
```diff
@@ -139,6 +139,63 @@ pub fn create_cube_sharp(position: [f32; 3], scale: f32) -> Mesh {
     Mesh::from_triangle_faces_with_vertices_and_normals(faces, vertex_positions, vertex_normals)
 }
 
+#[allow(dead_code)]
+pub fn create_box_sharp(center: Point3<f32>, scale: [f32; 3]) -> Mesh {
```
yanchith: Wouldn't it be better if we modified `create_cube_sharp` to support the use case this is intended for?

yanchith: It's only missing the non-uniform scale.

janper: okay

src/mesh/primitive.rs
```diff
@@ -85,58 +85,10 @@ pub fn create_cube_smooth(position: [f32; 3], scale: f32) -> Mesh {
...
-        // left
-        TriangleFace::new_separate(4, 7, 0, 5, 5, 5),
-        TriangleFace::new_separate(1, 0, 7, 5, 5, 5),
-    ];
-
-    Mesh::from_triangle_faces_with_vertices_and_normals(faces, vertex_positions, vertex_normals)
+    create_box_sharp(
+        Point3::new(position[0], position[1], position[2]),
+        [scale, scale, scale],
+    )
```
yanchith: Why do we need both box and cube? They are the same thing. If it is about the difference in default scale, cube doesn't care. We should have only one of these.

yanchith:
@janper  So now we have `cube_smooth` and `box_sharp`? :smile:

janper:
Yes. Mainly because I wasn't planning to touch any existing code, just needed the box. Nevertheless, the smooth cube is not used anywhere, not even in any tests. Should I remove it completely?

yanchith:
> Yes. Mainly because I wasn't planning to touch any existing code, just needed the box. Nevertheless, the smooth cube is not used anywhere, not even in any tests. Should I remove it completely?

If you think it will never be useful, than yeah, let's remove it. I was only talking about the asymmetry in names between those two. Btw, is box a more standard name than cube?

src/mesh/primitive.rs
janper: > > Yes. Mainly because I wasn't planning to touch any existing code, just needed the box. Nevertheless, the smooth cube is not used anywhere, not even in any tests. Should I remove it completely?
> 
> If you think it will never be useful, than yeah, let's remove it. 

Yes, let's remove it then. I don't feel like computing the normals for a box. I'm about to make a smooth normal calculator for an existing mesh tomorrow morning anyway. I'll also make the normal strategy, yay!

> Btw, is box a more standard name than cube?

Yes, I've never seen this being called a Cube. I think it's because a box is more general and cube is just it's subset.

janper:
> > Yes. Mainly because I wasn't planning to touch any existing code, just needed the box. Nevertheless, the smooth cube is not used anywhere, not even in any tests. Should I remove it completely?
> 
> If you think it will never be useful, than yeah, let's remove it. 

Yes, let's remove it then. I don't feel like computing the normals for a box. I'm about to make a smooth normal calculator for an existing mesh tomorrow morning anyway. I'll also make the normal strategy, yay!

> Btw, is box a more standard name than cube?

Yes, I've never seen this being called a Cube. I think it's because a box is more general and cube is just it's subset.

yanchith (approved):

PR #87, opened 2019-12-09
Fix the barycentric coords calculator

janper:
The original function (which the most beautiful part of the code, though) returned None for no obvious reason and it happened where it shouldn't. The current version works and was tested visually and against a different implementation of the same functionality (from https://observablehq.com/@infowantstobeseen/barycentric-coordinates#calc_barycentric)

### DIFF
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -12,7 +12,6 @@ pub fn compute_triangle_normal(
     Vector3::cross(&u, &v).normalize()
 }
 
-
 /// Computes barycentric coordinates of point P in triangle A, B, C.
 pub fn compute_barycentric_coords(
     a: Point2<f32>,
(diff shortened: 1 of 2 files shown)


yanchith:
@janper Please have a look at this https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=1f0f431a9b36af1c1f6efe6bdd85ace2

I tested both implementations and the new one can easily produce NaNs or infinities. On what good data did the original function fail you?

janper:
> @janper Please have a look at this https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=1f0f431a9b36af1c1f6efe6bdd85ace2
> 
> I tested both implementations and the new one can easily produce NaNs or infinities. On what good data did the original function fail you?

I was pulling a 3d grid onto a mesh sphere. One of the pulls failed due to the barycentric coords calculator returning None.

janper:
You're right. I found an error in the pull function. I'm fixing that instead.

PR #88, opened 2019-12-09
Ui filepicker

yanchith:
This adds another type of string parameter to the UI - a file path. The UI launches a filepicker when a file path string input is used.

@ondrowan This does not launch the filepicker in a separate thread on purpose. Janper assures me blocking all rendering is a perfectly ok user experience.

Items to do:

- [x] Test window resizing (to see if `calc_item_width` always does the correct thing)
- [x] Test on linux
- [x] Test on Windows
- [x] Test on macOS


src/ui.rs
```diff
@@ -748,3 +757,56 @@ fn push_disabled_style(ui: &imgui::Ui) -> (imgui::ColorStackToken, imgui::StyleS
...
     (color_token, style_token)
 }
+
+fn file_input(
+    ui: &imgui::Ui,
+    label: &imgui::ImStr,
+    file_ext_filter: Option<(&str, &str)>,
+    buffer: &mut imgui::ImString,
+) -> bool {
+    let edit_button_label = imgui::im_str!("Edit##{}");
```
janper: "Open"

janper (changes requested):
Technically it works, but the path looks like this: image

or this, in case of a relative path 
image

but if you the Edit button for the second time, it looks like this 
image and of course this doesn't work.

Also it's not possible to edit the path manually.

yanchith:
@janper

> Technically it works, but the path looks like this

Didn't test on windows yet. Looks like a windows string encoding thing. I'll look into it.

> or this, in case of a relative path

That looks correct to me. Is there a problem in how the relative path looks?

> Also it's not possible to edit the path manually.

That's on purpose. Do we want it editable?

> but if you the Edit button for the second time, it looks like this

How did you manage to do that? :joy: EDIT: Ok, I see. Rookie mistake on my part.

janper:
> That looks correct to me. Is there a problem in how the relative path looks?

I'd expect it to start with `.\`. On the other hand, relative paths are not so common on windows, so I guess there is no standard.

janper:
> That's on purpose. Do we want it editable?

hmmm... if copypaste works, then probably yes. if it doesn't, then don't bother

janper:
> How did you manage to do that? 😂 EDIT: Ok, I see. Rookie mistake on my part.

I tested thoroughly. Clicked the button twice

yanchith:
> > That looks correct to me. Is there a problem in how the relative path looks?
> 
> I'd expect it to start with `.\`. On the other hand, relative paths are not so common on windows, so I guess there is no standard.

Relative paths without `./` are valid AFAIK, but it's not a problem to add if you think it's important.

yanchith:
> hmmm... if copypaste works, then probably yes. if it doesn't, then don't bother

It doesn't. We have a low prio task to fix it.

janper:
> It doesn't. We have a low prio task to fix it.

Then let's not do it

janper:
> Relative paths without ./ are valid AFAIK, but it's not a problem to add if you think it's important.

Nah, let's not

janper:
image

yanchith:
~Yeah, the width of the button is hardcoded and I changed the content...~ Nope, it isn't hardcoded, I just forgot to use the computed width 😄

janper:
yup, it's just this now 
image

janper (approved):
Great, thanks!

PR #90, opened 2019-12-10
Return option for loop subdivision

janper:
If the loop subdivision function fails because of invalid input mesh (not triangulated or non-manifold), the interpreter doesn't hang but stops gracefully and allows the user to set proper parameters.

We might consider returning a more verbose refusal than `None` but I don't know how :-/ For the time being, this seems to be good enough.

### DIFF
--- a/src/mesh/smoothing.rs
+++ b/src/mesh/smoothing.rs
@@ -318,9 +318,12 @@ pub fn loop_subdivision(
                     mid_vertex_indices[edge_index] = Some(mid_vertex_index);
                 }
 
-                let mid_v1v2_index = mid_vertex_indices[0]?;
-                let mid_v2v3_index = mid_vertex_indices[1]?;
-                let mid_v3v1_index = mid_vertex_indices[2]?;
+                let mid_v1v2_index =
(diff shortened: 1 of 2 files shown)


src/mesh/smoothing.rs
```diff
@@ -320,12 +318,9 @@ pub fn loop_subdivision(
...
 
-                let mid_v1v2_index =
-                    mid_vertex_indices[0].expect("Must have been produced by earlier loop");
-                let mid_v2v3_index =
-                    mid_vertex_indices[1].expect("Must have been produced by earlier loop");
-                let mid_v3v1_index =
-                    mid_vertex_indices[2].expect("Must have been produced by earlier loop");
+                let mid_v1v2_index = mid_vertex_indices[0]?;
+                let mid_v2v3_index = mid_vertex_indices[1]?;
+                let mid_v3v1_index = mid_vertex_indices[2]?;
```
yanchith: Did this really crash here or you just weren't sure? From what I remember this really shouldn't happen, unlike the other condition that you fixed.

yanchith (commented):
Looks good! One question...

yanchith (approved):

PR #92, opened 2019-12-10
Fix collinearity, barycentric coordinates and point-in-line evaluation

janper:
There were several mathematical misconceptions in the affected functions. This PR should fix them and prepare to be used in compound operations, such as voxelization.


src/geometry.rs
```diff
@@ -27,7 +27,7 @@ pub fn compute_barycentric_coords(
     let ys = Vector3::new(ac.y, ab.y, pa.y);
     let ortho = xs.cross(&ys);
 
-    if f32::abs(ortho.z) < 1.0 {
+    if f32::abs(ortho.z) == 0.0 {
```
yanchith: One question remains... `==` vs `approx_eq`?

src/geometry.rs
```diff
@@ -39,11 +39,18 @@ pub fn compute_barycentric_coords(
...
+pub fn are_points_collinear(v0: &Point3<f32>, v1: &Point3<f32>, v2: &Point3<f32>) -> bool {
+    if v0.coords.relative_eq(&v1.coords, 0.0001, 0.0001)
+        || v0.coords.relative_eq(&v2.coords, 0.0001, 0.0001)
+    {
+        return true;
+    }
+    let n1 = v1 - v0;
+    let n2 = v2 - v0;
+    let cross = n1.cross(&n2);
+    cross.relative_eq(&Vector3::zeros(), 0.001, 0.001)
```
yanchith: This `relative_eq` has a different number of zeros than the ones above :smile: Can't we use default epsilon here?

yanchith (commented):
Looks good! A couple of questions.

src/geometry.rs
janper: you decide...

src/geometry.rs
janper: intentionally. the tolerance can be more benevolent for points (we don't know their scale) and should be more strict for vectors.
btw, i'm not sure how to use default epsilon with nalgebra - they always use number literals in their examples.

src/geometry.rs
yanchith: I'd go for approx for floats instead of  `==` 99% of the time.

src/geometry.rs
yanchith: > btw, i'm not sure how to use default epsilon with nalgebra - they always use number literals in their examples.

same as us, `nalgebra` leans on the `approx` crate for float comparisons. `approx` defines the following macros that fill in the default epsilons for you:

- `relative_eq!`
- `ulps_eq!`
- `abs_diff_eq!`

We should go for relative_eq most of the time.

https://docs.rs/approx/0.3.2/approx/macro.relative_eq.html

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&code=use%20nalgebra%3B%20%2F%2F%200.19.0%0Ause%20approx%3B%20%2F%2F%200.3.2%0A%0Afn%20main()%20%7B%0A%20%20%20%20let%20a%20%3D%20nalgebra%3A%3AVector3%3A%3Anew(1.0%2C%202.0%2C%203.0)%3B%0A%20%20%20%20let%20b%20%3D%20nalgebra%3A%3AVector3%3A%3Anew(1.0000000000000001%2C%202.0%2C%203.0)%3B%0A%20%20%20%20if%20approx%3A%3Arelative_eq!(a%2C%20b)%20%7B%0A%20%20%20%20%20%20%20%20println!(%22relative%20eq%22)%3B%0A%20%20%20%20%7D%20else%20%7B%0A%20%20%20%20%20%20%20%20println!(%22not%20relative%20eq%22)%3B%0A%20%20%20%20%7D%0A%7D

yanchith: The important thing to realize is that there are traits for floating point comparisons that approx defines and nalgebra implements. That is why you can use approx macros with nalgebra types.

src/geometry.rs
```diff
@@ -39,11 +39,18 @@ pub fn compute_barycentric_coords(
...
-pub fn are_points_colinear(v0: &Point3<f32>, v1: &Point3<f32>, v2: &Point3<f32>) -> bool {
-    let v0_normalized = v0.coords.normalize();
-    let v1_normalized = v1.coords.normalize();
-    let v2_normalized = v2.coords.normalize();
-    v0_normalized == v1_normalized && v0_normalized == v2_normalized
+///
+/// http://www.ambrsoft.com/TrigoCalc/Line3D/LineColinear.htm
+pub fn are_points_collinear(v0: &Point3<f32>, v1: &Point3<f32>, v2: &Point3<f32>) -> bool {
+    if v0.coords.relative_eq(&v1.coords, 0.0001, 0.0001)
+        || v0.coords.relative_eq(&v2.coords, 0.0001, 0.0001)
```
yanchith: We should also use default epsilons here:
```rust
    if approx::relative_eq!(v0, v1) || approx::relative_eq!(v0, v2) {
        return true;
    }
```

I checked and the tests pass.

yanchith (dismissed):

yanchith:
By the way, you don't need to extract coords from points to compare. Points can be compared directly.

yanchith (approved):

PR #93, opened 2019-12-10
Styling

yanchith:
The PR adds a new theme (`funky`), changes the font to a monospace one, and tweaks little details pretty much everywhere.


yanchith:
@janper I actually screwed up. Previously the background color wasn't light orange, but truly white, because I failed at math :joy: 

I now committed the fix, but please verify you like the new, fixed visuals better. I do, personally.

yanchith:
I noticed one more problem - selecting variables in combo boxed is unreadable in the funky theme. I'll try to see if that can be fixed.

janper (approved):
It looks great and I also like that we have an option to use the dark theme.

PR #94, opened 2019-12-11
Also allow capital OBJ extension

yanchith:
tinyfiledialogs is case sensitive, even if windows is not...


ondrowan:
It's quite late and I don't have time for further testing, but from what I've tried on my Windows machine, opening `oBj`, `OBJ` and such works even on master for me. I did some digging and NTFS seems to have an option to be case sensitive, but I don't know why any of you would have it turned on. But in the end I don't think this should affect how open dialogs work.

image

I checked also the implementation of `tinyfiledialogs` and it just creates `OPENFILENAMEW` structure and calls `GetOpenFileNameW` (both native functions from Windows API). Documentation doesn't mention anything about case sensitivity being a problem.

Since you've tried this under time pressure today, I'd probably retest this on your machines once again and see if it wasn't just some simple mistake.

yanchith:
Agreed. I'll try to dig a bit deeper too. I'm pretty sure about what I saw, but I'd like to understand why too.

yanchith:
@ondrowan hmm maybe doing more research here is irrelevant. On linux the file dialog is case sensitive and we would like it to accept capital case OBJ as well. Don't have macOS handy, but I expect it to be the same case there. Gonna check windows now.

yanchith:
Ok, don't know what happened on that day, but yeah, Janper's windows indeed behaves case insensitively here. I think we should still merge for the other two OSes though.

janper (approved):
Let's not spend more time with this

PR #97, opened 2019-12-14
Update wgpu and winit

yanchith:
This also adds one small optimization: screen textures and the swapchain are now only re-allocated on change. This should make screen resizing a bit faster and prevent some unnecessary allocations.

This resolves the following bugs:

- fullscreen on macOS having ultra low framerate or outright crashing the macOS
- wgpu related memory leaks:
    - Metal on macOS can still explosively grow in memory use, but now after some time always settles back down and no longer grows infinitely
    - Vulkan on Windows (and Linux) still has good memory use
    - D3D12 on Windows still has the same memory use (it still consumes about 5x of what Vulkan consumes)
- No noticeable FPS drop on any platform. I tested 10 minutes each of: Metal on macOS, Vulkan on Linux, Vulkan on Windows and D3D12 on Windows.

Other bugs, such as crashing on Windows when app is minimized are not fixed.

Please do your own tests and let me know if you find any regression.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,14 +57,6 @@ dependencies = [
 
 [[package]]
 name = "arrayvec"
-version = "0.4.11"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
- "nodrop 0.1.13 (registry+https://github.com/rust-lang/crates.io-index)",
-]
-
-[[package]]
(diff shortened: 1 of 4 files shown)


janper (approved):
Looks okay
vulkan-startup
vulkan-5-min
d3d12-startup
d3d12-5-min

PR #98, opened 2019-12-14
Work around exclusive fullscreen issues on macOS

yanchith:
Simply do not try to do exclusive fullscreen on macOS at all until
the issues are resolved upstream.

Note: this is not reviewable until #97 lands, as that contains a fix that makes fullscreen on macOS not crash your computer 🐈


yanchith:
@ondrowan gonna merge tomorrow unless there are objections.

PR #99, opened 2019-12-14
Exec empty progs

yanchith:
This adds the possibility of executing empty programs and thus resetting the viewport to empty state.


janper (approved):
<3

PR #100, opened 2019-12-14
Unify primitive generators and add box func

yanchith:
Now all primitive generating functions take `center`, `rotate`, `scale` and are able to transform the primitives they generate on their own. One exception is plane, whose `Plane` parameter already contains rotation.

All func definitions have been updated to take advantage of this fact.

Also added is the "Create Box" interpreter func.

This is precursor work for returning box as a default value when an error happens and there is no original mesh to return - e.g. when we are importing.

### DIFF
--- a/src/interpreter/ast.rs
+++ b/src/interpreter/ast.rs
@@ -156,17 +156,14 @@ impl fmt::Display for Expr {
 /// An expression that evaluates to a constant, literal value.
 #[derive(Debug, Clone, PartialEq)]
 pub enum LitExpr {
-    #[allow(dead_code)]
     Nil,
     #[allow(dead_code)]
     Boolean(bool),
     #[allow(dead_code)]
     Int(i32),
-    #[allow(dead_code)]
     Uint(u32),
-    #[allow(dead_code)]
     Float(f32),
-    #[allow(dead_code)]
+    Float2([f32; 2]),
     Float3([f32; 3]),
     String(Arc<String>),
 }
@@ -216,6 +213,17 @@ impl LitExpr {
         }
     }
 
+    /// Get the literal value if float2, otherwise panic.
+    ///
+    /// # Panics
+    /// This function panics when literal value is not a float2.
+    pub fn unwrap_float2(&self) -> [f32; 2] {
+        match self {
+            LitExpr::Float2(float2) => *float2,
+            _ => panic!("Literal expression not float2"),
+        }
+    }
+
     /// Get the literal value if float3, otherwise panic.
     ///
     /// # Panics
@@ -247,6 +255,7 @@ impl fmt::Display for LitExpr {
             LitExpr::Int(int) => write!(f, "<int {}>", int),
             LitExpr::Uint(uint) => write!(f, "<uint {}>", uint),
             LitExpr::Float(float) => write!(f, "<float {}>", float),
+            LitExpr::Float2(float2) => write!(f, "<float2 [{}, {}]>", float2[0], float2[1]),
             LitExpr::Float3(float3) => {
                 write!(f, "<float3 [{}, {}, {}]>", float3[0], float3[1], float3[2])
             }
--- a/src/interpreter/func.rs
+++ b/src/interpreter/func.rs
@@ -53,6 +53,7 @@ pub enum ParamRefinement {
     Int(IntParamRefinement),
     Uint(UintParamRefinement),
     Float(FloatParamRefinement),
+    Float2(Float2ParamRefinement),
     Float3(Float3ParamRefinement),
     String(StringParamRefinement),
     Mesh,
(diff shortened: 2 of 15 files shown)


yanchith:
One side note is that this also marks all func optional parameters as non-optional. The optional parameter feature wasn't really used, because the UI filled in the default value anyway. I didn't remove the optional parameters feature altogether, because we still might want to use it for something in the future, e.g. for parameters that don't have a good default value and can only sometimes be provided, like meshes.

janper (approved):
I'm not very happy about the implementation. Sure, it's more consistent than it was before but less elegant than using our own `Plane` type. 
Also you are using `Vector2` not as a vector but as a couple of values (geometry dimensions). I was intentionally avoiding this because semantically it's not a vector and you can't use vector operations for such values. The more generic `[f32; 2]` seems more appropriate. Similarly, we distinguish between a `Vector3` and `Point3` even though they are interchangeable and compatible (in other words it makes sense to threat the data as a point in one case and as a vector at other times) but because they are semantically different we don't mix them up. 

I'm not going to be like that and I'm approving this because it works but I'd like to see `// FIXME` notes where we could have used `Plane` and generic arrays instead of the nalgebra types. Also this is a very low priority task and I don't want to ask you to spend any more time on it.

janper:
The UI is okay the way it is. 
Also I don't mind using any nalgebra types inside of the functions, but I'd like to see the functions looking like this: `ellipsoid(base: Plane, radii: [f32;3])` `box(base: Plane, dimenstions: [f32;3])`. 
I know you did it technically correct but it's not very easy to use. 
I also hear you saying there are various conventions. Whenever I used any framework or an end-user software, there were many ways of achieving the same result. I also know other, equally valid conventions, for example `box(corner1: Point3, point2: Point3)` an then rotate (processing) or simply a unit box and then move, scale and rotate (Cinema4D, Blender). I'm proposing the convention taken from Rhino / Grasshopper. Also in Rhino/Grasshopper there are many box constructors. 
We rather only keep one way of doing things. Okay. So let's pick the most useful one. For me it's the one I mentioned for these reasons:

- The plane is easy to imagine and visualize. That's why it's used in Grasshopper. I know we don't visualize it but we need to use the functions and imagining or visualizing a rotation matrix is impossible for me.
- The Plane as we have it is still a vector math. It's easier for me to understand and handle than matrices. At the same time I don't mind the Plane use matrices under the hood!
- We can easily compare our results to Grasshopper results as long as we keep the same conventions.
- It's one input parameter instead to two (translate, rotate)
- regarding using the `Vector` for sizes, I'm surprised with your decision, knowing how strictly you are against any redundancies. There may be some reason for treating a voxel diagonal as a vector (or is there any?) but it makes no sense for an ellipsoid. 
- The dimensions are not a vector but rather three values bound to the plane X, Y and Z axes. There is never any case where we would use any functions related to vectors when it comes to dimensions.
- If the geometry is rotated, the diagonal vector is different from a diagonal geometry of the same size aligned to the world XY coordinate system.
- I started using nalgebra types as arguments only where there existed a special type for that specific argument - a Point a Plane (our own) or maybe a Vector (did I?). The size of a voxel, a box or an ellipsoid isn't one of the nalgebra types, therefore I decided to keep them as generic/basic as possible. 

All that being said, let's not make any more changes. I'll make the changes when/if needed for some other functionality. Otherwise we'll keep it as it is now, there is no need for another refactor.

yanchith:
@janper 

I already replied something similar prior on slack, but putting it here just for completeness.

In PR #86 you removed cube (which used `[f32; 3]`) and added box (which uses `Point3<f32>`).
https://github.com/subdgtl/HURBAN-Selector/pull/86/files#diff-659f0576442a890d51ad7f4527e8593bR30
I viewed that as intentional from your side and that's what motivated the conventions I picked here. 

Regarding `plane: Plane` vs `center: Point3<f32>` + `rotation: Rotation3<f32>`: Your points about ease of understanding and visualization make sense. I wouldn't mind changing center+rotation to Plane later for box and sphere (and any other primitives we might have by then). If we do so, we should provide conversions between the two representations for interoperability, because if nothing else, the UI currently provides center+rotation. You mention experience from end-user software: shouldn't we use plane in the UI too?

Regarding scale as `Vector3<f32>` vs `[f32; 3]`: This really depends on through what lens you view the vector. Either an arrow in space or a collection of numbers. Even in linear algebra it is sometimes insufficient to view it as just an arrow in space (e.g. look at the code that computes barycentric coordinates and the way it deconstructs vectors). Also, there are examples in the wild using `Vector3` for scale, so it's not without precedence:

- https://docs.rs/nalgebra/0.19.0/nalgebra/base/type.MatrixN.html#method.new_nonuniform_scaling
- https://docs.unity3d.com/ScriptReference/Vector3.Scale.html

In the end, this is all about conventions. There were no conventions present in the code I was refactoring (scale was just an `f32` and there was no rotation yet for box and sphere), so I picked some. I don't mind if we change them to something else, only that we should be consistent, when we do so.

PR #101, opened 2019-12-16
Voxelize

janper:
Introduces new type: `VoxelCloud`, which is a bit map containing voxel states and information about the voxel space. The `VoxelCloud` can be created form an existing mesh and converted into a boundary mesh.
There is a new operation for voxelizing an existing mesh.


src/mesh/primitive.rs
```diff
@@ -22,7 +22,7 @@ pub fn create_mesh_plane(plane: Plane, scale: Vector2<f32>) -> Mesh {
             + ( 0.5 * plane.y_vector() * scale.y),
     ];
 
-    let vertex_normals = vec![Vector3::new(0.0, 0.0, 1.0)];
+    let vertex_normals = vec![plane.normal()];
```
yanchith, line 25: Nice catch!

src/pull.rs
```diff
@@ -229,6 +229,17 @@ fn is_point_in_triangle(
...
+            Rotation3::rotation_between(&(-1.0 * plane.normal()), &Vector3::new(0.0, 0.0, 1.0))
+        {
+            // In case the triangle isn't horizontal rotate the triangle to
+            // become horizontal and rotate also the test point.
+            (
+                rotation_to_horizontal * triangle_vertex0,
+                rotation_to_horizontal * triangle_vertex1,
+                rotation_to_horizontal * triangle_vertex2,
+                rotation_to_horizontal * point,
+            )
```
yanchith: Hmmm, not sure I understand why negating the normal would work if the previous check fails.

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+use std::iter;
+
+use nalgebra::{Point3, Vector2, Vector3};
+
+use crate::convert::cast_usize;
+use crate::geometry;
+use crate::mesh::analysis::BoundingBox;
+use crate::mesh::primitive;
+use crate::mesh::tools;
+use crate::mesh::{Face, Mesh};
```
yanchith: Let's make these exports `super` instead of `crate`. That should make it easier to separate the `mesh` module to it's own crate later. (`super` imports should be below `crate` imports)

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+/// each direction.
+///
+/// The voxels are located at discrete coordinates. The voxel space starts at
+/// the cartesian space origin with voxel coordinates 0, 0, 0.
+#[derive(Debug, Clone)]
+pub struct VoxelCloud {
+    block_start: Point3<i32>,
+    block_dimensions: Vector3<u32>,
+    voxel_dimensions: Vector3<f32>,
+    // FIXME: change this to a bit map
```
yanchith: `FIXME: @Optimization Change this to a bit vector`

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+    voxel_map: Vec<bool>,
+}
+
+impl VoxelCloud {
+    // Define a new empty block of voxel space, which begins at `block_start`(in
```
yanchith: Doc comments begin with 3 slashes.

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+use crate::convert::cast_usize;
+use crate::geometry;
+use crate::mesh::analysis::BoundingBox;
+use crate::mesh::primitive;
+use crate::mesh::tools;
+use crate::mesh::{Face, Mesh};
+use crate::plane::Plane;
+
+/// A public type for setting/getting a voxel state
+pub enum Voxel {
```
yanchith: Since you never want to set an `OutOfBoundsVoxel`, all the setters should take `bool` and all the getters should return `Option<bool>` and this type is not necessary. As an example: `std::vec::Vec::get` also returns an `Option<T>` that is `None` if the index is out of bounds.

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+            voxel_map: vec![false; cast_usize(map_length)],
+        }
+    }
+
+    /// Calculate a voxel cloud from an existing mesh.
+    pub fn from_mesh(mesh: &Mesh, voxel_dimensions: &Vector3<f32>) -> Self {
+        // Determine the needed block of voxel space.
+        let b_box = BoundingBox::from_meshes(iter::once(mesh));
+
+        let min_x_index =
```
yanchith: Ugh, please extract minimum and maximum points to variables for readability :slightly_smiling_face:

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+            (b_box.minimum_point().y.max(b_box.maximum_point().y) / voxel_dimensions.y).ceil();
+        let min_z_index =
+            (b_box.minimum_point().z.min(b_box.maximum_point().z) / voxel_dimensions.z).floor();
+        let max_z_index =
+            (b_box.minimum_point().z.max(b_box.maximum_point().z) / voxel_dimensions.z).ceil();
+
+        let block_start = Point3::new(
+            min_x_index as i32 - 1,
+            min_y_index as i32 - 1,
+            min_z_index as i32 - 1,
```
yanchith: Let's use our own `cast_i32`. It shouldn't really be the case here, but sometimes float to integer conversion can lead to something really terrible where a crash would be preferable (https://github.com/rust-lang/rust/issues/10184)

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+        );
+        let block_dimensions = Vector3::new(
+            (max_x_index - min_x_index + 2.0) as u32,
+            (max_y_index - min_y_index + 2.0) as u32,
+            (max_z_index - min_z_index + 2.0) as u32,
```
yanchith: Similar to above, let's use our own `cast_u32`

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+                    let point_c = &mesh.vertices()[cast_usize(triangle_face.vertices.2)];
+                    // Calculate the density of points on the respective face
+                    let ab_distance_sq = nalgebra::distance_squared(point_a, point_b);
+                    let bc_distance_sq = nalgebra::distance_squared(point_b, point_c);
+                    let ca_distance_sq = nalgebra::distance_squared(point_c, point_a);
+                    let longest_edge_len = ab_distance_sq
+                        .max(bc_distance_sq.max(ca_distance_sq))
+                        .sqrt();
+                    // Number of face divisions (points) in each direction
+                    let divisions = (longest_edge_len / shortest_voxel_dimension).ceil() as usize;
```
yanchith, line 101: `cast_usize()`

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+            Voxel::On
+        } else {
+            Voxel::Off
+        }
+    }
+
+    /// Gets the state of a voxel containing the input point defined in model
+    /// space coordinates.
+    #[allow(dead_code)]
+    pub fn get_cartesian(&self, point: &Point3<f32>) -> Voxel {
```
yanchith: `get_` prefix is not used in rust 99% of time

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+
+        voxel_cloud
+    }
+
+    /// Gets the index to the voxel map from voxel coordinates relative to the
+    /// voxel space block start.
+    ///
+    /// The function doesn't clamp the result because it may be used also to
+    /// check if the coordinates are out of bounds.
+    fn relative_three_dimensional_coordinate_to_one_dimensional(
```
yanchith: Put private functions below public.

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+                }
+            }
+        }
+
+        voxel_cloud
+    }
+
+    /// For each existing voxel turn on all neighbor voxels to grow (offset) the
+    /// volumes stored in the voxel cloud.
+    pub fn grow(&self) -> Self {
```
yanchith, line 133: Couldn't this be more efficient if it were a `pub fn grow(&mut self) {`? If the caller needs the original, they can clone. (I don't really know why the space of the grown voxel space has to be different than the current one)

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+        // If the voxels in the existing voxel cloud reach the boundaries of the
+        // block, it's needed to grow the block by 1 in each direction.
+        let new_block_start = self.block_start - Vector3::new(-1, -1, -1);
+
+        let new_block_dimensions = self.block_dimensions + Vector3::new(2, 2, 2);
```
yanchith, line 138: It is not clear to me why we need to also grow the whole voxel space here. What is the reasoning?

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+            &new_block_start,
+            &new_block_dimensions,
+            &self.voxel_dimensions,
+        );
+
+        // Iterate through the existing voxel cloud.
+        for z_coord in 0..self.block_dimensions.z as i32 {
+            for y_coord in 0..self.block_dimensions.y as i32 {
+                for x_coord in 0..self.block_dimensions.x as i32 {
+                    let voxel_coords = Point3::new(x_coord as i32, y_coord as i32, z_coord as i32);
```
yanchith: These `as i32` casts are redundant.

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+        let mut voxel_cloud = VoxelCloud::new(
+            &new_block_start,
+            &new_block_dimensions,
+            &self.voxel_dimensions,
+        );
+
+        // Iterate through the existing voxel cloud.
+        for z_coord in 0..self.block_dimensions.z as i32 {
+            for y_coord in 0..self.block_dimensions.y as i32 {
+                for x_coord in 0..self.block_dimensions.x as i32 {
```
yanchith: `cast_i32`

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+    pub fn set_relative(&mut self, value: Voxel, relative_coords: &Point3<i32>) {
+        let index = self.relative_three_dimensional_coordinate_to_one_dimensional(relative_coords);
+        assert!(
+            index >= 0 && index < self.voxel_map.len() as i32,
+            "Coordinates out of bounds"
+        );
+        match value {
+            Voxel::On => self.voxel_map[cast_usize(index)] = true,
+            Voxel::Off => self.voxel_map[cast_usize(index)] = false,
+            Voxel::OutOfBounds => panic!("Can't set a voxel to be out of bounds"),
```
yanchith: What if we made this impossible by making `value: bool` instead of `Voxel`?

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+
+    /// Gets the index to the voxel map from voxel coordinates relative to the
+    /// voxel space block start.
+    ///
+    /// The function doesn't clamp the result because it may be used also to
+    /// check if the coordinates are out of bounds.
+    fn relative_three_dimensional_coordinate_to_one_dimensional(
+        &self,
+        relative_coords: &Point3<i32>,
+    ) -> i32 {
```
yanchith: Shouldn't this return a `u32` or `usize`? The "one dimensional" index is indexing into an array, so it should never be negative. Also please use `cast_i32`. The `as` operator does a bitwise transmute, which you almost certainly don' want :joy: 

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&code=fn%20main()%20%7B%0A%20%20%20%20%2F%2F%20%60as%20i32%60%20does%20bitwise%20transmute.%20Signed%20integers%20are%20in%20Two%27s%20complement%20representation.%0A%20%20%20%20println!(%22%7B%7D%22%2C%200u8%20as%20i8)%3B%0A%0A%20%20%20%20println!(%22%7B%7D%22%2C%20127u8%20as%20i8)%3B%0A%20%20%20%20println!(%22%7B%7D%22%2C%20128u8%20as%20i8)%3B%0A%0A%20%20%20%20println!(%22%7B%7D%22%2C%20254u8%20as%20i8)%3B%0A%20%20%20%20println!(%22%7B%7D%22%2C%20255u8%20as%20i8)%3B%0A%7D

The same applies to the other coordinate computing functions.

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+        self.relative_three_dimensional_coordinate_to_one_dimensional(&relative_coords)
+    }
+
+    /// Calculates the voxel-space coordinates of a voxel containing the input
+    /// point.
+    fn cartesian_to_voxel_coords(&self, point: &Point3<f32>) -> Point3<i32> {
+        Point3::new(
+            (point.x / self.voxel_dimensions.x).round() as i32,
+            (point.y / self.voxel_dimensions.y).round() as i32,
+            (point.z / self.voxel_dimensions.z).round() as i32,
```
yanchith: `cast_i32`

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+/// Voxel cloud is an abstract representation of points in a block of space. The
+/// block is delimited by its beginning and its dimensions, both in the units of
+/// the voxels. All voxels have the same dimensions, which can be different in
+/// each direction.
+///
+/// The voxels are located at discrete coordinates. The voxel space starts at
+/// the cartesian space origin with voxel coordinates 0, 0, 0.
+#[derive(Debug, Clone)]
+pub struct VoxelCloud {
+    block_start: Point3<i32>,
```
yanchith, line 25: I am not sure what `block_start` stands for. Is this the position of the base of the voxel clout in world space? Shouldn't it be a floating point vector in this case?

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+            (point.x / self.voxel_dimensions.x).round() as i32,
+            (point.y / self.voxel_dimensions.y).round() as i32,
+            (point.z / self.voxel_dimensions.z).round() as i32,
+        )
+    }
+
+    /// Calculates the center of a voxel in model-space coordinates from
+    /// absolute voxel coordinates (relative to the voxel space origin).
+    #[allow(dead_code)]
+    fn absolute_voxel_to_cartesian_coords(&self, absolute_coords: &Point3<i32>) -> Point3<f32> {
```
yanchith: This is not even public. If it's not used now, it won't be used later, because it is not visible.

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+        )
+    }
+
+    /// Calculates the center of a voxel in model-space coordinates from voxel
+    /// coordinates relative to the voxel block start.
+    fn relative_voxel_to_cartesian_coords(&self, relative_coords: &Point3<i32>) -> Point3<f32> {
+        Point3::new(
+            (relative_coords.x as i32 + self.block_start.x) as f32 * self.voxel_dimensions.x,
+            (relative_coords.y as i32 + self.block_start.y) as f32 * self.voxel_dimensions.y,
+            (relative_coords.z as i32 + self.block_start.z) as f32 * self.voxel_dimensions.z,
```
yanchith: `cast_i32`

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,499 @@
...
+        );
+
+        // Iterate through the existing voxel cloud.
+        for z_coord in 0..self.block_dimensions.z as i32 {
+            for y_coord in 0..self.block_dimensions.y as i32 {
+                for x_coord in 0..self.block_dimensions.x as i32 {
+                    let voxel_coords = Point3::new(x_coord as i32, y_coord as i32, z_coord as i32);
+                    let voxel_state = self.get_relative(&voxel_coords);
+                    // if the voxel is on
+                    if let Voxel::On = voxel_state {
```
yanchith: Unless there is a value to destructure, a simple comparison is more readable and terser: `if voxel_state == Voxel::On {`

yanchith (changes requested):
I didn't finish reading the PR (I skipped the implementation of `to_mesh` for now), but I have some thoughts I want to share and some questions to ask before proceeding with the review.

I think this needs some explanation of the coordinate systems used. I am not really sure what `block_start` stands for and in which coordinate space. I initially thought it was in the voxel cloud space and the mapping to world space had to be made externally, but I am not sure after reading the rest of the implementation. Why are both absolute and relative coords used?

Also important: except `as f32` and `as f64`, all primitive operator casts shouldn't really be used. They exist in rust for historical reasons and while their behavior is well defined in most cases, it is not very intuitive (unless you consider two's complement signed integer representation intuitive).

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&code=fn%20main()%20%7B%0A%20%20%20%20%2F%2F%20%60as%20i32%60%20does%20bitwise%20transmute.%20Signed%20integers%20are%20in%20Two%27s%20complement%20representation.%0A%20%20%20%20println!(%22%7B%7D%22%2C%200u8%20as%20i8)%3B%0A%0A%20%20%20%20println!(%22%7B%7D%22%2C%20127u8%20as%20i8)%3B%0A%20%20%20%20println!(%22%7B%7D%22%2C%20128u8%20as%20i8)%3B%0A%0A%20%20%20%20println!(%22%7B%7D%22%2C%20254u8%20as%20i8)%3B%0A%20%20%20%20println!(%22%7B%7D%22%2C%20255u8%20as%20i8)%3B%0A%7D

yanchith:
Naively, without seeing this implementation, I would expect there to be two getters and two setters:

- `pub fn voxel(coords: Vector3<u32>) -> Option<bool>;`
- `pub fn voxel_world_space(coords: Vector3<f32>) -> Option<bool>;`
- `pub fn set_voxel(coords: Vector3<u32>, value: bool);`
- `pub fn set_voxel_world_space(coords: Vector3<f32>, value: bool);`

I am not exactly sure what the relative coordinates in this PR are, but maybe they just need more explanation.

janper:
I'm going to implement the technical requests but for the rest you'll have to provide some arguments. More in my replies.

src/pull.rs
janper: It's a nalgebra thing: it's not clear when `rotation_between` returns None, but if it does, you can try to rotate an upside-down version of the triangle, which will be equivalent to what you wanted to achieve in the first place.

src/mesh/voxel_cloud.rs

src/mesh/voxel_cloud.rs

src/mesh/voxel_cloud.rs

src/mesh/voxel_cloud.rs
janper: how would you then check if the voxel is off OR out of bounds at the same time?

```
match self.get_relative(&(voxel_coords + Vector3::new(0, 0, 1))) {
    // if there isn't or if the current voxel is on the
    // boundary of the voxel space block
    Voxel::Off | Voxel::OutOfBounds => {
        // add a horizontal rectangle
        plane_meshes.push(primitive::create_mesh_plane(
            Plane::from_origin_and_plane(
                // above the voxel center half way the height of the voxel
                &(voxel_center + upwards_voxel_direction),
                // align it properly
                &top_plane,
            ),
            // and set its size to match the dimensions
            // of the top side of a voxel
            top_bottom_voxel_dimensions,
        ));
    }
    // if there is a neighbor above the current voxel,
    // it means no boundary side of the voxel box should
    // be materialized
    _ => {}
}
```

src/mesh/voxel_cloud.rs

src/mesh/voxel_cloud.rs

src/mesh/voxel_cloud.rs

src/mesh/voxel_cloud.rs
janper: ok

src/mesh/voxel_cloud.rs
janper: suggestions?

src/mesh/voxel_cloud.rs

src/mesh/voxel_cloud.rs
janper: You need to grow the space, otherwise you might (and certainly will in the second iteration) grow out of bounds. Therefore you need to allocate a new chunk of memory, so I don't see any benefits of mutating the existing `VoxelCloud`.

src/mesh/voxel_cloud.rs
janper: see above

src/mesh/voxel_cloud.rs

src/mesh/voxel_cloud.rs

src/mesh/voxel_cloud.rs
janper: Trying to be symmetrical with the getters.

src/mesh/voxel_cloud.rs
janper: No, see the comment right above and how this function is used. Plus, it's valid and more explicit to convert from one unit system to another without clamping. The caller can decide what to do with such cases. And i always treat the out-of-bounds scenarios where they should be treated. Also see how this is (second-hand) used in the `to_mesh` method.

src/mesh/voxel_cloud.rs

src/mesh/voxel_cloud.rs
janper: Certainly not. Block start is a coordinate in the voxel-space coordinate system describing the beginning of the voxel cloud block (basically a box-shaped chunk of voxel space). The voxel-space is world-aligned and starts at the world origin. I'm also describing this in the comments (line 21-27). The voxel space is discrete and you can't start it half way in a voxel, therefore its beginning has to be defined in the voxel-space coordinates. This way we keep voxel clouds with the same voxel size compatible and can perform collateral operations on them.

src/pull.rs
yanchith: I know. But it fails for zero vectors. If you negate a zero vector...

src/mesh/voxel_cloud.rs
yanchith: Out of bounds is not a voxel property. Coordinates can be out of bounds. Check if the coordinates are out of bounds OR if they are within bounds, check whether the voxel is off.

src/mesh/voxel_cloud.rs
yanchith: `cartesian`

src/mesh/voxel_cloud.rs
janper: Disagree! The `VoxelCloud` is a prominent data type and will soon be as important as the `Mesh` itself. Therefore whenever I'm implementing any functionality (in this case coordinate conversion), I'm doing all of it. Besides, this will be useful for volumetric boolean operations.

src/mesh/voxel_cloud.rs

src/mesh/voxel_cloud.rs

janper:
> I am not exactly sure what the relative coordinates in this PR are, but maybe they just need more explanation.

I thought they are explained in the comments, but ok:

- for **absolute voxel coords**  `[0,0,0]` is equal to the cartesian world origin. `[1,1,1]` is equal to `[voxel_size_x, voxel_size_y, voxel_size_z]` in cartesian model-space coords.
- for **relative voxel coords**  `[0,0,0]` is equal to the `block_start` in voxel-space coords, which is `[block_start_x * voxel_size_x, block_start_y * voxel_size_y, block_start_z * voxel_size_z]` in cartesian model-space corrds. `[1,1,1]` is equal to `[block_start_x + voxel_size_x, block_start_y + voxel_size_y, block_start_z + voxel_size_z]` in absolute voxel-space coords and `[(block_start_x + voxel_size_x) * voxel_size_x, (block_start_y + voxel_size_y) * voxel_size_y, (block_start_z + voxel_size_z) * voxel_size_z]` in cartesian model-space coords
- **model space cartesian coords** are always absolute and they are what we use anywhere else in HURBAN selector. in fact I should be more explicit and rename `cartesian_to_voxel_coords` to `cartesian_to_absolute_voxel_coords` and add `cartesian_to_relative_voxel_coords`

src/pull.rs
janper: okay, I'll check for more edge cases. `rotation_between` is not well documented.

src/mesh/voxel_cloud.rs
janper: ...ok...

src/mesh/voxel_cloud.rs
janper: isn't this an inconsistency? we are horribly verbose everywhere else and here we would just say one word. this is a getter of a voxel state (hence `get`) where the input are cartesian coordinates (hence `_cartesian`). what about `state_at_cartesian_coords` and similarly for other cooridnate systems?

yanchith:
> I'm going to implement the technical requests but for the rest you'll have to provide some arguments. 

The rest were questions, not requests :slightly_smiling_face:

src/mesh/voxel_cloud.rs
yanchith: If you really think this type is important, keep it *internally*, but never accept or return it from a public method.

src/mesh/voxel_cloud.rs
yanchith: Well I was just applying the rule to what you had there, but yeah, I like more explicit naming too. `state_at_cartesian_coords` is nice, or maybe `voxel_at_cartesian_coords`.

src/mesh/voxel_cloud.rs
janper: No, I'm going to change it to `Option<bool>` because it doesn't matter but I don't see why. Also why should I never return or accept it? We do it all over the place with normal strategies, errors, face types, so why not here?

src/mesh/voxel_cloud.rs
yanchith: Ok, I understand now, thanks!

src/mesh/voxel_cloud.rs
yanchith: This `enum` is an implementation detail that leaks through the API, and it is easy to make it so it doesn't. That's why I am asking.

src/mesh/voxel_cloud.rs
yanchith: Ok, I'll try to explain, because this is important. This particular instance is not catastrophic, but it does add needless complexity that is not contained within the module.

This enum is an implementation detail, a sentinel value you use in internal logic to detect out of bounds. The caller of public methods of `VoxelCloud` doesn't need to see it at all. If all setters panic when you give them the `OutOfBounds` value and there never is a valid scenario for the setters to receive this value, simply don't make it possible in the typesystem for them to receive this value. For getters, there already is a precedence: in all rust's collection, the `get` method returns an option if the element is not in the collection or is out of bounds.

I think the enum was an attempt on your part to make symmetric the internal and externally visible functionality, but our goal should be to keep the program complexity down, so we want to hide what we can.

yanchith: > We do it all over the place with normal strategies, errors, face types, so why not here?

You misunderstand. This is not about this type being an enum. This is about this exact enum providing more freedom than strictly necessary.

src/mesh/voxel_cloud.rs
yanchith: > This way we keep voxel clouds with the same voxel size compatible and can perform collateral operations on them.

I think this was the piece of information I was missing. Can you please add it to the comment?

src/mesh/voxel_cloud.rs
janper: `voxel_at_cartesian_coords` it is!

src/mesh/voxel_cloud.rs
yanchith: You are right that you in some cases use the result to detect out of bounds. Wouldn't it be better to make the bounds check here and return `Option<usize>` in this case? Will you ever want to know, exactly how much out of bounds this was?

> /// Gets the index to the voxel map from voxel coordinates ...

The word "index" immediately tells me this should be an unsigned integer.

janper: I prefer changing the comment to changing the logic. Please see how do I use this in the `to_mesh` first. I don't think the clamping semantically belongs to this function even though the `Option` is an option. It, however, adds too much mess to the caller.

yanchith:
@janper 

This conversation is getting long, so sorry if I missed your answer, but I think you still haven't answered my question:

> Why are both absolute and relative coords used?

Or more concretely, why would you ever want to use absolute coords? When we talked about the data structure offline, I always thought that we are going to index it with local coordinates only. The only uses of the absolute getters and setters are currently in `get_cartesian` and `set_cartesian`. When are absolute coords useful?

janper:
> > Why are both absolute and relative coords used?
> 
> Or more concretely, why would you ever want to use absolute coords? When we talked about the data structure offline, I always thought that we are going to index it with local coordinates only. The only uses of the absolute getters and setters are currently in `get_cartesian` and `set_cartesian`. When are absolute coords useful?

When performing operations on more voxel clouds, such as comparing, merging or boolean operations. It is also quite important to understand the voxel clouds as they are implemented, are tightly bound to the cartesian space. Actually, you can consider the voxel space infinite and the fact it's clamped is just an implementation (optimization) detail.

yanchith:
Ok, I think I understand what I need to. Please resolve the unambiguous issues and then we can do another review pass.

src/mesh/voxel_cloud.rs
janper: image

src/mesh/voxel_cloud.rs
janper: image

src/mesh/voxel_cloud.rs
janper: see above

janper:
hopefully everything implemented. ready for another review.

src/mesh/voxel_cloud.rs
yanchith: Well that sucks. Okay, let's skip using `cast_i32` for floating point to integer conversions. Let's still use it for integer conversions though.

EDIT: also don't forget the ever important rounding!

src/mesh/voxel_cloud.rs
yanchith: See my above comment. Let's skip the floating point to integer conversions for now. But at least add the missing `.round()` here. (or `ceil`/`floor`, whatever makes most sense)

src/mesh/voxel_cloud.rs
yanchith: Okay, I'll read `to_mesh`, but I don't expect to find anything there that will change my mind. I already skimmed the implementation and looked at where this is called there before I wrote any of my comments. If you think that `Option` adds "too much mess to the caller", think about how much mess a bounds check that's not there does.

src/mesh/voxel_cloud.rs
```diff
@@ -3,42 +3,36 @@ use std::iter;
...
-    Off,
-    OutOfBounds,
-}
+use super::{analysis::BoundingBox, primitive, tools, Face, Mesh};
 
-/// Voxel cloud is an abstract representation of points in a block of space. The
-/// block is delimited by its beginning and its dimensions, both in the units of
-/// the voxels. All voxels have the same dimensions, which can be different in
-/// each direction.
+/// Option<bool> cloud is an abstract representation of points in a block of
```
yanchith: You probably didn't want to text-replace this to `Option<bool>`

src/mesh/voxel_cloud.rs
```diff
@@ -117,20 +107,21 @@ impl VoxelCloud {
...
                                 // Calculate point position in model space
                                 let cartesian = geometry::barycentric_to_cartesian(
                                     &barycentric,
                                     &point_a,
                                     &point_b,
                                     &point_c,
                                 );
-                                // and set a voxel containing the point to be on
-                                voxel_cloud.set_cartesian(Voxel::On, &cartesian);
+                                // and set_voxel_at_absolute_coords a voxel containing the point to be on
```
yanchith: This comment looks like it suffered from erroneous mass text replace :slightly_smiling_face:

yanchith:
@janper

For all the float to integer conversions where the compiler yelled at you, you don't have to use `cast_*`, but **make sure there is rounding**. I'll add the safe casting function later.

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,456 @@
...
+            absolute_coords.z as f32 * self.voxel_dimensions.z,
+        )
+    }
+
+    /// Calculates the center of a voxel in model-space coordinates from voxel
+    /// coordinates relative to the voxel block start.
+    fn relative_voxel_to_cartesian_coords(&self, relative_coords: &Point3<i32>) -> Point3<f32> {
+        // FIXME: This can look like this
+        // `(relative_coords + self.block_start.coords) * self.voxel_dimensions.coords`
+        // if it is possible to convert from Point3<i32> to Point3<f32>
```
yanchith: I don't think `nalgebra` defines conversions between the vectors, because there's 4 valid conversions you'd like to make, and it cannot choose for you the default one (truncate, floor, ceiling and round). You can write the custom conversions yourself, if you think it's worth it. If you don't think we'll be doing this a lot, I'd just remove this fixme altogether. Same goes for the other fixme a couple of lines above you added.

If you happen to define the custom conversions, make sure you handle edge cases where appropriate.

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,456 @@
...
+        match self.absolute_three_dimensional_coordinate_to_one_dimensional(absolute_coords) {
+            Some(index) => Some(self.voxel_map[index]),
+            None => None,
+        }
+    }
+
+    /// Gets the state of a voxel containing the input point defined in model
+    /// space coordinates.
+    #[allow(dead_code)]
+    pub fn get_cartesian(&self, point: &Point3<f32>) -> Option<bool> {
```
yanchith: I think you forgot to rename this one to `voxel_at_cartesian_coords`.

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,456 @@
...
+    /// Calculates the center of a voxel in model-space coordinates from voxel
+    /// coordinates relative to the voxel block start.
+    fn relative_voxel_to_cartesian_coords(&self, relative_coords: &Point3<i32>) -> Point3<f32> {
+        // FIXME: This can look like this
+        // `(relative_coords + self.block_start.coords) * self.voxel_dimensions.coords`
+        // if it is possible to convert from Point3<i32> to Point3<f32>
+        Point3::new(
+            (cast_i32(relative_coords.x) + self.block_start.x) as f32 * self.voxel_dimensions.x,
+            (cast_i32(relative_coords.y) + self.block_start.y) as f32 * self.voxel_dimensions.y,
+            (cast_i32(relative_coords.z) + self.block_start.z) as f32 * self.voxel_dimensions.z,
```
yanchith: Wait! I didn't notice before, but the relative coords are already i32. There is no need to cast at all here.

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,456 @@
+use std::f32;
+use std::iter;
+
+use nalgebra::{Point3, Vector2, Vector3};
+
+use crate::convert::{cast_i32, cast_usize};
+use crate::geometry;
+use crate::plane::Plane;
+
+use super::{analysis::BoundingBox, primitive, tools, Face, Mesh};
```
yanchith: You don't need to change it here, but for the future, we don't use the nested import style. I would enforce this in rustfmt, but unfortunately it doesn't have this option. The nested style can be quite unreadable, if overdone.

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,456 @@
...
+        voxel_dimensions: &Vector3<f32>,
+    ) -> Self {
+        assert!(
+            block_dimensions.x > 0 && block_dimensions.y > 0 && block_dimensions.z > 0,
+            "One or more block dimensions are 0"
+        );
+        assert!(
+            voxel_dimensions.x > 0.0 && voxel_dimensions.y > 0.0 && voxel_dimensions.z > 0.0,
+            "One or more voxel dimensions are 0.0"
+        );
```
yanchith, line 49: Would be good to at least have these corner cases tested.

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,456 @@
...
+        VoxelCloud {
+            block_start: *block_start,
+            block_dimensions: *block_dimensions,
+            voxel_dimensions: *voxel_dimensions,
+            voxel_map: vec![false; cast_usize(map_length)],
+        }
+    }
+
+    /// Calculate a voxel cloud from an existing mesh.
+    pub fn from_mesh(mesh: &Mesh, voxel_dimensions: &Vector3<f32>) -> Self {
```
yanchith, line 60: This is a perfect candidate for a snapshot test. We can use one of our primitives. You will need to add this derive to be able to serialize the voxel cloud: `serde::Serialize`.

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,456 @@
...
+        let block_start = Point3::new(
+            min_x_index as i32 - 1,
+            min_y_index as i32 - 1,
+            min_z_index as i32 - 1,
+        );
+        let block_dimensions = Vector3::new(
+            (max_x_index - min_x_index + 2.0) as u32,
+            (max_y_index - min_y_index + 2.0) as u32,
+            (max_z_index - min_z_index + 2.0) as u32,
+        );
```
yanchith: Why are we widening the block here? `grow` already widens it for itself, and `floor`/`ceil` should ensure we don't need the padding.

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,456 @@
...
+
+        // Going to populate the mesh with points as dense as the smallest voxel dimension.
+        let shortest_voxel_dimension = voxel_dimensions
+            .x
+            .min(voxel_dimensions.y.min(voxel_dimensions.z));
+
+        let mut voxel_cloud = VoxelCloud::new(&block_start, &block_dimensions, voxel_dimensions);
+
+        // Iterate through mesh faces.
```
yanchith: In general, in inline code comments, we should document the *why*, not the *what*. This particular comment is not very useful. For comparison, the previous comment "Going to populate the mesh with ..." is very useful.

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,456 @@
...
+        );
+
+        // Iterate through the existing voxel cloud.
+        for z in 0..self.block_dimensions.z {
+            for y in 0..self.block_dimensions.y {
+                for x in 0..self.block_dimensions.x {
+                    let voxel_coords = Point3::new(cast_i32(x), cast_i32(y), cast_i32(z));
+                    let voxel_state = self.voxel_at_relative_coords(&voxel_coords);
+                    // if the voxel is on
+                    if let Some(true) = voxel_state {
```
yanchith: Hmm, shouldn't we `expect` this, technically?

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,456 @@
...
+        voxel_cloud
+    }
+
+    /// Gets the state of a voxel defined in voxel coordinates relative to the
+    /// voxel block start.
+    pub fn voxel_at_relative_coords(&self, relative_coords: &Point3<i32>) -> Option<bool> {
+        match self.relative_three_dimensional_coordinate_to_one_dimensional(relative_coords) {
+            Some(index) => Some(self.voxel_map[index]),
+            None => None,
+        }
```
yanchith: Since you are functional programming geek :smile: , you will like that this can be simplified to:

```rust
self.relative_three_dimensional_coordinate_to_one_dimensional(relative_coords)
    .map(|index| self.voxel_map[index])
```

You don't need to do it if you don't want to though. Applies to similar cases in other getters.

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,456 @@
...
+    /// space coordinates.
+    #[allow(dead_code)]
+    pub fn get_cartesian(&self, point: &Point3<f32>) -> Option<bool> {
+        let voxel_coords = self.cartesian_to_absolute_voxel_coords(point);
+        self.voxel_at_absolute_coords(&voxel_coords)
+    }
+
+    /// Sets the state of a voxel defined in voxel coordinates relative to the
+    /// voxel block start.
+    pub fn set_voxel_at_relative_coords(&mut self, state: bool, relative_coords: &Point3<i32>) {
```
yanchith: Only noticed this now, but setters usually take key as the first parameter, and value as the second. (e.g. `HashMap`)

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,456 @@
...
+            .min(self.voxel_dimensions.y.min(self.voxel_dimensions.z));
+        // and weld naked edges
+        tools::weld(&joined_voxel_mesh, (min_voxel_dimension as f32) / 4.0)
+    }
+
+    /// Computes an index to the linear representation of the voxel block from
+    /// voxel coordinates relative to the voxel space block start.
+    ///
+    /// The function doesn't clamp the result because it may be used also to
+    /// check if the coordinates are out of bounds.
```
yanchith: Thank you for implementing this change, but we also need to update the comment. Same for absolute below.

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,456 @@
...
+    fn relative_three_dimensional_coordinate_to_one_dimensional(
+        &self,
+        relative_coords: &Point3<i32>,
```
yanchith, line 377: Hmm, wouldn't it make more sense if relative coords were unsigned? A negative relative coordinate is always out of bounds.

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,456 @@
...
+    #[allow(dead_code)]
+    fn cartesian_to_relative_voxel_coords(&self, point: &Point3<f32>) -> Point3<i32> {
+        Point3::new(
+            (point.x / self.voxel_dimensions.x).round() as i32 - self.block_start.x,
+            (point.y / self.voxel_dimensions.y).round() as i32 - self.block_start.y,
+            (point.z / self.voxel_dimensions.z).round() as i32 - self.block_start.z,
+        )
+    }
+
+    /// Calculates the center of a voxel in model-space coordinates from
```
yanchith, line 428: Where is the center of a voxel in world space? If we have a voxel space with voxel dimensions `[1.0, 1.0, 1.0]` and block start `[0, 0, 0]`, where is the center of a voxel for relative (and in this case absolute too) coords `[0, 0, 0]`? I think your current implementation defines it to be at `[0.0, 0.0, 0.0]`, but another potentially valid approach would be `[0.5, 0.5, 0.5]` - depending on what voxel space coords `[0, 0, 0]` mean when transformed to world space (center vs corner). Have you given some thought which of these would be better suited for us?

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,456 @@
...
+            },
+        ];
+
+        // Iterate through the voxel cloud
+        for z in 0..self.block_dimensions.z {
+            for y in 0..self.block_dimensions.y {
+                for x in 0..self.block_dimensions.x {
+                    // If the current voxel is on
+                    let voxel_coords = Point3::new(cast_i32(x), cast_i32(y), cast_i32(z));
+                    if let Some(true) = self.voxel_at_relative_coords(&voxel_coords) {
```
yanchith: Don't we want to `expect` this?

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,456 @@
...
+                        // calculate the position of its center in model space coordinates
+                        let voxel_center = self.relative_voxel_to_cartesian_coords(&voxel_coords);
+                        // top
```
yanchith: Probably misplaced comment?

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,456 @@
...
+                                Some(false) | None => {
+                                    // add a horizontal rectangle
+                                    plane_meshes.push(primitive::create_mesh_plane(
+                                        Plane::from_origin_and_plane(
+                                            // above the voxel center half way the height of the voxel
+                                            &(voxel_center + helper.direction_to_wall),
+                                            // align it properly
+                                            &helper.plane,
+                                        ),
+                                        // and set_voxel_at_absolute_coords its size to match the dimensions
```
yanchith: Another mass text replace victim.

yanchith (changes requested):
Thank you for incorporating the changes! :heart: 

I was now able to finish the review. I do have a second batch of various questions and suggestions. The questions are just that - don't view them as requests, I am merely confirming my understanding (and sometimes maybe discovering an error in the process :cat: ).

Overall I think we are very close!

src/mesh/voxel_cloud.rs
```diff
@@ -0,0 +1,456 @@
...
+        self.set_voxel_at_absolute_coords(state, &voxel_coords)
+    }
+
+    /// Calculates a simple triangulated welded mesh from the current state of
+    /// the voxel cloud.
+    ///
+    /// For watertight meshes this creates both, outer and inner boundary mesh.
+    /// There is also a high risk of generating a non-manifold mesh if some
+    /// voxels touch only diagonally.
+    pub fn to_mesh(&self) -> Option<Mesh> {
```
yanchith, line 235: This is also a very good candidate for a snapshot test, don't you think?

src/mesh/voxel_cloud.rs
janper: do you think it's worth it? i find this way too straight forward. I'd rather put some effort into testing more risky parts

src/mesh/voxel_cloud.rs
janper: yes, exactly like this

src/mesh/voxel_cloud.rs
janper: better safe than sorry. but you are right, `floor` and `ceil` should suffice.

src/mesh/voxel_cloud.rs
janper: is there a special implementation of `map` for `Option`?

src/mesh/voxel_cloud.rs
janper: exactly. that's how i check if the voxel is on the boundary of the voxel space in `to_mesh`:

```
        // if there isn't or if the current voxel is on the
                                // boundary of the voxel space block
                                Some(false) | None => {
```

src/mesh/voxel_cloud.rs
janper: In our case it's the centers that matter, so voxel's `[0, 0, 0]` center is at `[0.0, 0.0, 0.0]`. The voxel materialises symmetrically around its center. I find this more consistent than shifting the voxel space coordinate system half-a-voxel size to align the materialised voxel's boundary with the world-space verges. It simply feels better both mathematically and semantically and that's how i understand pictures like this:
image

src/mesh/voxel_cloud.rs
yanchith: Yes. `Option` has many combinators. They are hardcoded (i.e. not generic) in rust, because unlike haskell, rust doesn't have higher-kinded types and monads, but they are present.

https://doc.rust-lang.org/std/option/enum.Option.html#method.map

src/mesh/voxel_cloud.rs
janper: ~actually, it panicks without a positive padding and makes holes without a negative one. perhaps i should investigate more but i suspect the rounding in cartesian to voxel coords calculations might put a model-space coord out of voxel block bounds. that explains just the panic, though.~

janper: catched both bugs:

1. when converting the voxel coordinates into boolean vector index, it's not enough to check if the resulting index is within bounds but rather it's necessary to check each voxel XYZ to be within theyr respecitve bounds
2. the block dimensions are `max_index + 1`, because the index is always 1 lower than the length

src/mesh/voxel_cloud.rs
janper: after struggling with failing tests I realized the snapshot testing won't work because this function uses `weld` and that uses `HashMap`, therefore the result is always in a different order :/

janper: I also reviewed the weld function and for now I don't see a solution without the `HashMap` except implementing some sort of a map manually - which doesn't seem right

janper:
Ready for another review. There are some zero normal vectors that make the result look awful but i made a separate task for that. I suspect the `weld` function or even smooth normals calculation to cause it.

src/mesh/voxel_cloud.rs
yanchith: Okay, let's skip it for now.

src/mesh/voxel_cloud.rs
yanchith: Good job! Yeah, I totally forgot that indexing at `[3, 3, 3]` in a voxel space `[2, 2, 10]` ends up within bounds of the array, but not the voxel space.

yanchith (approved):
Nice!

PR #102, opened 2019-12-16
Make Face to face topology calculation linear, optimize vertex to vertex

janper:


src/mesh/topology.rs
```diff
@@ -1,16 +1,9 @@
...
 
-use crate::convert::cast_u32;
+use crate::convert::{cast_u32, cast_usize};
 
 use super::{Face, Mesh};
 
-// FIXME: @Optimization Explore whether the topologies wouldn't be
-// better served by being backed by a `Vec` instead of
-// `HashMap`. Ideally, we'd also create a wrapper struct that casts
-// the indices to/from u32 as necessary.
```
yanchith, line 12: This second part about wrapper structs of the fixme is not yet done. Let's keep it.

src/mesh/topology.rs
```diff
@@ -19,43 +12,74 @@ use super::{Face, Mesh};
 /// contain before it spills into heap. Implementation detail.
 const MAX_INLINE_NEIGHBOR_COUNT: usize = 8;
 
+pub type VertexToFaceRelation = SmallVec<[u32; MAX_INLINE_NEIGHBOR_COUNT]>;
```
yanchith: Nit: would be good if the comments and naming here were consistent as with the other topologies. This one is called calculate, the others are called compute. Also the comments are different, ideally let's stick to a single comment style for these things (just see the difference in comments between this and f2f, for example).

src/mesh/topology.rs
```diff
@@ -19,43 +12,74 @@ use super::{Face, Mesh};
...
+
+/// Calculates topological relations (neighborhood) of mesh vertex -> faces.
+/// Returns a Map (key: vertex index, value: list of its neighboring faces indices)
+pub fn calculate_vertex_to_face_topology(mesh: &Mesh) -> Vec<VertexToFaceRelation> {
+    calculate_vertex_to_face_topology_from_components(mesh.faces(), cast_u32(mesh.vertices().len()))
+}
+
+pub fn calculate_vertex_to_face_topology_from_components(
```
yanchith: Hmm, since this split is trivial to make, let's make it once we need it. It's not used anywhere else in this PR, so I would prefer if we only had just one function - the one that takes the whole mesh. Do you have some concrete plans with this?

src/mesh/topology.rs
```diff
@@ -19,43 +12,74 @@ use super::{Face, Mesh};
...
-                                || triangle_face.contains_unoriented_edge(e3);
-                            if face_contains_edge
-                                && from_face_index != to_face_index
-                                && !neighbors.contains(&to_face_index_u32)
-                            {
-                                neighbors.push(to_face_index_u32);
-                            }
-                        }
+                let vertices = &triangle_face.vertices;
+                for first_vertex_in_face in &v2f[cast_usize(vertices.0)] {
```
yanchith: Isn't this a face index? The current name suggests it is a vertex index. Hmm, maybe something like `face_for_first_vertex` would be more clear. Also below for the face for second vertex.

Also, I am not sure we want to rely on the optimizer here, so if possible, let's access the vectors only once and use those values below.

```rust
let faces_for_1st_vertex = &v2f[cast_usize(vertices.0)];
let faces_for_2nd_vertex = &v2f[cast_usize(vertices.1)];
let faces_for_3rd_vertex = &v2f[cast_usize(vertices.2)];
```

src/mesh/topology.rs
```diff
@@ -64,27 +88,25 @@ pub fn compute_face_to_face_topology(mesh: &Mesh) -> FaceToFaceTopology {
...
                 let vertex_indices = &[f.vertices.0, f.vertices.1, f.vertices.2];
                 for i in 0..vertex_indices.len() {
-                    let neighbors = v2v.entry(vertex_indices[i]).or_insert_with(SmallVec::new);
-
                     let neighbor_candidate1 = vertex_indices[(i + 1) % 3];
                     let neighbor_candidate2 = vertex_indices[(i + 2) % 3];
 
-                    if !neighbors.contains(&neighbor_candidate1) {
-                        neighbors.push(neighbor_candidate1)
+                    if !v2v[cast_usize(vertex_indices[i])].contains(&neighbor_candidate1) {
```
yanchith: Similar to above, I am not sure the optimizer can see this, so I'd extract `v2v[cast_usize(vertex_indices[i])]` to a variable.

src/mesh/topology.rs
```diff
@@ -19,43 +12,74 @@ use super::{Face, Mesh};
...
+    }
+
+    v2f
+}
+
 /// A topology containing neighborhood relations between faces. Two
 /// faces are neighbors if and only if they they share an unoriented
 /// edge.
-pub type FaceToFaceTopology = HashMap<u32, SmallVec<[u32; MAX_INLINE_NEIGHBOR_COUNT]>>;
+pub type FaceToFaceRelation = SmallVec<[u32; MAX_INLINE_NEIGHBOR_COUNT]>;
```
yanchith: This change invalidated the comment above it. Let's keep the code and comments in sync - either revert the change or update the comment. I have a slight preference to keeping the type definition as it was (with the exception of changing the hashmap to a vector). Also for `VertexToVertexTopology` below, and ideally also for `VertexToFaceRelation` above.

The reason for leaving the types as they were is because they were placeholders for the actual wrapper structs mentioned in the above fixme you removed. All we will have to do when we implement the fixme is substitute the type alias for the wrapper. Ideally very little caller call would have to be modified (the exception being if the caller did not yet use the type alias), but the topology would become more typesafe.

src/mesh/topology.rs
```diff
@@ -19,43 +12,74 @@ use super::{Face, Mesh};
...
 
 /// Computes face to face topology for mesh.
-pub fn compute_face_to_face_topology(mesh: &Mesh) -> FaceToFaceTopology {
-    let mut f2f: HashMap<u32, SmallVec<[u32; 8]>> = HashMap::new();
+pub fn compute_face_to_face_topology(
+    mesh: &Mesh,
+    v2f: &[VertexToFaceRelation],
+) -> Vec<FaceToFaceRelation> {
+    let mut f2f: Vec<SmallVec<[u32; 8]>> = vec![SmallVec::new(); mesh.faces().len()];
```
yanchith, line 68: It is really cool that we can do this!

yanchith (commented):
Very nice PR :heart:  The logic seems sound!

I have a couple of comments regarding very important things, such as consistency in comments and naming (:joy:) , and also perhaps about relying on the optimizer a bit too much. See inline.

src/mesh/topology.rs
janper: I already have another PR ready using this.

src/mesh/topology.rs
janper: Yes I agree and I originally had it like that but Clippy made me do it this way. The reason is you can't use a Vector as a function parameter - see `loop_subdivision` or `laplacian_smoothing`.

yanchith: Hmm. In this case I would define the type aliases as they were and not use them as function parameters yet (so clippy is satisfied). Alternatively I would add clippy allow attributes with fixmes to remove them, but that seems more complex.

src/mesh/topology.rs
```diff
@@ -85,12 +101,14 @@ pub fn compute_face_to_face_topology(
...
-/// an edge.
+/// Neighborhood / connection relations between vertices. Two vertices are
+/// neighbors if and only if they they are end points of an edge.
 pub type VertexToVertexRelation = SmallVec<[u32; MAX_INLINE_NEIGHBOR_COUNT]>;
 
-/// Computes vertex to vertex topology for mesh.
+/// Computes topological relations (connections) of mesh vertex -> vertices.
+///
+/// Output: The index represents a vertex index, the value is a list of vertices
+/// connected to the respective face with an edge.
```
yanchith: "respective face" -> "respective vertex"

src/mesh/topology.rs
janper: > In this case I would define the type aliases as they were and not use them as function parameters yet

Why is that a good solution? It doesn't give the caller (or linter) any hint, so why not keeping them unwrapped?
Also why is it better than the status quo?

yanchith: I am trying to figure out a way of this working without creating the full-blown wrappers right now. Yes, as you are saying, as a 3rd alternative, we can also merge the comments on the type aliases with the functions that create them and delete the type aliases for now. They will come back as the wrappers at some point.

janper: going for the 3rd option. the comments are already there

yanchith (approved):

PR #103, opened 2019-12-18
Fix weld generating zero vector normals in some cases

janper:
The weld function calculates normals as an average of all normals associated to a respective vertex (a vertex can be referenced by multiple faces, each of them may couple it with a different normal). The original weld function deduplicated associated normals, which was not correct. 
Some optimizations were made too and the tests were adjusted to the current algorithm.

### DIFF
--- a/src/mesh/analysis.rs
+++ b/src/mesh/analysis.rs
@@ -331,9 +331,11 @@ pub fn triangulated_mesh_genus(vertex_count: usize, edge_count: usize, face_coun
 
 /// Checks if two meshes are similar.
 ///
-/// Two mesh geometries are similar when they are visually similar
-/// (see the definition of `are_visually_similar`), and they have the
-/// same number of vertices and normals.
+/// Two mesh geometries are similar when they are visually similar (see the
+/// definition of `are_visually_similar`), and they have the same number of
+/// vertices and normals. Therefore they are going to be treated the same by all
+/// functions of this software and all their transformations result in similar
+/// mesh geometries.
 #[allow(dead_code)]
 pub fn are_similar(mesh1: &Mesh, mesh2: &Mesh) -> bool {
     mesh1.vertices().len() == mesh2.vertices().len()
@@ -357,10 +359,11 @@ pub fn are_similar(mesh1: &Mesh, mesh2: &Mesh) -> bool {
 /// normals are identical, because one mesh may reuse (share) vertices and
 /// normals in more faces and the other doesn't (applies to all or some faces).
 ///
-/// They mesh geometries are not necessarily identical in memory but they look
+/// The mesh geometries are not necessarily identical in memory but they look
 /// the same. If the number of vertices differs, the mesh geometries don't share
 /// the same qualities (they are not welded in the same places and at least one
-/// of them is not watertight) are not going to be treated the same by some
+/// of them is not watertight). Despite of that, they are considered to be
+/// visually similar but they are not going to be treated the same by some
 /// functions of this software and all their transformations result in different
 /// mesh geometries.
 #[allow(dead_code)]
--- a/src/mesh/analysis.rs
+++ b/src/mesh/analysis.rs
@@ -362,10 +362,9 @@ pub fn are_similar(mesh1: &Mesh, mesh2: &Mesh) -> bool {
 /// The mesh geometries are not necessarily identical in memory but they look
 /// the same. If the number of vertices differs, the mesh geometries don't share
 /// the same qualities (they are not welded in the same places and at least one
-/// of them is not watertight). Despite of that, they are considered to be
-/// visually similar but they are not going to be treated the same by some
-/// functions of this software and all their transformations result in different
-/// mesh geometries.
+/// of them is not watertight). Despite that they are considered visually
+/// similar, they are not going to be treated the same by some functions of this
+/// software and all their transformations result in different mesh geometries.
 #[allow(dead_code)]
 pub fn are_visually_similar(mesh1: &Mesh, mesh2: &Mesh) -> bool {
     struct UnpackedFace {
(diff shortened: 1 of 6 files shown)


src/mesh/analysis.rs
```diff
@@ -357,10 +359,11 @@ pub fn are_similar(mesh1: &Mesh, mesh2: &Mesh) -> bool {
 /// normals are identical, because one mesh may reuse (share) vertices and
 /// normals in more faces and the other doesn't (applies to all or some faces).
 ///
-/// They mesh geometries are not necessarily identical in memory but they look
+/// The mesh geometries are not necessarily identical in memory but they look
 /// the same. If the number of vertices differs, the mesh geometries don't share
 /// the same qualities (they are not welded in the same places and at least one
-/// of them is not watertight) are not going to be treated the same by some
+/// of them is not watertight). Despite of that, they are considered to be
```
yanchith: Nit: "Despite that they are considered visually similar, they are not going to be treated the same..."

src/mesh/tools.rs
```diff
@@ -534,38 +531,38 @@ mod tests {
         Mesh::from_triangle_faces_with_vertices_and_normals(faces, vertices, vertex_normals)
     }
 
-    pub fn open_box_sharp_mesh(position: [f32; 3], scale: f32) -> Mesh {
+    pub fn open_box_sharp_mesh() -> Mesh {
         let vertex_positions = vec![
             // back
-            v(-1.0, 1.0, -1.0, position, scale), //0
```
yanchith, line 540: I believe you can now delete the `v` function as this was it's last use. You can check for this yourself by running `cargo check --tests`. (`cargo clippy --tests` works also, but there are so many errors there that you probably don't want to see it... or maybe you should :smile: )

src/mesh/tools.rs
```diff
@@ -625,54 +622,45 @@ mod tests {
...
+            TriangleFace::new_separate(5, 7, 6, 5, 7, 6),
+            TriangleFace::new_separate(1, 5, 6, 1, 5, 6),
+            TriangleFace::new_separate(1, 6, 2, 1, 6, 2),
+            TriangleFace::new_separate(0, 2, 6, 0, 2, 6),
+            TriangleFace::new_separate(3, 7, 4, 3, 7, 4),
+            TriangleFace::new_separate(4, 7, 5, 4, 7, 5),
+            TriangleFace::new_separate(0, 6, 3, 0, 6, 3),
+            TriangleFace::new_separate(3, 6, 7, 3, 6, 7),
+            TriangleFace::new_separate(1, 4, 5, 1, 4, 5),
+            TriangleFace::new_separate(1, 2, 4, 1, 2, 4),
```
yanchith: Since indices for vertices and normals are exactly the same, you can use `TriangleFace::new` here. I see you always pick `new_separate` even if not necessary. Is that just habit or do you have a reason for that? We could remove `new` and rename `new_separate` to `new`.

src/mesh/tools.rs
```diff
@@ -625,54 +622,45 @@ mod tests {
...
-            n(-1.0 / 3.0, 1.0 / 3.0, 1.0 / 3.0),
-            n(1.0 / 3.0, 1.0 / 3.0, 1.0 / 3.0),
-            n(1.0 / 3.0, 1.0 / 3.0, -1.0 / 3.0),
-            // front
-            n(-1.0 / 3.0, -1.0 / 3.0, -1.0 / 3.0),
-            n(1.0 / 3.0, -1.0 / 3.0, -1.0 / 3.0),
-            n(1.0 / 3.0, -1.0 / 3.0, 1.0 / 3.0),
-            n(-1.0 / 3.0, -1.0 / 3.0, 1.0 / 3.0),
+        let normals = vec![
+            n(0.25, 0.5, 0.25),
```
yanchith: Since you are already doing cleanup, could you please remove the `n` function in this file? I think we can just use `Vector3::new`

yanchith (commented):
Nice!

Left a couple of inline comments.

Also: Does weld have more information for computing normals than a mesh constructor would have? I am asking whether we shouldn't just use the smooth normals from the other PR and simplify weld's logic by allowing it to not care about normals at all.

janper:
> Also: Does weld have more information for computing normals than a mesh constructor would have? I am asking whether we shouldn't just use the smooth normals from the other PR and simplify weld's logic by allowing it to not care about normals at all.

Actually, the way weld calculates normals is different from the `NormalStrategy::Smooth`. 
The strategy ignores the existing normals (if they exist) and instead first calculates triangle normals for all the faces containing the vertex and the calculating their average. 
Weld, on the other hand calculates an average normal bound to the original vertices that become welded into one. The original normals don't need to be perpendicular to the faces and there are many different scenarios that can happen - for example just one of many faces is not welded.

The question is whether we appreciate this way of calculating the welded normals. I can imagine some exotic use cases, when the user intentionally messed up normals and wants to keep them that way but tbh I've never seen such an approach before. If that is not useful, then we can simply ask which strategy should be used and calculate the normals anew.

src/mesh/analysis.rs
janper: ok

src/mesh/tools.rs
janper: good!
so should I see the clippy or not?

src/mesh/tools.rs
janper: I don't know. It just seems more correct. The case for current `new` seems to be very specific and is a subset of the `new_separate`. If we should make a change, I suggest keeping both constructors but making the explicit one default?
And yes, I should have used the other one here.

src/mesh/tools.rs
janper: ok

src/mesh/tools.rs
yanchith: Well you probably should see it, but don't fix any of the warnings/errors yet. We will probably only fix the clippy errors at some point. Having clippy warnings in tests is fine.

src/mesh/tools.rs
yanchith: I am ok also with what you are suggesting. I would also be ok with just one constructor. Up to you.

yanchith:
> Actually, the way weld calculates normals is different from the NormalStrategy::Smooth.

kk, I thought so. In this case, let's keep the logic, since it's already done. We can revisit the question later. Originally I thought we were going to preserve as much of original data as possible everywhere, but it sure seems like lot of work, so I am happy we usually don't.

yanchith (approved):

PR #104, opened 2019-12-18
Implement smooth normal calculation strategy

janper:
Implement smooth normal calculation strategy and use it where it belongs.


src/geometry.rs
```diff
@@ -7,7 +7,7 @@ pub fn compute_triangle_normal(
     p3: &Point3<f32>,
 ) -> Vector3<f32> {
     let u = p2 - p1;
-    let v = p3 - p1;
+    let v = p3 - p2;
```
yanchith: Are you sure about this? Did you discover something wrong with the original? I am asking because it changed a snapshot, but the difference is within floating point precision errors, so it seems like just noise.

src/mesh/smoothing.rs
```diff
@@ -74,12 +74,13 @@ pub fn laplacian_smoothing(
         }
     }
 
-    // FIXME: Calculate smooth normals for the result once we support them
     (
-        Mesh::from_faces_with_vertices_and_normals(
-            mesh.faces().to_vec(),
+        Mesh::from_triangle_faces_with_vertices_and_computed_normals(
```
yanchith: Instead of doing this, don't we want to add `Mesh::from_faces_with_vertices_and_computed_normals()`? (Note the absent word `triangle`). This is true especially if we are going to use this often. `from_triangle_faces_with_vertices_and_computed_normals` should be able to just delegate to that by adapting it's face iterator with `.map(Face::from)`.

src/mesh/smoothing.rs
```diff
@@ -550,15 +551,53 @@ mod tests {
...
-        assert_eq!(relaxed_mesh_0.faces().len(), mesh.faces().len(),);
-        assert_eq!(relaxed_mesh_1.faces().len(), mesh.faces().len(),);
-        assert_eq!(relaxed_mesh_10.faces().len(), mesh.faces().len(),);
-        assert_eq!(relaxed_mesh_0.vertices().len(), mesh.vertices().len(),);
-        assert_eq!(relaxed_mesh_1.vertices().len(), mesh.vertices().len(),);
-        assert_eq!(relaxed_mesh_10.vertices().len(), mesh.vertices().len(),);
-        assert_eq!(relaxed_mesh_0.normals().len(), mesh.normals().len());
-        assert_eq!(relaxed_mesh_1.normals().len(), mesh.normals().len());
-        assert_eq!(relaxed_mesh_10.normals().len(), mesh.normals().len(),);
+        assert_eq!(
```
yanchith: I think the assert conditions in this test did not really change. Let's revert to the original for smaller diff.

src/mesh/smoothing.rs
```diff
@@ -74,12 +74,13 @@ pub fn laplacian_smoothing(
         }
     }
 
-    // FIXME: Calculate smooth normals for the result once we support them
```
yanchith, line 77: Do we also want to do `loop_subdivision` in this PR? (no is also an answer)

yanchith, line 77: Isn't the correct thing to do to tell the function what kind of normals we want it to compute? Or are smooth normals always desired and sharp normals never?

We don't need to expose this choice to the UI, but having it would also allow you to not modify existing tests and snapshots, but only add a couple of ones for new behavior.

src/mesh/snapshots/tests__uv_sphere_2_3_after_1_iteration_of_loop_subdivision.snap
```diff
@@ -1,6 +1,6 @@
 ---
-source: src/mesh_smoothing.rs
```
yanchith, line 2: This snapshot must have changed becuase the formula for sharp normals changed slightly. Could you explain why the change is necessary? Otherwise I would revert the change - the values seem almost the same, so I suspect the difference is only caused by different arithmetic order and floating point precision.

src/mesh/mod.rs
```diff
@@ -114,23 +110,53 @@ impl Mesh {
                                 &vertices_collection[cast_usize(v3)],
                             );
 
-                            normals_collection.push(face_normal);
-
+                            normals_collection_sharp.push(face_normal);
                         }
-                        // FIXME: once we add other kinds of faces, they must panic here
-                        // _ => panic!("Face must be a triangle, we just created it"),
```
yanchith, line 114: Let's not remove this fixme. It is still valid. We are in a constructor that creates a mesh from triangles - any other kinds of faces are really programmer error here.

src/mesh/mod.rs
```diff
@@ -114,23 +110,53 @@ impl Mesh {
...
+                assert_eq!(
+                    normals_collection_sharp.capacity(),
+                    faces_collection_sharp.len()
+                );
+                (faces_collection_sharp, normals_collection_sharp)
+            }
+            NormalStrategy::Smooth => {
+                let faces_collection_smooth: Vec<_> = faces
+                    .into_iter()
+                    .map(|(i1, i2, i3)| TriangleFace::new_separate(i1, i2, i3, i1, i2, i3))
```
yanchith: You don't need `TriangleFace::new_separate` here. `new` is enough, because the indices are shared.

src/mesh/mod.rs
```diff
@@ -114,23 +110,53 @@ impl Mesh {
...
                         }
-                        // FIXME: once we add other kinds of faces, they must panic here
-                        // _ => panic!("Face must be a triangle, we just created it"),
                     }
                 }
+                assert_eq!(normals_collection_sharp.len(), faces_collection_sharp.len());
+                assert_eq!(
+                    normals_collection_sharp.capacity(),
+                    faces_collection_sharp.len()
+                );
```
yanchith, line 143: :heart:

src/mesh/mod.rs
```diff
@@ -796,6 +822,35 @@ fn remove_orphan_vertices_and_normals(
...
+        let shared_faces_count_f32 = shared_face_indices.len() as f32;
+        let mut normal: Vector3<f32> = Vector3::zeros();
+        for face_index in shared_face_indices {
+            match faces[cast_usize(*face_index)] {
+                Face::Triangle(face) => {
+                    normal += geometry::compute_triangle_normal(
+                        &vertices[cast_usize(face.vertices.0)],
+                        &vertices[cast_usize(face.vertices.1)],
+                        &vertices[cast_usize(face.vertices.2)],
+                    ) / shared_faces_count_f32;
```
yanchith: This will end up being normalized, right?

yanchith (commented):
I really like the implementation! Left a couple of followup questions though.

src/geometry.rs
janper: My reason was that I wasn't sure about the previous implementation. The current version respects the face winding, whereas the previous one calculated a cross product from a correct half-edge (vector) and a reverted one.

src/mesh/smoothing.rs
janper: yes, let's do it that way

src/mesh/smoothing.rs
janper: The last two did change. I added the message because the test kept failing and it didn't say where. This way I knew.

src/mesh/smoothing.rs
janper: Hmmm, yes, let's

janper: I need to have a better look how deeply integrated is the normal calculation into the functions. If it is easy, I'm in for giving the choice. We can also expose it in the UI (of we already support boolean inputs).

src/mesh/snapshots/tests__uv_sphere_2_3_after_1_iteration_of_loop_subdivision.snap
janper: See my comment above. If you want, I can search for some reference and see how to calculate the triangle normal properly.

src/mesh/mod.rs
janper: ok

src/mesh/mod.rs
janper: ok

src/mesh/mod.rs
janper: ?
just a copy-paste

src/mesh/mod.rs
janper: Now this is the crucial question! I've been thinking the same when I realized we never guarantee the mesh normals are normalized! I believe we should normalize all the normals and I think the best place would be the mesh constructor. 
The reason for normalizing are comparison methods and welding (see the other PR).

src/mesh/mod.rs
```diff
@@ -796,6 +822,35 @@ fn remove_orphan_vertices_and_normals(
...
+
+    let mut normals: Vec<Vector3<f32>> = Vec::with_capacity(vertex_count);
+
+    for shared_face_indices in vertex_to_face_topology {
+        let shared_faces_count_f32 = shared_face_indices.len() as f32;
+        let mut normal: Vector3<f32> = Vector3::zeros();
+        for face_index in shared_face_indices {
+            match faces[cast_usize(*face_index)] {
+                Face::Triangle(face) => {
+                    normal += geometry::compute_triangle_normal(
```
janper, line 838: I'm thinking now it would be better to calculate all face normals upfront and then use the lookup table for these calculations. That way we reduce calls of this function to 1/3. Which is worse, memory allocation or heavier computation?

src/mesh/mod.rs
yanchith: Really? I thought the assert wasn't present previously, but it indeed is good that we now have it. Or maybe I just didn't see it.

src/mesh/mod.rs
yanchith: Depends. If the calculation is linear (it is in this case), I would say memory allocation is worse. If the calculation is quadratic and memory allocation helps, I tend to prefer allocation unless I can somehow prove the data is very small.

src/geometry.rs
yanchith: From what I understand about cross product (which is not much, arguably), this change shouldn't make a difference in practice. Both angles between `p1->p2` and `p1->p3` (the original), and `p1->p2` and `p2->p3` (the introduced change) will never be more than 180 degrees, meaning they should produce the same cross product vector.

src/mesh/smoothing.rs
janper: ~~we do a lot of calculation inside `from_triangle_faces_with_vertices_and_computed_normals`. it works because we're not inputting an actual face but merely three numbers. the rest is being calculated.
if we make a generic constructor, we need to make some of decisions, like where should we calculate sharp/smooth normals, what to do with existing normal indices inside the faces?
for now i prefer not doing anything. let's keep it as it is and revisit the issue when we support non-triangular faces.~~

src/mesh/mod.rs
janper: i decided to normalize the normals in mesh constructors

src/mesh/smoothing.rs
yanchith: Oh I see the conundrum now :slightly_frowning_face:

Ok. What you did is good enough for now, but we might have to change the structure of our data a bit, because this is ugly. I'll create a trello task for this.

src/mesh/smoothing.rs
```diff
@@ -826,12 +830,13 @@ mod tests {
...
             2,
             3,
+            NormalStrategy::Sharp,
         );
         let v2v = topology::compute_vertex_to_vertex_topology(&mesh);
         let v2f = topology::compute_vertex_to_face_topology(&mesh);
         let f2f = topology::compute_face_to_face_topology(&mesh, &v2f);
 
-        let subdivided_mesh = loop_subdivision(&mesh, &v2v, &f2f)
+        let subdivided_mesh = loop_subdivision(&mesh, &v2v, &f2f, NormalStrategy::Smooth)
```
yanchith: The reason I asked you to parametrize `laplacian_smoothing` and `loop_subdivision` with normal strategy is so that we don't have to modify the tests and snapshots by still using the `NormalStrategy::Sharp` here.

src/mesh/tools.rs
```diff
@@ -815,13 +816,20 @@ mod tests {
         let sphere_with_synced_winding =
             synchronize_mesh_winding(&sphere_with_faces_one_flipped, &f2f);
 
+        println!("original {}", &sphere);
+        println!(
+            "sphere_with_faces_one_flipped {}",
+            &sphere_with_faces_one_flipped
+        );
+        println!("sphere_with_synced_winding {}", &sphere_with_synced_winding);
```
yanchith: Forgot to remove these?

src/mesh/tools.rs
```diff
@@ -790,12 +790,13 @@ mod tests {
             Point3::origin(),
```
yanchith: If changes to this test happened only for debugging purposes, let's discard these changes.

yanchith (changes requested):
I believe the existing snapshots (or any other tests, for that matter) should not be modified as a result of this PR. That's the reason I wanted to parametrize existing functions with `NormalStrategy`.

However, feel free to add new snapshot or unit tests using the `Smooth` strategy.

janper:
> I believe the existing snapshots (or any other tests, for that matter) should not be modified as a result of this PR. That's the reason I wanted to parametrize existing functions with `NormalStrategy`.

I've returned most of the snapshots to their original state (as much as it was possible) but there was no normal strategy for laplacian smoothing before, so the snapshots had to change.

yanchith:
> > I believe the existing snapshots (or any other tests, for that matter) should not be modified as a result of this PR. That's the reason I wanted to parametrize existing functions with `NormalStrategy`.
> 
> I've returned most of the snapshots to their original state (as much as it was possible) but there was no normal strategy for laplacian smoothing before, so the snapshots had to change.

I see. Ok.

janper:
I don't get the current CI error. It's referring to something that is in the master, not here. Also the line numbers don't match.

yanchith:
Weird indeed. It seems it checked out the correct branch though. I'll investigate more tomorrow.

src/mesh/smoothing.rs
```diff
@@ -74,12 +75,11 @@ pub fn laplacian_smoothing(
         }
     }
 
-    // FIXME: Calculate smooth normals for the result once we support them
     (
-        Mesh::from_faces_with_vertices_and_normals(
+        Mesh::from_faces_with_vertices_and_computed_normals(
             mesh.faces().to_vec(),
```
yanchith: Is it possible to not allocate here? Try `mesh.faces().iter()` or `mesh.faces().iter().copied()` or `mesh.faces().iter().cloned()`.

src/mesh/smoothing.rs
```diff
@@ -534,31 +535,51 @@ mod tests {
...
+        assert_eq!(relaxed_mesh_10.faces().len(), mesh.faces().len());
+        assert_eq!(relaxed_mesh_0.vertices().len(), mesh.vertices().len());
+        assert_eq!(relaxed_mesh_1.vertices().len(), mesh.vertices().len());
+        assert_eq!(relaxed_mesh_10.vertices().len(), mesh.vertices().len());
+        // In this specific case nothing changes, therefore the smoothened mesh
+        // should be equal to the original mesh.
         assert_eq!(relaxed_mesh_0.normals().len(), mesh.normals().len());
-        assert_eq!(relaxed_mesh_1.normals().len(), mesh.normals().len());
-        assert_eq!(relaxed_mesh_10.normals().len(), mesh.normals().len(),);
+        assert_eq!(relaxed_mesh_1.normals().len(), mesh.vertices().len());
```
yanchith, line 584: I'd add a small comment here explaining that the `Smooth` strategy creates meshes with normal count equal to vertex count

yanchith (dismissed):
Nice!

yanchith (approved):

PR #105, opened 2019-12-21
Fix mesh winding for mesh from voxel cloud

janper:


src/mesh/voxel_cloud.rs
```diff
@@ -520,4 +520,34 @@ mod tests {
...
+    #[test]
+    fn test_voxel_cloud_single_voxel_to_mesh() {
+        let mut voxel_cloud = VoxelCloud::new(
+            &Point3::origin(),
+            &Vector3::new(1, 1, 1),
+            &Vector3::new(1.0, 1.0, 1.0),
+        );
+        voxel_cloud.set_voxel_at_relative_coords(&Point3::new(0, 0, 0), true);
+
+        let voxel_mesh = voxel_cloud.to_mesh().expect("Should be a mesh");
```
yanchith: `.unwrap()` is fine in tests

src/mesh/voxel_cloud.rs
```diff
@@ -520,4 +520,34 @@ mod tests {
...
+    #[test]
+    fn test_voxel_cloud_single_voxel() {
+        let mut voxel_cloud = VoxelCloud::new(
+            &Point3::origin(),
+            &Vector3::new(1, 1, 1),
+            &Vector3::new(1.0, 1.0, 1.0),
+        );
+        voxel_cloud.set_voxel_at_relative_coords(&Point3::new(0, 0, 0), true);
+
+        insta::assert_json_snapshot!("single_voxel", &voxel_cloud);
```
yanchith: What exactly is this testing? If it tests whether `set` affects subsequent `get`s (i.e. really `set`s the data), this should be a unit test testing the following:

1.) Create new voxel cloud
2.) Get a voxel, it should be `false`
2.) Set a voxel to `true`
3.) Get the same voxel, it should now be `true`

Snapshot tests are for writing tests that would be really impractical to write otherwise. This test, however, is as simple as they can be and the snapshot just adds a lot of semantic overhead.

I would also append `_get_set` or something similar to the test name so the reader knows what this test is supposed to do without reading the code and guessing.

src/mesh/voxel_cloud.rs
```diff
@@ -283,8 +283,8 @@ impl VoxelCloud {
                 //left
                 plane: Plane::new(
                     &Point3::origin(),
-                    &Vector3::new(0.0, 1.0, 0.0),
-                    &Vector3::new(0.0, 0.0, -1.0),
+                    &Vector3::new(0.0, -1.0, 0.0),
```
yanchith, line 286: So this before produced a normal facing in the other direction?

src/mesh/voxel_cloud.rs
```diff
@@ -520,4 +520,34 @@ mod tests {
...
+            &Vector3::new(1, 1, 1),
+            &Vector3::new(1.0, 1.0, 1.0),
+        );
+        voxel_cloud.set_voxel_at_relative_coords(&Point3::new(0, 0, 0), true);
+
+        insta::assert_json_snapshot!("single_voxel", &voxel_cloud);
+    }
+
+    #[test]
+    fn test_voxel_cloud_single_voxel_to_mesh() {
```
yanchith: Maybe append `_produces_synchronized_mesh`, if that's really what this is testing.

yanchith (commented):
See inline.

yanchith:
Haha, I just noticed `hunt-for-reverted-faces` :smile: Did you find Ricky Baker yet?

src/mesh/voxel_cloud.rs
janper: This particular test was originally intended to test `to_mesh` but that function is nor fit for snapshot testing. This is what remained and looking back, I think it's useless. Since we both did some work on it, I'll convert it t what you suggest.

src/mesh/voxel_cloud.rs
janper: yes, ~there were several~ I mixed-up several planes. the results now look ok and there is also a test

yanchith (approved):

PR #106, opened 2019-12-23
Voxel cloud resize

janper:
Please, have a preliminary look and let me know if you like it at all. Then I'll add some tests to make sure it works as it should. Only then I'll ask you to make a proper review.

This will be needed for boolean operations with voxel clouds and later on for voxel-based hybridization.


src/mesh/voxel_cloud.rs
```diff
@@ -445,6 +633,28 @@ impl VoxelCloud {
...
         )
     }
+
+    /// Converts absolute voxel coordinates into relative voxel coordinates
+    #[allow(dead_code)]
+    fn absolute_to_relative_voxel_coords(&self, absolute_coords: &Point3<i32>) -> Point3<i32> {
+        absolute_coords - self.block_start.coords
+    }
+
+    /// Converts absolute voxel coordinates into relative voxel coordinates
```
yanchith: Comment copy paste

src/mesh/voxel_cloud.rs
```diff
@@ -368,6 +368,194 @@ impl VoxelCloud {
...
+
+        let start_offset = Vector3::new(left_offset, front_offset, bottom_offset);
+        let resized_block_start = self.block_start - start_offset;
+        self.resize_from_start_and_dimensions(&resized_block_start, &resized_block_dimensions)
+    }
+
+    /// Resize the voxel cloud block to exactly fit the volumetric geometry.
+    /// Returns None for empty the voxel cloud.
+    #[allow(dead_code)]
+    pub fn shrink_to_fit(&self) -> Option<Self> {
```
yanchith: Just a thought, but aren't voxel cloud singularities (empty voxel clouds) valid? Does this have to return option? Of course, if we allow empty voxel clouds, `to_mesh` will have to return `Option`, but it already does that because of `weld`.

src/mesh/voxel_cloud.rs
```diff
@@ -368,6 +368,194 @@ impl VoxelCloud {
         tools::weld(&joined_voxel_mesh, (min_voxel_dimension as f32) / 4.0)
     }
 
+    /// Resize the voxel cloud block to match new block start and block dimensions.
+    ///
+    /// This clips the outstanding parts of the original voxel cloud.
+    #[allow(dead_code)]
+    pub fn resize_from_start_and_dimensions(
```
yanchith: Since this doesn't actually mutate the existing voxel cloud, I'd change the naming here a bit. I started to use `clone_with_` in other parts of the project for similar things. In this case: `clone_with_new_start_and_dimensions`. Same for the other methods.

Also see the contradictory comment below :joy:

yanchith: I know you really don't like mutation, but shouldn't all these methods be mutating the original? Is there other reasoning for these methods to clone? The one reason I can think about is that you would very often have to re-allocate the backing array anyway, but even this could be prevented in some cases. I think `Vec::shrink_to_fit` e.g. doesn't always have to reallocate. I'd like to hear your thoughts here.

src/mesh/voxel_cloud.rs
```diff
@@ -368,6 +368,194 @@ impl VoxelCloud {
...
+        right_offset: i32,
+        bottom_offset: i32,
+        left_offset: i32,
+        front_offset: i32,
+        rear_offset: i32,
+    ) -> Self {
+        let x_dimension = cast_i32(self.block_dimensions.x) + right_offset;
+        let y_dimension = cast_i32(self.block_dimensions.y) + rear_offset;
+        let z_dimension = cast_i32(self.block_dimensions.z) + top_offset;
+        assert!(
```
yanchith: Aren't zero dimensions actually ok in theory? (You would have to modify the `new` and `to_mesh` functions)

You could clamp cast these coordinates to u32 (see our convert module).

src/mesh/voxel_cloud.rs
```diff
@@ -368,6 +368,194 @@ impl VoxelCloud {
...
+        let resized_block_start = self.block_start - start_offset;
+        self.resize_from_start_and_dimensions(&resized_block_start, &resized_block_dimensions)
+    }
+
+    /// Resize the voxel cloud block to exactly fit the volumetric geometry.
+    /// Returns None for empty the voxel cloud.
+    #[allow(dead_code)]
+    pub fn shrink_to_fit(&self) -> Option<Self> {
+        // Scan left to right to find the first voxel
+        if let Some(left_offset) = (0..self.block_dimensions.x).find(|x| {
```
yanchith: I didn't read the following code carefully, but if I understand what it is trying to do, I'd do it differently - most notably, only iterate the whole voxel cloud once with linear complexity instead of multiple times with quadratic complexity.

What you are trying to do in this function is compute offsets to pass to `resize_from_offsets`. What if you only iterated over the boolean array, and for each iteration computed the voxel space coordinates if the current boolean is true? You would use this the same way you find a maximum/minimum element in a collection - start with a sentinel value of offsets (maximum and minimum points) and slowly tighten the bounds as you discover new voxels.

Once you reach the end and minimum point is still larger than maximum point, you know the resulting voxel cloud has zero dimensions.

yanchith (commented):
Read through it. I quite like the resize functions, but I think they could benefit from being mutable. This would yield tremendous optimization possibilities by re-allocating less and only updating the voxel cloud descriptors - the view through which you read the data.

You also do a lot of random access into the arrays. This is not terrible, but it also isn't great, as it could cause a lot of cache misses if you iterate in a different order. You currently always iterate in the order z->y->x, which coincides with how the memory looks, but re-arranging this would still be correct, only perform 200x worse. I'd personally choose to iterate over the backing array whenever possible and compute other coordinates from that, because it more clearly documents our attempt to hit cache lines better.

One thing I really believe can be done a lot better in this PR is `shrink_to_fit`. I left an inline comment with suggestions.

src/mesh/voxel_cloud.rs
janper: I'm not strictly against empty voxel clouds but they will cause also a lot of other issues (i. e. all calculations from and to the underlying slice will have the check if the size is non-zero and the option none they return will mean more than out of bounds; or do we allow only one block dimension to be zero and the other not?). 
Also what would the zero-sized voxel cloud be good for? Is there any mathematical convention we should follow? Or is there any practical reason (I can imagine creating an empty voxel cloud and only redefine it when you already know where the voxels are going to be; or even automatically allocating the block as you keep inserting the voxels)? And is is worth it?
If it is a better design, I'll do it, I'm just not sure.

yanchith: I am not strongly *for* this, but the precedence is n-dimensional arrays, or even any kind of collections. You can create empty vectors (`Vec::new()`), and they don't allocate anything in rust.

> all calculations from and to the underlying slice will have the check if the size is non-zero and the option none they return will mean more than out of bounds; or do we allow only one block dimension to be zero and the other not

They already have to check out of bounds. If any dimension is 0, the whole length of the bit vector should be 0, right? Any indexing into a 0 length vector is out of bounds. This is not a special case.

> Also what would the zero-sized voxel cloud be good for? Is there any mathematical convention we should follow?

The only value I see is having less to think about. I think it is both a mathematics and an engineering/CS convention that zero is a totally valid length for any data collection. I believe that only in a very few (if any) places would we require more special casing, if we allowed voxel clouds with zero dimensions.

janper:
Sorry for the long PR. This change deals with several issues:

- some fixes in the existing logic
- allow for zero length voxel clouds
- refactor that allows for treating the voxel cloud in a mutable way
- mutable resize, shrink and growth
- while refactoring, I've (hopefully) optimized also the existing functions - now they iterate sequentially through the underlying voxel map

Most of the functionality is tested but I'd like to know which tests would you like me ti implement.

src/mesh/voxel_cloud.rs
```diff
@@ -128,111 +128,261 @@ impl VoxelCloud {
...
+
+    #[allow(dead_code)]
+    pub fn voxel_dimensions(&self) -> &Vector3<f32> {
+        &self.voxel_dimensions
+    }
+
+    #[allow(dead_code)]
+    pub fn voxel_map_len(&self) -> usize {
+        self.voxel_map.len()
+    }
```
yanchith: This should probably remain an implementation detail. If/When we replace the boolean array with a bit array, this will change. We could, however, have `fn voxel_count() -> usize;` or similar, which always returns `block_dimensions.x * block_dimensions.y * block_dimensions.z`.

src/mesh/voxel_cloud.rs
```diff
@@ -128,111 +128,261 @@ impl VoxelCloud {
...
+
+    #[allow(dead_code)]
+    pub fn block_dimensions(&self) -> &Vector3<u32> {
+        &self.block_dimensions
+    }
+
+    #[allow(dead_code)]
+    pub fn voxel_dimensions(&self) -> &Vector3<f32> {
+        &self.voxel_dimensions
+    }
```
yanchith: Nit: I'd return these types as owned (just remove the `&`s). There will likely be no performance difference and the functions will be easier to use. I'd still pass them by references when the types are parameters.

src/mesh/voxel_cloud.rs
```diff
@@ -128,111 +128,261 @@ impl VoxelCloud {
...
+            Vector3::new(1, 0, 0),
+            Vector3::new(0, -1, 0),
+            Vector3::new(0, 1, 0),
+            Vector3::new(0, 0, -1),
+            Vector3::new(0, 0, 1),
+        ];
+
         // If the voxels in the existing voxel cloud reach the boundaries of the
         // block, it's needed to grow the block by 1 in each direction.
-        let new_block_start = self.block_start - Vector3::new(-1, -1, -1);
```
yanchith, line 136: Nice catch, the old code sure looks incorrect :+1:

src/mesh/voxel_cloud.rs
```diff
@@ -128,111 +128,261 @@ impl VoxelCloud {
...
+                    ) {
+                        Some(original_index) => original_voxel_map[original_index],
+                        _ => false,
+                    }
+            }
+        }
+    }
+
+    /// Resize the voxel cloud block to exactly fit the volumetric geometry.
+    /// Returns None for empty the voxel cloud.
```
yanchith: This comment line is no longer true.

src/mesh/voxel_cloud.rs
```diff
@@ -128,111 +128,261 @@ impl VoxelCloud {
...
+    #[allow(dead_code)]
+    pub fn shrink_to_fit(&mut self) {
+        let mut min: Vector3<i32> = Vector3::new(i32::MAX, i32::MAX, i32::MAX);
+        let mut max: Vector3<i32> = Vector3::new(i32::MIN, i32::MIN, i32::MIN);
```
yanchith: Nit: You can use `i32::{min,max}_value()` associated function don't have to import the `i32` module. (The functions are newer than the modules, that's why there are 2 ways to do this. I think the modules level constants will be deprecated eventually)

src/mesh/voxel_cloud.rs
```diff
@@ -128,111 +128,261 @@ impl VoxelCloud {
...
+                if relative_coords.z < min.z {
+                    min.z = relative_coords.z;
+                }
+                if relative_coords.z > max.z {
+                    max.z = relative_coords.z;
+                }
+            }
+        }
+        // It's enough to check one of the values because if anything is found,
+        // all the values would change.
```
yanchith, line 369: I agree, but would feel a bit safer with asserts guarding this.

src/mesh/voxel_cloud.rs
```diff
@@ -367,84 +518,186 @@ impl VoxelCloud {
...
+}
+
+/// Calculates the absolute coordinates of the block end
+#[allow(dead_code)]
+fn calculate_block_end(block_start: &Point3<i32>, block_dimensions: &Vector3<u32>) -> Point3<i32> {
+    Point3::new(
+        block_start.x + cast_i32(block_dimensions.x),
+        block_start.y + cast_i32(block_dimensions.y),
+        block_start.z + cast_i32(block_dimensions.z),
+    )
```
yanchith: Since these 3 are quite trivial, I'd only add them once we really need them. Otherwise we are just paying for them with compile time.

src/mesh/voxel_cloud.rs
```diff
@@ -128,111 +128,261 @@ impl VoxelCloud {
...
+    pub fn voxel_map_len(&self) -> usize {
+        self.voxel_map.len()
+    }
+
     /// For each existing voxel turn on all neighbor voxels to grow (offset) the
     /// volumes stored in the voxel cloud.
-    pub fn grow(&self) -> Self {
+    ///
+    /// FIXME: This is not the most efficient way of doing this, but this
+    /// function will become obsolete with Distance field.
```
yanchith: I agree, but let's let's put this FIXME in a normal comment, not a doc comment.

src/mesh/voxel_cloud.rs
```diff
@@ -128,111 +128,261 @@ impl VoxelCloud {
...
     pub fn set_voxel_at_cartesian_coords(&mut self, point: &Point3<f32>, state: bool) {
-        let voxel_coords = self.cartesian_to_absolute_voxel_coords(point);
+        let voxel_coords = cartesian_to_absolute_voxel_coords(point, &self.voxel_dimensions);
         self.set_voxel_at_absolute_coords(&voxel_coords, state);
     }
 
+    /// Resize the voxel cloud block to match new block start and block dimensions.
+    ///
+    /// This clips the outstanding parts of the original voxel cloud.
+    #[allow(dead_code)]
```
yanchith: `#[allow(dead_code)]` works recursively. Only the following really need to be marked:

- `block_start`
- `block_dimensions`
- `voxel_dimensions`
- `voxel_map_len`
- `voxel_at_cartesian_coords`
- `set_voxel_at_relative_coords`
- `shrink_to_fit`
- `cartesian_to_relative_voxel_coords`
- `absolute_voxel_to_cartesian_coords`

(This already includes that some of the trivial dead functions shouldn't probably be included)

yanchith (changes requested):
Nice work! I've got a couple of minor inline comments.

As for the things here I think are are worth testing:

- `resize` from zero to nonzero dimensions
    - should initially be all `false`
    - should have correct block dimensions and underlying voxel count
- `resize` from nonzero to zero dimensions
    - should have correct block dimensions and underlying voxel count
-  `resize` from between two nonzero dimensions
    - should have correct block dimensions and underlying voxel count
    - newly grown area should contain `false` voxels (if applicable)
    - old voxels should still be present (if applicable)
    - outstanding voxels should be clipped (if applicable)
- `shrink_to_fit` with non-empty data should shrink to the data
- `shrink_to_fit` with empty data should shrink to empty dimensions

All of these should be pretty easily unit-testable.

src/mesh/voxel_cloud.rs
janper: I'm not sure if we want to replace the boolean array with anything else. Two reasons:

1. I've already relied too much on the existing data structure in the later features (sure, some refactor would do).
2. We'll most probably prefer using distance fields. In fact, I think we can eventually replace the current voxels entirely. I've developed the voxels further because it's a simplified version of the distance fields, so I thought it will be easier to implement and test the features with voxels and only then refactor it to distance fields. 

I'm not entirely sure how will it go, but I'll slowly start working on it and we'll see.

Btw, roughly, how much faster/efficient would the bit array be?

Regarding this specific method, I think it can be removed. I'm not using this outside of the struct anyway and form inside there are other, more explicit ways.

src/mesh/voxel_cloud.rs
janper: They'll come back with the very next PR ;)

src/mesh/voxel_cloud.rs
janper: #wasntme Autoformatter did this ;)

src/mesh/voxel_cloud.rs
yanchith: > We'll most probably prefer using distance fields

Yes, bit arrays don't make sense for distance fields.

> Btw, roughly, how much faster/efficient would the bit array be?

Probably not much faster, but more space efficient.

> Regarding this specific method, I think it can be removed.

I am fine with that. The less dead code we have, the better.

src/mesh/voxel_cloud.rs
yanchith: bad autoformat, bad!

yanchith:
Nice! Please resolve conflicts before I review though.

src/mesh/voxel_cloud.rs
```diff
@@ -1,9 +1,9 @@
-use std::f32;
 use std::iter;
+use std::{f32, i32};
```
yanchith: You can remove the `i32` import, because we are now using `i32::max_value()` instead of `i32::MAX`. Don't know why the compiler doesn't warn us here.

yanchith (dismissed):
Just one last nit and we are good to go.

PR #107, opened 2019-12-28
Extract largest mesh from a group

janper:
After importing a 3d-scanned mesh this helps to remove mesh crumbs and some erroneous artifacts.

While reviewing also please have a look at this: https://trello.com/c/RYwqyS1P/258-does-import-automatically-join-meshes-because-after-importing-an-obj-with-more-objects-any-extract-returns-the-whole-contents-of

### DIFF
--- a/src/interpreter_funcs/extract_largest.rs
+++ b/src/interpreter_funcs/extract_largest.rs
@@ -57,7 +57,7 @@ impl Func for FuncExtractLargest {
         let mut mesh = mesh_iter.next().expect("Array must not be empty");
         let mut largest_face_count = mesh.faces().len();
 
-        while let Some(current_mesh) = mesh_iter.next() {
+        for current_mesh in mesh_iter {
             let current_face_count = current_mesh.faces().len();
             if current_face_count > largest_face_count {
                 largest_face_count = current_face_count;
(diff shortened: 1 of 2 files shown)


src/interpreter_funcs/extract_largest.rs
```diff
@@ -0,0 +1,76 @@
...
+                let current_mesh = mesh_array
+                    .get_refcounted(current_index)
+                    .expect("Array must not be empty");
+                let current_face_count = current_mesh.faces().len();
+                if current_face_count > largest_face_count {
+                    largest_face_count = current_face_count;
+                    mesh = current_mesh;
+                }
+            }
+        }
```
yanchith: Nit this would probably be a bit cleaner with `MeshArrayValue::iter_refcounted()`. Pseudocode:

```rust
let iter = mesh_array.iter_refcounted();
let mut mesh = iter.next().expect("Array must not be empty");
while let Some(mesh) = iter.next() {
    // your code...
}
```

yanchith (dismissed):

yanchith:
Looks good now, clippy has a valid suggestion though.

yanchith (approved):

PR #109, opened 2020-01-02
Calculate bounding box from voxel cloud, voxel volume and vice versa

janper:


src/mesh/analysis.rs
```diff
@@ -434,6 +434,38 @@ pub fn are_visually_similar(mesh1: &Mesh, mesh2: &Mesh) -> bool {
             .all(|f| unpacked_faces1.clone().any(|g| f == g))
 }
 
+/// Check if the two bounding boxes intersect / share any portion of space
+
+// https://math.stackexchange.com/questions/2651710/simplest-way-to-determine-if-two-3d-boxes-intersect
+#[allow(dead_code)]
+pub fn do_bounding_boxes_intersect(b1: &BoundingBox, b2: &BoundingBox) -> bool {
```
yanchith: Don't we want this to be a method on bounding box?

src/mesh/voxel_cloud.rs
```diff
@@ -51,17 +51,10 @@ impl VoxelCloud {
         }
     }
 
-    /// Calculate a voxel cloud from an existing mesh.
-    pub fn from_mesh(mesh: &Mesh, voxel_dimensions: &Vector3<f32>) -> Self {
-        assert!(
-            voxel_dimensions.x > 0.0 && voxel_dimensions.y > 0.0 && voxel_dimensions.z > 0.0,
-            "One or more voxel dimensions are 0.0"
-        );
```
yanchith: I think we lost this assert in refactoring.

src/mesh/voxel_cloud.rs
```diff
@@ -127,19 +128,26 @@ impl VoxelCloud {
...
+    /// Returns voxel cloud block dimensions in voxel units
     #[allow(dead_code)]
-    pub fn voxel_dimensions(&self) -> Vector3<f32> {
-        self.voxel_dimensions
+    pub fn block_dimensions(&self) -> &Vector3<u32> {
+        &self.block_dimensions
+    }
+
+    /// Returns single voxel dimensions in model space units
+    pub fn voxel_dimensions(&self) -> &Vector3<f32> {
```
yanchith: This looks like bad merge, There was already a version of this function returning the owned `Vector3`, which was commented on in previous PR.

src/mesh/voxel_cloud.rs
```diff
@@ -127,19 +128,26 @@ impl VoxelCloud {
...
+    /// Returns voxel cloud block end in absolute voxel coordinates
+    pub fn block_end(&self) -> Point3<i32> {
+        calculate_block_end(&self.block_start, &self.block_dimensions)
     }
 
+    /// Returns voxel cloud block dimensions in voxel units
     #[allow(dead_code)]
-    pub fn voxel_dimensions(&self) -> Vector3<f32> {
-        self.voxel_dimensions
+    pub fn block_dimensions(&self) -> &Vector3<u32> {
```
yanchith: Bad merge. Let's stay with returning owned `Vector3`.

src/mesh/voxel_cloud.rs
```diff
@@ -708,6 +728,33 @@ fn relative_voxel_to_cartesian_coords(
...
+#[allow(dead_code)]
+fn relative_to_absolute_voxel_coords(
+    relative_coords: &Point3<i32>,
+    block_start: &Point3<i32>,
+) -> Point3<i32> {
+    relative_coords + block_start.coords
+}
+
+/// Calculates the absolute coordinates of the block end
+fn calculate_block_end(block_start: &Point3<i32>, block_dimensions: &Vector3<u32>) -> Point3<i32> {
```
yanchith: I would inline this in its only use site.

src/mesh/voxel_cloud.rs
```diff
@@ -708,6 +728,33 @@ fn relative_voxel_to_cartesian_coords(
...
+}
+
+/// Converts absolute voxel coordinates into relative voxel coordinates
+#[allow(dead_code)]
+fn relative_to_absolute_voxel_coords(
+    relative_coords: &Point3<i32>,
+    block_start: &Point3<i32>,
+) -> Point3<i32> {
+    relative_coords + block_start.coords
+}
```
yanchith: Same as in previous PR. These two are still not used. I'd add them only once they are used.

src/mesh/voxel_cloud.rs
```diff
@@ -554,6 +537,43 @@ impl VoxelCloud {
         // and weld naked edges
         tools::weld(&joined_voxel_mesh, (min_voxel_dimension as f32) / 4.0)
     }
+
+    #[allow(dead_code)]
+    pub fn calculate_bounding_box(&self) -> BoundingBox {
+        let voxel_dimensions = self.voxel_dimensions();
```
yanchith: Doesn't need to be a method call. `self.voxel_dimensions` is a field.

src/mesh/voxel_cloud.rs
```diff
@@ -554,6 +537,43 @@ impl VoxelCloud {
         // and weld naked edges
         tools::weld(&joined_voxel_mesh, (min_voxel_dimension as f32) / 4.0)
     }
+
+    #[allow(dead_code)]
+    pub fn calculate_bounding_box(&self) -> BoundingBox {
```
yanchith: I'd just name this `bounding_box()`

src/mesh/voxel_cloud.rs
```diff
@@ -554,6 +537,43 @@ impl VoxelCloud {
...
+            Point3::new(
+                block_end.x as f32 * voxel_dimensions.x,
+                block_end.y as f32 * voxel_dimensions.y,
+                block_end.z as f32 * voxel_dimensions.z,
+            ),
+        )
+    }
+
+    #[allow(dead_code)]
+    pub fn calculate_bounding_box_of_contained_volume(&self) -> BoundingBox {
```
yanchith: I'd just name this `volume_bounding_box()`

src/mesh/voxel_cloud.rs
```diff
@@ -287,59 +294,49 @@ impl VoxelCloud {
         resized_block_start: &Point3<i32>,
         resized_block_dimensions: &Vector3<u32>,
     ) {
-        if resized_block_start != &self.block_start
```
yanchith: Let's undo this change. This looks like reverting to an older version of the code without the if condition guarding the resize that I commented on previous PR.

src/mesh/voxel_cloud.rs
```diff
@@ -287,59 +294,49 @@ impl VoxelCloud {
...
+    /// original block start and zero block dimensions.
+    pub fn compute_volume_boundaries(&self) -> (Point3<i32>, Vector3<u32>) {
         let mut min: Vector3<i32> =
             Vector3::new(i32::max_value(), i32::max_value(), i32::max_value());
         let mut max: Vector3<i32> =
             Vector3::new(i32::min_value(), i32::min_value(), i32::min_value());
-        for (i, v) in self.voxel_map.iter().enumerate() {
-            if *v {
+        for i in 0..self.voxel_map.len() {
+            if self.voxel_map[i] {
```
yanchith: I think the previous version was slightly better.

src/mesh/voxel_cloud.rs
```diff
@@ -365,44 +362,30 @@ impl VoxelCloud {
...
-            assert_eq!(
-                max.y,
-                i32::min_value(),
-                "Voxel cloud emptiness check failed"
-            );
-            assert_eq!(
-                max.z,
-                i32::min_value(),
-                "Voxel cloud emptiness check failed"
-            );
```
yanchith: Let's not loose these asserts.

src/mesh/voxel_cloud.rs
```diff
@@ -365,44 +362,30 @@ impl VoxelCloud {
                 }
             }
         }
+        let (block_start, block_dimensions) =
```
yanchith: I think you don't need to `let` this, the last expression of the function can be an `if/else` block

yanchith (commented):
Nice refactor :+1:  It unfortunately brought back some changes we already resolved in previous PRs. I marked them with inline comments.

We could also be more consistent on the whole `compute` vs `calculate` thing :smile:

src/mesh/voxel_cloud.rs
yanchith: Or something similar.

src/mesh/analysis.rs
janper: why not

src/mesh/voxel_cloud.rs
janper: I wished I'd had these when making the previous PR. Let's please keep such functions. There never seems to be a good time to add them, but they are very useful when they are around.

src/mesh/analysis.rs
```diff
@@ -103,6 +103,38 @@ impl BoundingBox {
     pub fn diagonal_length(&self) -> f32 {
         nalgebra::distance(&self.minimum_point, &self.maximum_point)
     }
+
+    /// Check if the two bounding boxes intersect / share any portion of space
+
+    // https://math.stackexchange.com/questions/2651710/simplest-way-to-determine-if-two-3d-boxes-intersect
```
yanchith: Don't change it now, but for the future, feel free to include source links in the doc comments. Also try not having blank lines between doc comments and the things they should be attached to. I ignored the first few cases of you doing this in previous PRs, but I haven't seen this style used anywhere else.

src/mesh/analysis.rs
```diff
@@ -103,6 +103,38 @@ impl BoundingBox {
...
+    #[allow(dead_code)]
+    pub fn do_bounding_boxes_intersect(&self, other: &BoundingBox) -> bool {
```
yanchith: Now that this is a method, we can name it much better, e.g. `intersects` or `intersects_with`.

yanchith:
> I wished I'd had these when making the previous PR. Let's please keep such functions. There never seems to be a good time to add them, but they are very useful when they are around.

Not sure which PR you are referring to, but if you mean #106, you had them and they were unused, same as here. They were not present in any other PR I've seen.

janper:
https://github.com/subdgtl/HURBAN-Selector/pull/109/commits/6b1c52d070911c10772d56af6df63c55d167e505

yanchith:
I never had anything against `block_end`, only those unused, `#[dead_code]` oneliner functions.

janper:
ok

yanchith (commented):
Getting closer

src/mesh/voxel_cloud.rs
```diff
@@ -538,6 +547,43 @@ impl VoxelCloud {
...
+            Point3::new(
+                (volume_start.x as f32 - 0.5) * voxel_dimensions.x,
+                (volume_start.y as f32 - 0.5) * voxel_dimensions.y,
+                (volume_start.z as f32 - 0.5) * voxel_dimensions.z,
+            ),
+            Point3::new(
+                (volume_start.x as f32 + volume_dimensions.x as f32 + 0.5) * voxel_dimensions.x,
+                (volume_start.y as f32 + volume_dimensions.x as f32 + 0.5) * voxel_dimensions.y,
+                (volume_start.z as f32 + volume_dimensions.x as f32 + 0.5) * voxel_dimensions.z,
+            ),
```
yanchith, line 606: If I understand correctly, adding and subtracting 0.5 here will increase the size of the bounding box. Imagine this situation.

```rust
let bb1 = initial_bounding_box;
let mut vc1 = VoxelCloud::from_bounding_box();
vc1.set_all(true); // We don't actually have this method, but nevermind

let bb2 = vc1.volume_bounding_box();
let vc2 = VoxelCloud::from_bounding_box();

// I think it is intuitive that this assert holds, but I don't think it will for the current situation.
// We use `floor`/`ceil` in `VoxelCloud::from_bounding_box` and we add/subtract 0.5 in `VoxelCloud::volume_bounding_box()`
assert_eq!(vc1.voxel_dimensions, vc2.voxel_dimensions);
```

yanchith: I actually realized that this also happens with our current `to_mesh` implementation. (as you noted on slack, the original behavior gives the same bounding box the mesh returned from `to_mesh` would).

So if you repeat my previous comment and go `bounding_box` -> `voxel_cloud` -> `mesh` -> `bounding_box` -> `voxel_cloud`, it will keep growing.

I am just noting this, I am not sure this is a problem. Would be a nice property if this didn't happen, though. One pretty easy fix (I think) for this would be to use `round` instead of `floor`/`ceil` in `VoxelCloud::from_bounding_box`, but not sure about how this would interact with other assumptions we have. Thoughts?

src/mesh/analysis.rs
janper: autoformatter requires either a blank line starting with `///` or `//` to keep a newline in the doc. without this blank line, If the comment block starts with `///`, autoformatter adds the third slash to the `//` block. that's why I keep the blank line.
I don't mind adding this into the triple slash doc scope though.
formatter

src/mesh/voxel_cloud.rs
janper: I understand your point, however the intedned use was different. I'd prefer renaming the `volume_bounding_box` to something else and adding a method doing exactly what you expect. 

This is the intended use:
- `bounding_box()` - to compare more voxel clouds if they describe the same parts of space. downside - bounding boxes are defined in the model (world) coordinates, so there is an extra conversion. the good part - you can compare voxel clouds with voxels of various dimensions. in fact, all the upcoming voxel operations requires the voxel clouds to contain voxels of the same size, but maybe they don't have to. on the other hand, the way we'll use them, we'll always use voxels of the same size.
- `from_bounding_box()` - after comparing more bounding boxes, you can generate their intersection or union and then in return generate a new voxel cloud based on this.
- `volume_bounding_box()` is a more efficient way of `to_mesh()` -> `BoundingBox::from_mesh()`. it's not intedned to be used for more operations with bounding boxes. I'm fine with renaming this method.
- I'm not against implementing also what you suggest. It will be useful for `shrink_to_fit`, however that one is already done differently and some refactor implementing some `resize_to_bounding_box` will be required.

src/mesh/analysis.rs
yanchith: Which autoformatter? `rustfmt` doesn't format comments at all, I think. This looks like a VS code specific thing.

src/mesh/analysis.rs
```diff
@@ -105,10 +105,9 @@ impl BoundingBox {
     }
 
     /// Check if the two bounding boxes intersect / share any portion of space
-
-    // https://math.stackexchange.com/questions/2651710/simplest-way-to-determine-if-two-3d-boxes-intersect
+    /// https://math.stackexchange.com/questions/2651710/simplest-way-to-determine-if-two-3d-boxes-intersect
```
yanchith, line 111: Again, no need to change this now, but note that these comments are markdown and that they do have a usual format they should be in (when in doubt, look at std documentation). This is how this actually renders:

rendered

src/mesh/voxel_cloud.rs
yanchith: Okay, let's stick with the rename for now and we can revisit later if this is painful. Some documentation would be nice, so that we don't need to relearn the problem later again. I'll add it to this PR now.

yanchith (dismissed):

yanchith (approved):

PR #110, opened 2020-01-04
Replace "calculate" with "compute" and remove obsolete FIXME

janper:
Please have a look if you agree. No problem if not.


yanchith (approved):
I think some of the snapshot testing FIXMEs were not actually done, but those those are kinda low prio anyway and were quite nondescript, so we probably wouldn't even know what to do with them.

PR #112, opened 2020-01-06
Add voxel-based boolean operations on mesh geometry

janper:
All operations first convert the geometry into voxel clouds , then perform the logical operations and eventually convert the voxel clouds into meshes. 

**Intersection**: boolean AND - keeps those voxels, that are set to true in both voxel cloud operands
**Union**: boolean OR - keeps those voxels, that are set to true in either voxel cloud operands
**Difference**: keeps those voxels, that are set to true in the first voxel cloud operand but not set to true in the second

This commit implements the Intersection and Union operations on two voxel clouds, one of which is being mutated and on a group (slice of voxel clouds), when a new voxel cloud is being created.

**Questions**:

- should we implement UI operations for groups? do we want to promote the groups that much?
- does it make sense to keep two separate versions of operations - one mutable on two voxel clouds, one immutable on many?
- shouldn't we only keep one of them, say the mutable one and turn it from _one vs. one_ to _one vs. many_?
- are the tests sufficient?


src/mesh/voxel_cloud.rs
```diff
@@ -164,6 +164,18 @@ impl VoxelCloud {
...
+    #[allow(dead_code)]
+    pub fn contains_voxels(&self) -> bool {
+        self.voxel_map.iter().any(|v| *v)
+    }
+
+    /// Checks if the voxel cloud contains any voxel / volume
+    #[allow(dead_code)]
+    pub fn is_empty(&self) -> bool {
+        self.voxel_map.is_empty()
+    }
```
janper: From outside the scope there is no good way of checking these. 
In some cases it might be more efficient or safer to check these before performing any other operation. So far we don't seem to have such situation (or maybe I've overlooked it).

src/mesh/voxel_cloud.rs
```diff
@@ -228,6 +239,139 @@ impl VoxelCloud {
         }
     }
 
+    /// Computes boolean intersection (logical AND operation) of the current and
+    /// another Voxel cloud. The Voxel clouds have to intersect and the voxel
+    /// dimension must be equal. The current Voxel cloud will be mutated and
+    /// resized to the size and position of an intersection of the two Voxel
+    /// clouds (not the contained volumes).
+    pub fn boolean_intersection_mut(&mut self, other: &VoxelCloud) {
```
janper: This could be rewritten so that `other` is many `others`. The we can get rid of the immutable version of this function. On the downside, the caller has to take care of preparing a good "self" for such an operation. Specifically, working with mesh groups will require popping the first one and then exposing it to the rest of the group.

src/mesh/voxel_cloud.rs
```diff
@@ -228,6 +239,139 @@ impl VoxelCloud {
...
+
+    /// Computes boolean union (logical OR operation) of two Voxel clouds. The
+    /// voxel dimensions of both input Voxel clouds must be equal. he current
+    /// Voxel cloud will be mutated and resized to contain both input Voxel
+    /// clouds (not the contained volumes).
+    ///
+    /// # Warning
+    /// If the input Voxel clouds are far apart, the resulting voxel cloud might
+    /// be huge.
+    pub fn boolean_union_mut(&mut self, other: &VoxelCloud) {
```
janper: Same as above

src/mesh/voxel_cloud.rs
```diff
@@ -228,6 +239,139 @@ impl VoxelCloud {
...
+                self.voxel_map[i] |= v;
+            }
+        }
+    }
+
+    /// Computes boolean difference of the current Voxel cloud minus the other
+    /// Voxel cloud. The current Voxel cloud will be modified so that voxels,
+    /// that are on in both Voxel clouds will be turned off, while the rest
+    /// remains intact.
+    pub fn boolean_difference_mut(&mut self, other: &VoxelCloud) {
```
janper: Different than above, but I can imagine implementing `self - union_of_others`

yanchith:
> This commit implements the Intersection and Union operations on two voxel clouds, one of which is being mutated and on a group (slice of voxel clouds), when a new voxel cloud is being created.

Looks like difference is implemented too.

src/interpreter_funcs/mod.rs
```diff
@@ -21,6 +24,9 @@ use self::transform::FuncTransform;
 use self::voxelize::FuncVoxelize;
 use self::weld::FuncWeld;
 
+mod boolean_difference;
+mod boolean_intersection;
+mod boolean_union;
```
yanchith: Don't we want to call these `voxel_boolean_{difference,intersection,union}` so that we leave space for mesh boolean ops?

src/interpreter_funcs/boolean_union.rs
```diff
@@ -0,0 +1,109 @@
...
+
+impl error::Error for FuncBooleanUnionError {}
+
+pub struct FuncBooleanUnion;
+
+impl Func for FuncBooleanUnion {
+    fn info(&self) -> &FuncInfo {
+        &FuncInfo {
+            name: "Union",
+            return_value_name: "Union mesh",
```
yanchith: Capitalization

src/interpreter_funcs/boolean_union.rs
```diff
@@ -0,0 +1,109 @@
...
+                refinement: ParamRefinement::Mesh,
+                optional: false,
+            },
+            ParamInfo {
+                name: "Mesh",
+                refinement: ParamRefinement::Mesh,
+                optional: false,
+            },
+            ParamInfo {
+                name: "Voxel size",
```
yanchith: Capitalization

src/interpreter_funcs/boolean_union.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    fn flags(&self) -> FuncFlags {
+        FuncFlags::PURE
+    }
+
+    fn param_info(&self) -> &[ParamInfo] {
+        &[
+            ParamInfo {
+                name: "Mesh",
```
yanchith: Don't we want to name these "Mesh 1" and "Mesh 2"? Same for all others.

src/interpreter_funcs/boolean_intersection.rs
```diff
@@ -0,0 +1,123 @@
...
+
+impl error::Error for FuncBooleanIntersectionError {}
+
+pub struct FuncBooleanIntersection;
+
+impl Func for FuncBooleanIntersection {
+    fn info(&self) -> &FuncInfo {
+        &FuncInfo {
+            name: "Intersection",
+            return_value_name: "Intersection mesh",
```
yanchith: Capitalization

src/interpreter_funcs/boolean_intersection.rs
```diff
@@ -0,0 +1,123 @@
...
+                refinement: ParamRefinement::Mesh,
+                optional: false,
+            },
+            ParamInfo {
+                name: "Mesh",
+                refinement: ParamRefinement::Mesh,
+                optional: false,
+            },
+            ParamInfo {
+                name: "Voxel size",
```
yanchith: Capitalization

src/interpreter_funcs/boolean_difference.rs
```diff
@@ -0,0 +1,109 @@
...
+
+impl error::Error for FuncBooleanDifferenceError {}
+
+pub struct FuncBooleanDifference;
+
+impl Func for FuncBooleanDifference {
+    fn info(&self) -> &FuncInfo {
+        &FuncInfo {
+            name: "Difference",
+            return_value_name: "Difference mesh",
```
yanchith: Capitalization

src/interpreter_funcs/boolean_difference.rs
```diff
@@ -0,0 +1,109 @@
...
+                refinement: ParamRefinement::Mesh,
+                optional: false,
+            },
+            ParamInfo {
+                name: "Mesh",
+                refinement: ParamRefinement::Mesh,
+                optional: false,
+            },
+            ParamInfo {
+                name: "Voxel size",
```
yanchith: Capitalization

src/mesh/voxel_cloud.rs
```diff
@@ -19,7 +19,7 @@ use super::{primitive, tools, Face, Mesh};
 /// voxel-space coordinates. The voxel space starts at the cartesian space
 /// origin with voxel coordinates 0, 0, 0. Voxel clouds with the same voxel size
 /// are compatible and collateral operations be performed on them.
-#[derive(Debug, Clone, serde::Serialize)]
+#[derive(Debug, Clone, serde::Serialize, PartialEq)]
```
yanchith: Derive order should be: Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, nonstd derives. Le'ts move PartialEq before serde::Serialize.

src/mesh/voxel_cloud.rs
yanchith: The doc comments of these two methods probably shouldn't be the same. One looks for set voxels while the other tells you whether the voxel block is a singularity. Also, you are missing periods at the end (I know you disagree, but let's stick to a convention that already exists in `std`).

src/mesh/voxel_cloud.rs
yanchith: One other and &mut self is fine, as this is practically a mul assign op. https://doc.rust-lang.org/std/ops/trait.MulAssign.html

(don't overload the operator though)

src/mesh/voxel_cloud.rs
```diff
@@ -228,6 +239,163 @@ impl VoxelCloud {
...
+    /// Computes boolean intersection (logical AND operation) of the current and
+    /// another Voxel cloud. The Voxel clouds have to intersect and the voxel
+    /// dimension must be equal. The current Voxel cloud will be mutated and
+    /// resized to the size and position of an intersection of the two Voxel
+    /// clouds (not the contained volumes).
+    pub fn boolean_intersection_mut(&mut self, other: &VoxelCloud) {
+        assert!(
+            approx::relative_eq!(self.voxel_dimensions, other.voxel_dimensions()),
+            "The two voxel clouds don't contain voxels of the same size"
+        );
```
yanchith: Should this really be an assert? Let's at least document it in the `# Panics` block in the doc comment. Same for all other boolean ops, even the helper `many` ones.

src/mesh/voxel_cloud.rs
```diff
@@ -228,6 +239,163 @@ impl VoxelCloud {
...
+
+        // Find volume common to both voxel clouds.
+        // FIXME: This could be delegated to BoundingBox. @Optimization Compute
```
yanchith: Sorry for the long comment, but I have thoughts here. In the order of importance:

1) After looking at the usage code, I think we will want two versions of each function. One that produces the bounding box by applying the boolean operation on the bounding boxes of the original voxel clouds (what we have now), and one that produces a bounding box as if `shrink_to_fit` was called on the results, but instead of calling `shrink_to_fit`, we rather compute the final bounding box from the volumes.

2) As seen in previous comment, `BoundingBox` could use logical operations. Let's not do immutable ones though, since it is a `Copy` type.

3) This FIXME is currently incorrect. `BoundingBox` does not guarantee the ordering of it's points. We might want to change the internal representation of it so that it does, maybe something like `origin: Point3` and `diagonal: Vector3`, where the diagonal will always have to be positive in all components. `BoundingBox` would still provide getters so that the outside could get the min and max point, same as now. All constructors would need to be fixed to ensure the new guarantee.

(I know you wanted to do this earlier, but what neither of us realized at that time is that we can normalize the bounding box representation to origin + diagonal.)

4) I just now realized that these are integers. Maybe we want to have a parametric bounding box that works for integers too? Most of the methods would only work on the floating point one, but some of the would work for integers too. We will also need to implement conversions between the two bounding boxes. Have look at this code to see how conditional impl blocks work: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.resize . You may want to use the `num_traits` crate to get the generic constraints. If this looks too complex, it may be simpler to have a separate type for the integer (i32) bounding box, but my gut says polymorphism is the way to go here. We can start our way with the duplicate implementation, and then unify them.

4) Let's do what this FIXME suggests, using volume bounding boxes converted to integer bounding boxes.

Optional) While at this, I would remove the unused `new_separate` constructor of `BoundingBox`. I don't think it will ever be used, especially since we have `new` that does exactly the same and is imo always preferable.

Same comment applies everywhere you put this FIXME.

src/mesh/voxel_cloud.rs
```diff
@@ -228,6 +239,163 @@ impl VoxelCloud {
...
+        );
+
+        // If the voxel clouds don't share the same chunk of space, there can't
+        // be any intersection.
+        assert!(
+            (max_coord.coords - min_coord.coords)
+                .iter()
+                .all(|v| *v >= 0),
+            "The two voxel clouds don't intersect"
+        );
```
yanchith: Empty voxel cloud is fine, we don't need to assert.

src/mesh/voxel_cloud.rs
```diff
@@ -228,6 +239,163 @@ impl VoxelCloud {
...
+            )
+            .expect("The current voxel map out of bounds");
+
+            // Perform boolean AND on voxel states of both voxel clouds
+            self.voxel_map[i] &= other
+                .voxel_at_absolute_coords(&absolute_coords)
+                .expect("The other voxel map out of bounds");
+        }
+
+        // FIXME: consider calling shrink_to_fit
```
yanchith: I'd try to decide this in this PR. Got any arguments for and against? Also, I wouldn't go for shrink_to_fit, but rather just resize to the final size the first time around, using the (possibly integer based) bounding box. Whatever we do, we should document the behavior in doc comments.

Same applies to `boolean_union` and `boolean_difference`.

src/mesh/voxel_cloud.rs
```diff
@@ -228,6 +239,163 @@ impl VoxelCloud {
         }
     }
 
+    /// Computes boolean intersection (logical AND operation) of the current and
+    /// another Voxel cloud. The Voxel clouds have to intersect and the voxel
+    /// dimension must be equal. The current Voxel cloud will be mutated and
+    /// resized to the size and position of an intersection of the two Voxel
+    /// clouds (not the contained volumes).
+    pub fn boolean_intersection_mut(&mut self, other: &VoxelCloud) {
```
yanchith: Let's drop the `_mut` suffix. AFAIK it is only conventionally used to get a mutable reference to a data structure, which is not the current case (see `iter_mut`, `borrow_mut` or `as_mut`, e.g.). Be sure to let me know, if you found a place in `std` that uses the `_mut` suffix differently and/or in the sense you are using it now. I couldn't find it.

Same for all other `_mut` suffixes.

src/mesh/voxel_cloud.rs
```diff
@@ -628,6 +802,170 @@ impl VoxelCloud {
...
+pub fn boolean_intersection_many<'a, T>(voxel_clouds: T) -> VoxelCloud
+where
+    T: IntoIterator<Item = &'a VoxelCloud> + Clone,
+{
+    // Check if all the input voxel clouds contain voxels of the same size.
+    let first_voxel_dimensions = voxel_clouds
+        .clone()
+        .into_iter()
+        .next()
+        .expect("The list of voxel clouds is empty")
```
yanchith: This shouldn't be a problem. empty list -> empty intersection -> empty resulting voxel cloud. All ok. We can short circuit here.

src/mesh/voxel_cloud.rs
```diff
@@ -228,6 +239,163 @@ impl VoxelCloud {
...
+                i,
+                &min_coord,
+                &block_dimensions,
+            )
+            .expect("The current voxel map out of bounds");
+
+            // Perform boolean AND on voxel states of both voxel clouds
+            self.voxel_map[i] &= other
+                .voxel_at_absolute_coords(&absolute_coords)
+                .expect("The other voxel map out of bounds");
```
yanchith: Shouldn't we set the voxel to false instead of panicking? `unwrap_or(false)`

src/mesh/voxel_cloud.rs
```diff
@@ -228,6 +239,163 @@ impl VoxelCloud {
...
+        // Iterate through the block of space containing both voxel clouds.
+        for i in 0..self.voxel_map.len() {
+            let absolute_coords = one_dimensional_to_absolute_three_dimensional_coordinate(
+                i,
+                &min_coord,
+                &block_dimensions,
+            )
+            .expect("The current voxel map out of bounds");
+            // If the other voxel cloud exists on the current absolute
+            // coordinate, perform boolean OR, otherwise don;t change the
```
yanchith: "don't"

src/mesh/voxel_cloud.rs
```diff
@@ -228,6 +239,163 @@ impl VoxelCloud {
...
+            let absolute_coords = one_dimensional_to_absolute_three_dimensional_coordinate(
+                i,
+                &self.block_start,
+                &self.block_dimensions,
+            )
+            .expect("The current voxel map out of bounds");
+
+            // If the other voxel clouds contains a voxel at the position,
+            // remove the existing voxel from the target voxel cloud
+            if self.voxel_map[i] {
```
yanchith: This `if` seems redundant.

src/mesh/voxel_cloud.rs
```diff
@@ -628,6 +802,170 @@ impl VoxelCloud {
...
+        // If all the input voxel clouds contain a voxel at the given absolute
+        // coordinate, put there a voxel to the intersection voxel cloud.
+        intersection_voxel_cloud.voxel_map[i] = voxel_clouds.clone().into_iter().all(|vc| {
+            vc.voxel_at_absolute_coords(&absolute_coords)
+                .expect("The other voxel map out of bounds")
+        });
+    }
+
+    // FIXME: consider calling shrink_to_fit
+    intersection_voxel_cloud
```
yanchith: What if we reused the already existing mutating function instead of re-implementing the same logic? We can fill the initial voxel cloud with `true` and intersect it with the others. IIRC, this shouldn't cause resizes, as intersecting can only make the voxel cloud smaller, never larger.

src/mesh/voxel_cloud.rs
```diff
@@ -628,6 +802,170 @@ impl VoxelCloud {
...
+        min_coord.y = min_coord.y.max(block_start.y);
+        min_coord.z = min_coord.z.max(block_start.z);
+        max_coord.x = max_coord.x.min(block_end.x);
+        max_coord.y = max_coord.y.min(block_end.y);
+        max_coord.z = max_coord.z.min(block_end.z);
+    }
+
+    // If the voxel clouds don't share the same chunk of space, there can't be
+    // any intersection.
+    assert!(
```
yanchith: Is this really assertworthy? I would say this is valid input for which we can provide valid (empty) output.

src/mesh/voxel_cloud.rs
```diff
@@ -628,6 +802,170 @@ impl VoxelCloud {
...
+pub fn boolean_union_many<'a, T>(voxel_clouds: T) -> VoxelCloud
+where
+    T: IntoIterator<Item = &'a VoxelCloud> + Clone,
+{
+    // Check if all the input voxel clouds contain voxels of the same size.
+    let first_voxel_dimensions = voxel_clouds
+        .clone()
+        .into_iter()
+        .next()
+        .expect("The list of voxel clouds is empty")
```
yanchith: Not a problem. Short circuit to empty voxel cloud result.

src/mesh/voxel_cloud.rs
```diff
@@ -628,6 +802,170 @@ impl VoxelCloud {
...
+        union_voxel_cloud.voxel_map[i] = voxel_clouds.clone().into_iter().any(|vc| {
+            match vc.voxel_at_absolute_coords(&absolute_coords) {
+                Some(true) => true,
+                _ => false,
+            }
+        });
+    }
+
+    // FIXME: consider calling shrink_to_fit
+    union_voxel_cloud
```
yanchith: Don't we want to reuse the existing mutable functionality here? It shouldn't even be less efficient, as we can compute the final volume bounding box for union easily.

yanchith (commented):
Overall a nice piece of functionality :+1: 

As always, I do think we can improve on some parts.

Moreover, I am not sure I am a fan of the `many` functions, even more so because they are unused. But I did review them, so if you feel strongly about keeping them, we can.

I agree with your FIXME regarding using bounding boxes. I think all the six boolean operation methods can become much simpler, so I consider it worth doing in this PR.

src/interpreter_funcs/boolean_difference.rs
```diff
@@ -0,0 +1,109 @@
...
+
+use crate::interpreter::{
+    Float3ParamRefinement, Func, FuncError, FuncFlags, FuncInfo, ParamInfo, ParamRefinement, Ty,
+    UintParamRefinement, Value,
+};
+use crate::mesh::voxel_cloud::VoxelCloud;
+
+#[derive(Debug, PartialEq)]
+pub enum FuncBooleanDifferenceError {
+    WeldFailed,
```
yanchith: This is missing the error case when the difference produces a empty voxel cloud.

src/interpreter_funcs/boolean_intersection.rs
```diff
@@ -0,0 +1,123 @@
...
+use crate::interpreter::{
+    Float3ParamRefinement, Func, FuncError, FuncFlags, FuncInfo, ParamInfo, ParamRefinement, Ty,
+    UintParamRefinement, Value,
+};
+use crate::mesh::voxel_cloud::VoxelCloud;
+
+#[derive(Debug, PartialEq)]
+pub enum FuncBooleanIntersectionError {
+    WeldFailed,
+    NotIntersecting,
```
yanchith: Instead of this, I'd change this to the more general error case when the resulting voxel cloud is empty. If you want to keep to give the user more info, that's fine as well, but in that case we need another error variant, because the interesection can be empty even if the voxel clouds intersect.

yanchith (commented):
Found 2 bugs when playing with it

yanchith:
I know we expect this, but the user maybe won't:
booleans

While very cool, this is caused by the way we voxelize meshes. Shouldn't the voxel cloud be computed from the actual volume of the mesh instead of just the faces? (I imagine the math for this will be nontrivial)

Also, shouldn't converting a voxel cloud back to mesh create non-hollow meshes?

yanchith:
> should we implement UI operations for groups? do we want to promote the groups that much?

I don't think we have enough information to know this yet. Let's re-evaluate a bit later. It may be worth considering changing everything that's currently mesh to a group, but this may have unforeseen consequences.

> does it make sense to keep two separate versions of operations - one mutable on two voxel clouds, one immutable on many?

Already partially addressed this. I believe after implementing some of the suggestions in this PR, the `many` versions will become so trivial that we might not even want them.

> shouldn't we only keep one of them, say the mutable one and turn it from one vs. one to one vs. many?

Interesting thought. I am not opposed, but let's do that only once we have a usecase... otherwise it's wasted work.

> are the tests sufficient?

Didn't review them properly yet, because I suggested functionality changes that will change the tests too. Their quantity seems sufficient, though :joy:

src/interpreter_funcs/mod.rs
janper: yup

src/interpreter_funcs/boolean_union.rs
janper: ok

src/mesh/voxel_cloud.rs
janper: i can imagine this would work also for voxel clouds with voxels of different sizes - by requesting the other through cartesian coordinates of self. why not, it just seems a bit dirty.

src/mesh/voxel_cloud.rs
janper: but is it? 
1. where is the `block_start` then?
2. how would it then perform `to_mesh` when we don't allow empty meshes?
3. also, if we don't allow empty meshes, why would we allow zero length voxel clouds?

src/mesh/voxel_cloud.rs
janper: we can avoid `shrink_to_fit` for optimization reasons (save one crawl through the block) but using just bounding boxes is not enough - at least not upfront. i'm currently trimming (or growing) the block without checking if the output contains any actual voxels. it's just a chunk of space that potentially can contain some voxels. we'll only know after the boolean operation what is the actual size of the space that the voxels really occupy. i can keep storing the boundary extrems while performing the boolean operation and then perform resize but it is exaclty what `shrink_to_fit` does except the search for active voxels, which is just one pass through the block.

janper: unless you mean generating a bounding box from the volume of the result of the boolean operation and then trimming to that bounding box <- which seems like an overkill

src/mesh/voxel_cloud.rs

src/mesh/voxel_cloud.rs
janper: that should never happen due to the resize of the target voxel cloud. if it does, there is an error in the logic, hence the assert

src/mesh/voxel_cloud.rs
janper: is `if` slower than assigning a value even if unnecessary?

src/mesh/voxel_cloud.rs
janper: see my other question about empty voxel clouds

src/mesh/voxel_cloud.rs
janper: true, we can do it. or remove the immutable version whatsoever

src/mesh/voxel_cloud.rs
janper: 1. I'm not sure we need both versions. In fact, there should be no difference between the two resulting voxel clouds. The cast there is just some block of space in a voxel cloud is only due to optimization reasons. The voxel cloud should be treated as endless, I think. Therefore I think we always want to shrink it to fit, unless the processing is more expensive than storing larger data. Also, what do you mean by _ compute the final bounding box from the volumes_ ? do you want to scan each source voxel cloud for volumes prior to the boolean operation? I think it is more expensive than calling `shrink_to_fit` at the end, isn't it?

janper: 2. I think I don't understand

janper: 3. we already had this - the constructor ordered inputs so that it was guaranteed that one point was min and the other was max. why is start + diagonal better when you need to check the diagonal to be all positive? it's easier to input two points and genrate proper min/max points. there can't be an error, whereas i'm not sure how would i treat a negative diagonal. 
i prefer returning to sorted points + we should implement more methods, like boolean operations (union, intersecion) of two bounding boxes and materialization of all 8 corners.

~btw, i just remembered where did my idea came from~ 
This used to be my reference but looking back at it, I'm not so sure I like it anymore. Still, have a look:
https://github.com/postspectacular/toxiclibs/blob/master/src.core/toxi/geom/AABB.java#L59

janper: 4. exactly, i've been thinking the same. unless we want to convert all the cooridnates into cartesian, which should also work.

janper: 5. okay

janper: i agree to remove the new separate

janper:
> I know we expect this, but the user maybe won't:
> booleans
> 
> While very cool, this is caused by the way we voxelize meshes. Shouldn't the voxel cloud be computed from the actual volume of the mesh instead of just the faces? (I imagine the math for this will be nontrivial)
> 
> Also, shouldn't converting a voxel cloud back to mesh create non-hollow meshes?

another PR solving exactly this is awaiting

janper:
> > shouldn't we only keep one of them, say the mutable one and turn it from one vs. one to one vs. many?
> 
> Interesting thought. I am not opposed, but let's do that only once we have a usecase... otherwise it's wasted work.

i can actually pick the logic from the immutable versions and put them into the mutable - this way we get that basically for free. i'm still worried about the ergonomy of its use - how would we pick the first one?

src/mesh/voxel_cloud.rs
yanchith: I am not absolutely sure what the correct behavior should be, but one more point towards using cartesian coords instead is that if we ever make `VoxelCloud` available as a variable type, the user will be able to create all kinds of voxel clouds and won't be able to not trigger this assert.

src/mesh/voxel_cloud.rs
yanchith: > The voxel cloud should be treated as endless, I think. Therefore I think we always want to shrink it to fit, unless the processing is more expensive than storing larger data.

Interesting thought. In this case, yes, we would want only one version and they would be equivalent. We will need to implement custom `Eq` and document this.

> Also, what do you mean by _ compute the final bounding box from the volumes_ ? do you want to scan each source voxel cloud for volumes prior to the boolean operation? I think it is more expensive than calling shrink_to_fit at the end, isn't it?

Yes. You are possibly right from the time perspective, but resizes fragment memory and the scan would still be linear. We can leave it for a `FIXME: @Optimization` though.

yanchith: > 2. I think I don't understand

I imagine we could have `BoundingBox::intersect(other: &BoundingBox) -> BoundingBox`, and similar for other boolean operations. Of course, it would behave a bit weird, so maybe we don't want to use the boolean operator nomenclature, but it seems clear enough to me.

The result would always be just one bounding box.

yanchith: > why is start + diagonal better when you need to check the diagonal to be all positive? it's easier to input two points and genrate proper min/max points. there can't be an error, whereas i'm not sure how would i treat a negative diagonal.

You are right that one is not worse than the other - they are exactly equivalent. Either way you have to _correctly generate_ the min/max points or origin+diagonal.

> i prefer returning to sorted points

Just to clarify, by sorted you mean constructed from the individual components, right? Because sorting is not sufficient.

> + we should implement more methods, like boolean operations (union, intersecion) of two bounding boxes and materialization of all 8 corners.

This is my point number 2.

yanchith: > exactly, i've been thinking the same. unless we want to convert all the cooridnates into cartesian, which should also work.

We could theoretically go either way, but having voxel-space (integer) bounding boxes will not force you to convert between the two coord spaces even more. I already think we have some subtle errors in the conversions we will have to hunt down later, so I'd rather not add more.

src/mesh/voxel_cloud.rs
yanchith: > where is the block_start then?

Tough. But since the cloud has no dimensions, we can perhaps afford to just pick something arbitrary, like the original block start. 

> how would it then perform to_mesh when we don't allow empty meshes?

`to_mesh` already returns an option for this exact reason. Well, for that and for weld too.

> also, if we don't allow empty meshes, why would we allow zero length voxel clouds?

You said earlier that two voxel clouds are equivalent if they have the same volumes, even if their bounding boxes or voxel map data is different. This same argument I think makes all the storage metadata question irrelevant. A different question would be, whether we want to allow a voxel cloud without a set volume. But that would be even more problematic.

Also. Maybe we should revisit allowing empty meshes :slightly_frowning_face:

src/mesh/voxel_cloud.rs
yanchith: > we'll only know after the boolean operation what is the actual size of the space that the voxels really occupy.

Apologies. You are right for intersection and difference. What I am suggesting is still possible for union.

yanchith: > i can keep storing the boundary extrems while performing the boolean operation and then perform resize but it is exaclty what shrink_to_fit does except the search for active voxels, which is just one pass through the block.

We can save it for a `FIXME: @Optimization`

src/mesh/voxel_cloud.rs
yanchith: yes, unless the compiler notices and removes the if for you.

yanchith:
>  i'm still worried about the ergonomy of its use - how would we pick the first one?

It is `self`.

src/mesh/voxel_cloud.rs
yanchith: Kk, I just realized the difference between resizing in intersection and union/difference. You are right.

src/mesh/voxel_cloud.rs
janper: Taking into account our conversation about the fact, that the block is just an implementation detail and the voxel clouds should appear infinite, shouldn't we entirely hide getters like `is_empty` (and check if the voxel cloud is zero in the `contains_voxels` instead) and `block_start`, `block_dimensions`, `block_end`?

src/mesh/voxel_cloud.rs
janper: let's do it with the cartesian coords. it is more general, allows for nice voxel operations and doesn't cost much

src/mesh/voxel_cloud.rs
yanchith: If we go that route, `is_empty` doesn't make any sense. I am a bit more reluctant to remove `block_start`, `block_dimensions` and `block_end`, but they probably can be supplemented by `volume_bounding_box` or `mesh_bounding_box`. Moreover, current `bounding_box` doesn't make sense as a `pub` method either and its name can be freed up, perhaps.

Just a note that in tests can still assert against the internal structure even when there isn't a getter.

janper:
Ready for another round of review. Please, start over as so much has changed.

Cargo.toml
```diff
@@ -33,6 +33,7 @@ tinyfiledialogs = "3.3.5"
 tobj = { version = "0.1.10", features = ["log"] }
 wgpu = { version = "0.4.0", features = ["vulkan"] }
 winit = "0.20.0-alpha5"
+num-traits = "0.2.11"
```
yanchith: Alphabetical please 😂

src/bounding_box.rs
```diff
@@ -0,0 +1,593 @@
...
+                },
+            ),
+        }
+    }
+
+    /// Creates a new bounding box from an iterator of points. The resulting
+    /// bounding box will encompass all the input points. The resulting bounding
+    /// box will be defined in the units of the input points. If the input list
+    /// is empty, a zero size box at the word origin will be created.
+    pub fn from_points<'a, I>(points: I) -> Option<Self>
```
yanchith: The doc comment and behavior disagree. Interestingly enough, I am torn between the two, both have merit.

I'd probably go for the way the implementation does it, not the doc comment, just because it is a bit more explicit - let's update the documentation.

src/bounding_box.rs
```diff
@@ -0,0 +1,593 @@
...
+    where
+        I: IntoIterator<Item = &'a Point3<T>> + Clone,
+    {
+        // Return None if there are no input points.
+        points.clone().into_iter().peekable().peek()?;
```
yanchith: I might be wrong, but I think both `into_iter` and `peekable` are redundant here. I think you want:

```rust
points.clone().next()?;
```

src/bounding_box.rs
```diff
@@ -0,0 +1,593 @@
...
+        }
+    }
+
+    /// Creates a new bounding box from an iterator of points. The resulting
+    /// bounding box will encompass all the input points. The resulting bounding
+    /// box will be defined in the units of the input points. If the input list
+    /// is empty, a zero size box at the word origin will be created.
+    pub fn from_points<'a, I>(points: I) -> Option<Self>
+    where
+        I: IntoIterator<Item = &'a Point3<T>> + Clone,
```
yanchith: After having some time to think on this, let's solve our optimization FIXMEs right away. Both points and bboxes are `Copy`, so let's change these iterators to require them owned. The need for intermediate allocations should go away.

src/bounding_box.rs
```diff
@@ -0,0 +1,593 @@
...
+    }
+
+    /// Creates a new bounding box encompassing all the input bounding boxes.
+    /// The resulting bounding box will be defined in the units of the input
+    /// bounding boxes.
+    pub fn union<'a, I>(bounding_boxes: I) -> Option<Self>
+    where
+        I: IntoIterator<Item = &'a BoundingBox<T>>,
+    {
+        // FIXME: @Optimization Remove the allocation, try from_fn or successors
```
yanchith: After having some time to think on this, let's solve our optimization FIXMEs right away. Both points and bboxes are `Copy`, so let's change these iterators to require them owned. The need for intermediate allocations should go away.

The callers can very easily add `.copied()` to their iterators.

src/bounding_box.rs
```diff
@@ -0,0 +1,593 @@
...
+
+        Some(BoundingBox {
+            minimum_point: Point3::new(min_x, min_y, min_z),
+            maximum_point: Point3::new(max_x, max_y, max_z),
+        })
+    }
+
+    /// Creates a new bounding box encompassing all the input bounding boxes.
+    /// The resulting bounding box will be defined in the units of the input
+    /// bounding boxes.
```
yanchith, line 110: Let's document when this returns `None` (When there are no bounding boxes)

src/bounding_box.rs
```diff
@@ -0,0 +1,593 @@
...
+        BoundingBox::from_points(points.iter())
+    }
+
+    /// Creates a new bounding box so that it encloses the block of space
+    /// common to all the input bounding boxes, including the current one.
+    ///
+    /// Returns None if there is no intersection.
+    pub fn intersection<'a, I>(bounding_boxes: I) -> Option<Self>
+    where
+        I: IntoIterator<Item = &'a BoundingBox<T>> + Clone,
```
yanchith: Same as above. Let's prevent future headaches by requiring owned bounding boxes.

src/bounding_box.rs
```diff
@@ -0,0 +1,593 @@
...
+    {
+        // Return None if there are no input bounding boxes.
+        bounding_boxes.clone().into_iter().peekable().peek()?;
```
yanchith: Same as above:

```rust
bounding_boxes().clone().next()?;
```

src/bounding_box.rs
```diff
@@ -0,0 +1,593 @@
...
+    }
+
+    /// Gets the minimum point of the bounding box. All the components of the
+    /// minimum point are the minimum values of the bounding box coordinates.
+    pub fn minimum_point(&self) -> Point3<T> {
+        self.minimum_point
+    }
+
+    /// Gets the minimum point of the bounding box. All the components of the
+    /// minimum point are the maximum values of the bounding box coordinates.
```
yanchith: 2x minumum point -> maximum point

src/bounding_box.rs
```diff
@@ -0,0 +1,593 @@
...
+            ((self.minimum_point.z as f32 + self.maximum_point.z as f32) / 2.0).round() as i32,
+        )
+    }
+
+    /// Computes the diagonal vector of the current bounding box.
+    /// # Warning
+    /// The diagonal dimensions are `maximum point coordinates - minimum point
+    /// coordinates + 1`. For a singularity bounding box, this returns diagonal
+    /// size 1, 1, 1 because that is the size of a single voxel.
+    pub fn diagonal(&self) -> Vector3<u32> {
```
yanchith: I see what you are trying to do with the `u32`, but I'd be fine if this returned `Vector3<i32>`. Also I can't convince myself that the behavior discussed in the warning section is correct. I don't think this should know anything about voxels, but even for voxels you might want a space that encompasses no voxels. I still have to read the rest of the PR and make up my mind based on how I see this used, but this certainly looks fishy in isolation.

src/bounding_box.rs
```diff
@@ -0,0 +1,593 @@
...
+    fn test_bounding_box_union_single_i32() {
+        let bb = BoundingBox::new(
+            &Point3::new(4_i32, 14_i32, 24_i32),
+            &Point3::new(2_i32, 22_i32, 33_i32),
+        );
+
+        let bb_union = BoundingBox::union(iter::once(&bb)).unwrap();
+
+        assert_eq!(bb_union.minimum_point, Point3::new(2_i32, 14_i32, 24_i32));
+        assert_eq!(bb_union.maximum_point, Point3::new(4_i32, 22_i32, 33_i32));
```
yanchith: You should be able to just `assert_eq!(bb, bb_union);`

src/bounding_box.rs
```diff
@@ -0,0 +1,593 @@
...
+        let bb_intersection = BoundingBox::intersection(iter::once(&bb)).unwrap();
+
+        assert_eq!(
+            bb_intersection.minimum_point,
+            Point3::new(2_i32, 14_i32, 24_i32)
+        );
+        assert_eq!(
+            bb_intersection.maximum_point,
+            Point3::new(4_i32, 22_i32, 33_i32)
+        );
```
yanchith: Should be able to just `assert_eq(bb, bb_intersection);`

src/bounding_box.rs
```diff
@@ -0,0 +1,593 @@
...
+            Point3::new(2_i32, 14_i32, 28_i32)
+        );
+        assert_eq!(
+            bb_intersection.maximum_point,
+            Point3::new(4_i32, 15_i32, 33_i32)
+        );
+    }
+
+    #[test]
+    fn test_bounding_box_intersection_two_non_intersecting_i32() {
```
yanchith, line 432: The bounding boxes in this test actually do intersect. Probably not testing what you meant.

src/bounding_box.rs
```diff
@@ -0,0 +1,593 @@
...
+
+    #[test]
+    fn test_bounding_box_diagonal_i32() {
+        let bb = BoundingBox::new(
+            &Point3::new(-1_i32, -1_i32, -1_i32),
+            &Point3::new(1_i32, 2_i32, 3_i32),
+        );
+
+        let diagonal = bb.diagonal();
+        let diagonal_correct = Vector3::new(3_u32, 4_u32, 5_u32);
```
yanchith: This just doesn't seem right

src/mesh/mod.rs
```diff
@@ -359,6 +360,12 @@ impl Mesh {
         &self.normals
     }
 
+    pub fn bounding_box(&self) -> Option<BoundingBox<f32>> {
```
yanchith: The `Mesh` is never empty (at least for our current representation). Let's do the `expect` here so we don't need to elsewhere.

src/interpreter_funcs/voxelize.rs
```diff
@@ -74,6 +74,13 @@ impl Func for FuncVoxelize {
                 }),
                 optional: false,
             },
+            ParamInfo {
+                name: "Fill closed volumes",
```
yanchith: Capitalization

src/interpreter_funcs/voxel_boolean_union.rs
```diff
@@ -0,0 +1,126 @@
...
+                name: "Grow",
+                refinement: ParamRefinement::Uint(UintParamRefinement {
+                    default_value: Some(1),
+                    min_value: None,
+                    max_value: None,
+                }),
+                optional: false,
+            },
+            ParamInfo {
+                name: "Fill closed volumes",
```
yanchith: Capitalization

src/interpreter_funcs/voxel_boolean_intersection.rs
```diff
@@ -0,0 +1,123 @@
...
+                name: "Grow",
+                refinement: ParamRefinement::Uint(UintParamRefinement {
+                    default_value: Some(1),
+                    min_value: None,
+                    max_value: None,
+                }),
+                optional: false,
+            },
+            ParamInfo {
+                name: "Fill closed volumes",
```
yanchith: Capitalization

src/interpreter_funcs/voxel_boolean_difference.rs
```diff
@@ -0,0 +1,126 @@
...
+                name: "Grow",
+                refinement: ParamRefinement::Uint(UintParamRefinement {
+                    default_value: Some(1),
+                    min_value: None,
+                    max_value: None,
+                }),
+                optional: false,
+            },
+            ParamInfo {
+                name: "Fill closed volumes",
```
yanchith: Capitalization

src/interpreter_funcs/voxel_boolean_difference.rs
```diff
@@ -0,0 +1,126 @@
...
+use nalgebra::Vector3;
+
+use crate::interpreter::{
+    BooleanParamRefinement, Float3ParamRefinement, Func, FuncError, FuncFlags, FuncInfo,
+    LogMessage, ParamInfo, ParamRefinement, Ty, UintParamRefinement, Value,
+};
+use crate::mesh::voxel_cloud::VoxelCloud;
+
+#[derive(Debug, PartialEq)]
+pub enum FuncBooleanDifferenceError {
```
yanchith, line 15: Doesn't this also need the `EmptyMesh` error case as intersecation has?

src/mesh/voxel_cloud.rs
```diff
@@ -228,6 +232,124 @@ impl VoxelCloud {
...
+                    }
+                    self.shrink_to_fit();
+                    // Return here because any other option needs to wipe the
+                    // current voxel cloud.
+                    return;
+                }
+            }
+        }
+
+        // If the two voxel clouds don't  or one of them is empty, then wipe
```
yanchith: "don't intersect"

yanchith (commented):
I am running out of steam, so this review will be split in two parts.

This is the first part. I managed to review everything except `voxel_cloud.rs`, which I will do tomorrow morning. Please don't act on this review yet, I'd like to review the whole thing before the ground shifts.

The only real issue I've found so far is that i32 bounding boxes have diagonals larger than what math intuition would suggest. I know why you did it this way, but I just disagree. The jury's still out though - I need to read the usage code in `voxel_cloud.rs` to decide. I think I may have a solution, stay tuned. Otherwise, the whole `bounding_box.rs` is beautiful ❤️

Cargo.toml
janper: never touched this file. it was `cargo add`

src/mesh/voxel_cloud.rs
```diff
@@ -567,64 +712,184 @@ impl VoxelCloud {
...
+                        (volume_start.y as f32 + volume_dimensions.x as f32) * voxel_dimensions.y,
+                        (volume_start.z as f32 + volume_dimensions.x as f32) * voxel_dimensions.z,
+                    ),
+                )
+            })
+    }
+
+    /// Returns the bounding box in voxel units of the current voxel cloud after
+    /// shrinking to fit just the nonempty voxels.
+    #[allow(dead_code)]
```
yanchith: This is not dead code

src/mesh/voxel_cloud.rs
```diff
@@ -335,10 +469,16 @@ impl VoxelCloud {
...
+    /// voxel-space units
+    pub fn resize_to_voxel_space_bounding_box(&mut self, bounding_box: &BoundingBox<i32>) {
+        self.resize(&bounding_box.minimum_point(), &bounding_box.diagonal());
+    }
+
     /// Computes boundaries of volumes contained in voxel cloud. Returns tuple
     /// (block_start, block_dimensions). For empty voxel clouds returns the
     /// original block start and zero block dimensions.
-    pub fn compute_volume_boundaries(&self) -> (Point3<i32>, Vector3<u32>) {
+    pub fn compute_volume_boundaries(&self) -> Option<(Point3<i32>, Vector3<u32>)> {
```
yanchith: Couldn't this be not `pub`? Don't forget to move down to private methods.

src/mesh/voxel_cloud.rs
```diff
@@ -228,6 +232,124 @@ impl VoxelCloud {
...
+                                &block_dimensions,
+                            )
+                            .expect("The current voxel map out of bounds"),
+                            self.voxel_dimensions,
+                        );
+
+                        // Perform boolean AND on voxel states of both voxel clouds.
+                        self.voxel_map[i] &= other
+                            .voxel_at_cartesian_coords(&cartesian_coords)
+                            .unwrap_or(false);
```
yanchith, line 274: Hmm, since we resized `self` to just the intersection, shouldn't this be possible to `expect`?

src/mesh/voxel_cloud.rs
```diff
@@ -228,6 +232,124 @@ impl VoxelCloud {
...
+
+    /// Computes boolean union (logical OR operation) of two Voxel clouds. The
+    /// current Voxel cloud will be mutated and resized to contain both input
+    /// Voxel clouds' volumes.
+    ///
+    /// # Warning
+    /// If the input Voxel clouds are far apart, the resulting voxel cloud may
+    /// be huge.
+    pub fn boolean_union(&mut self, other: &VoxelCloud) {
+        let valid_bounding_boxes: Vec<_> =
```
yanchith: This allocation won't be needed once we change `BoundingBox::union` iterator parameter to take bounding boxes by value.

src/mesh/voxel_cloud.rs
```diff
@@ -228,6 +232,124 @@ impl VoxelCloud {
...
+                    self.voxel_dimensions,
+                );
+                // If the other voxel cloud exists on the current absolute
+                // coordinate, perform boolean OR, otherwise don't change the
+                // existing voxel.
+                if let Some(true) = other.voxel_at_cartesian_coords(&cartesian_coords) {
+                    self.voxel_map[i] = true;
+                }
+            }
+            self.shrink_to_fit();
```
yanchith, line 330: IIRC, this (and only this) shrink to fit is not necessary, because we are doing boolean OR and the bounding box was constructed by ORing the original volume bounding boxes. Let's leave a comment here instead why we can omit shrink to fit.

src/mesh/voxel_cloud.rs
```diff
@@ -567,64 +712,184 @@ impl VoxelCloud {
...
+        while let Some(one_dimensional) = queue_to_process.pop_front() {
+            // Calculate the relative coord of the currently processed voxel.
+            // Will be needed to calculate its neighbors.
+            let coord = one_dimensional_to_relative_three_dimensional_coordinate(
+                one_dimensional,
+                &self.block_dimensions,
+            )
+            .expect("Coord out of bounds");
+            // Check all the neighbors
+            for neighbor_offset in neighbor_offsets.iter() {
```
yanchith: Nit: `for  neighbor_offset in &neighbor_offsets {`

yanchith (changes requested):
Overall looks great! Every inline comment I have is practically just a technicality. There is just one correctness issue, and that is the diagonal for integer bounding boxes.

Ok, so the artificial +1 for each component in the integer bounding box is a bug, that probably did help in some previous implementation I didn't see to cover up another bug. Good news is I don't think the cover-up is necessary for the code as presented in the last commit of this PR. I checked both the code, the tests, and ran the functionality, and fortunately fixing the issue does not break anything else (except for the test that expects the bad diagonal).

To give you a starting intuition, we should be thinking about (integer, but also float) bounding box diagonals as lengths, not indexes/positions. Think about computer science arrays and what the difference between an index and length is there - the same should apply here.

An integer bounding box with both minimum and maximum points at `[0, 0, 0]` has a diagonal vector `[0, 0, 0]`, which means there can be no voxels in it. A bounding box with minimum `[0, 0, 0]` and maximum `[0, 0, 1]` has a diagonal of `[0, 0, 1]`, and there still cannot exist any voxels in it, because it degenerates into a smaller dimension.

In `compute_volume_boundaries()` you correctly add +1 when computing `block_dimension`, converting index to length, therefore if there is at least one `true` voxel in the voxel cloud, the bounding box computed in `volume_bounding_box()` will have a diagonal vector of `[1, 1, 1]`. This should mean that there is no reason for us to add +1 artificially in the diagonal vector computation - there is no way the diagonal would have a zero component for voxel clouds containing at least one voxel.

On another point: we talked in person about overriding `Eq` for voxel clouds. Let's not do that yet unless we need it for something, I think we still need to give it some more thought.

src/bounding_box.rs
janper:
```rust 
points.clone().into_iter().next()?;
```

src/mesh/voxel_cloud.rs
janper: I'm worried we can accidentally hit an out-of-bounds voxel due to possible different voxel dimensions. not sure, just a gut feeling it can happen and this autofix here seems good enough

src/bounding_box.rs
yanchith: Actually neither :frowning_face: . I just realized we are cloning collections left and right. Stay tuned. There has to be a better way.

yanchith: Fortunately, for your case (at least here), there is a solution:

1) Remove the `Clone` bound
2) Use:
```rust
    let mut points_iter = points.into_iter().peekable();
    points_iter.peek()?;
    // ... later use points_iter
```

I'll add a trello task to hunt down the clones all over the project. I know you have the trick from me. I just didn't realize we are not cloning the iterator, but the original source.

src/bounding_box.rs
yanchith: See https://github.com/subdgtl/HURBAN-Selector/pull/112/files#r368123560

src/bounding_box.rs
yanchith: https://github.com/subdgtl/HURBAN-Selector/pull/117

yanchith: Anyway, the guideline is that you want to use `IntoIterator` over `Iterator` as a bound, because it is more generic. But if you need to clone the iterator, you need `Iterator + Clone`, and shouldn't use `IntoIterator + Clone`, because that could accidentally clone the collection. We had a couple of these bounds incorrect in the codebase, but fortunately we were not cloning collections, just iterators.

src/mesh/voxel_cloud.rs
janper: here again, are we sure the voxel cloud block will be tight even if the other voxel clouds have different voxel dimensions? maybe it does, i'm just not sure

src/interpreter_funcs/voxel_boolean_difference.rs
janper: Added empty voxel cloud error / check everywhere it seemed to be relevant for better UI error reporting

src/mesh/voxel_cloud.rs
yanchith: You are right. I forgot that the coords were cartesian. 👍

src/mesh/voxel_cloud.rs
yanchith: np, let's leave it like this

src/lib.rs
```diff
@@ -398,13 +398,14 @@ impl CameraInterpolation {
...
+            .map(|mesh| mesh.bounding_box())
             .collect();
 
-        let (target_origin, target_radius) = match BoundingBox::union(bounding_boxes.iter()) {
-            Some(bounding_box) => (bounding_box.center(), bounding_box.diagonal().norm() / 2.0),
-            None => (Point3::origin(), 1.0),
-        };
+        let (target_origin, target_radius) =
+            match BoundingBox::union(bounding_boxes.iter().copied()) {
+                Some(bounding_box) => (bounding_box.center(), bounding_box.diagonal().norm() / 2.0),
```
yanchith: Hmm, what's with this `norm` thing?

yanchith (dismissed):
Thank you ❤️

yanchith (approved):

PR #111, opened 2020-01-05
Add Windows installer

ondrowan:
To try this, follow the steps from main README / README of installer. Most of the files are just copied from [innodependencyinstaller](https://github.com/domgho/innodependencyinstaller). Redistributables are contained in installer itself. I tried it on my computer and it installed everything properly, but I'd prefer if we tired it also on some clean installation of Windows to make sure everything works as intended. The only thing that is missing for now is icon, but that can be added later.

PR #114, opened 2020-01-10
Use Display impls of interpreter errors for error tooltips in gui

yanchith:
This is just the initial (low-hanging fruitish) part of the story and is not very satisfactory, but is imo better than status quo.

The problem is that the reported errors are just pure `Display` impls, which do miss a lot of context information because of the denormalized nature of our program representation. We will probably later transition away from bare `Display` impls for and into a dedicated context based error string building, where the interpreter runtime will substitute proper names for the provided IDs.

Anyway, please have a look.


janper (changes requested):
I think it's visually okay. in both, the dark and the funky theme. 
Two notes though:

1. It's redundant (and horribly geeky) to yield the entire message like `Function >func 9003> errored with "Error message" on stmt 1`. It's obvious which function errored because it's highlighted. It's obvious it has errored because it's red. The message itself is important. The fact it errored due to `stmt 1` is also important but it's so cryptic that even I had to take a couple of moments to realize its meaning. So let's omit also this information until we have an option to use the description string of the statement causing the error. For now I'd only keep the message in quotes, without the quotes.
2. _This doesn't need to be a part of this PR._ The full version of the error message as it is now seems to belong into some general log console. We currently don't have any and you explained to me that @ondrowan spent some time removing the default text console but I'm strongly for returning it even in the release build. That is our only way of knowing what happened if something happens. It's not only good for proper error messages but also for panic messages. I know we eventually shouldn't have any, but that would mean removing also all expects. I doubt we'll ever do that and also I doubt we'll ever be sure we don't panic anymore. Even while testing this I encountered a panic situation I didn't expect (when trying to voxelize a box sized 0, 0, 0 - which btw. I think we shouldn't allow creating as it's impossible to transform it to anything else than a point, which we don't support).

yanchith:
@janper

I totally agree with your point number 1, but it is not that easy. I don't think we can do what you suggest cheaply, without doing the full-blown solution I tried to avoid. I'll try to explain.

By the time the message gets to be displayed, all structure is gone, it is just text. The "message in quotes" is the message of a runtime error originating from a func. But there are other errors you would want to display (like dynamic type errors, which do also happen). Doing what you suggest would mean disregarding those. I personally think they are valuable.

I see 3 alternatives:

1.) If this PR isn't valuable as-is, close it and do a full-blown solution.
2.) Display just runtime errors and do not give any error feedback on other types of errors. Do a full-blown solution later. This is a compromise, but I don't like it very much.
3.) Merge functionality as is, do a full-blown solution later.

I prefer them in the order of 3 > 1 > 2.

Regarding your point 2: 

Who are these raw logs for? The developer or the user? I would argue the user should never see those. The developer can already find them in the logs directory on their machine. On Windows, that's in the local appdata, in the HURBAN Selector directory.

Also, you should understand, that there are 2 orthogonal concepts:

- debug vs release build (they differ mostly in the amount of applied compiler optimizations)
- internal vs dist build (whether the build is for developers or users)

I still think it is correct that the console is hidden in the dist build. We can unhide it for all internal builds, both debug and release.

As for never panicking - that is the noble goal no program ever reaches, but all programs should strive for.

janper:
> I totally agree with your point number 1, but it is not that easy. I don't think we can do what you suggest cheaply, without doing the full-blown solution I tried to avoid. I'll try to explain.
> 
> By the time the message gets to be displayed, all structure is gone, it is just text. The "message in quotes" is the message of a runtime error originating from a func. But there are other errors you would want to display (like dynamic type errors, which do also happen). Doing what you suggest would mean disregarding those. I personally think they are valuable.

Can we at least change the wording? Say `Function 9003 errored with "Error message" on input 1` and then keep it as is?

janper:
> Who are these raw logs for? The developer or the user? I would argue the user should never see those.

How will the users report runtime errors? Because they will... Shall we then navigate them to local appdata? I'm fine if it is so. 

Also, I should really check the installer. Does it add the program to the start menu? Does it have its own folder? If so, I'd put there also a link to the log file or log folder. We'll have to be able to respond to user error reports.

yanchith:
> Does it add the program to the start menu? Does it have its own folder? If so, I'd put there also a link to the log file or log folder. We'll have to be able to respond to user error reports.

Please ask these questions in #111 😂

yanchith:
> How will the users report runtime errors? Because they will... Shall we then navigate them to local appdata? I'm fine if it is so.

This is the cheapest way to do it. There would be better ways if this were a much larger company and much larger software.

yanchith:
> 
> 
> > I totally agree with your point number 1, but it is not that easy. I don't think we can do what you suggest cheaply, without doing the full-blown solution I tried to avoid. I'll try to explain.
> > By the time the message gets to be displayed, all structure is gone, it is just text. The "message in quotes" is the message of a runtime error originating from a func. But there are other errors you would want to display (like dynamic type errors, which do also happen). Doing what you suggest would mean disregarding those. I personally think they are valuable.
> 
> Can we at least change the wording? Say `Function 9003 errored with "Error message" on input 1` and then keep it as is?

Yes we can. The original error message was aimed at the developer, but we can repurpose it temporarily. Let's do that, if you agree.

janper (dismissed):

janper (approved):
Is the FIXME note descriptive enough? Will it be possible to understand it a couple of months from now?

yanchith:
@janper 

> Is the FIXME note descriptive enough? Will it be possible to understand it a couple of months from now?

Sorry, I totally missed your comment. If you have any suggestions, let me know and I'll add it to the FIXME.

PR #115, opened 2020-01-14
Implement local transformation option in the UI operation

janper:
Implement an option to scale and rotate around object's local center.

### DIFF
--- a/src/interpreter_funcs/transform.rs
+++ b/src/interpreter_funcs/transform.rs
@@ -77,7 +77,7 @@ impl Func for FuncTransform {
                 optional: false,
             },
             ParamInfo {
-                name: "Rotate and scale around object center",
+                name: "Transform around object center",
                 refinement: ParamRefinement::Boolean(BooleanParamRefinement {
                     default_value: true,
                 }),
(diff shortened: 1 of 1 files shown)


src/interpreter_funcs/transform.rs
```diff
@@ -91,25 +100,56 @@ impl Func for FuncTransform {
...
+                let n1 = translation_to_origin.transform_vector(n);
+                let n2 = user_transformation.transform_vector(&n1);
+                final_translation.transform_vector(&n2)
+            });
+
+            Mesh::from_faces_with_vertices_and_normals(
+                mesh.faces().iter().copied(),
+                vertices_iter,
+                normals_iter,
+            )
```
janper, line 140: I realize that it would be enough to return only `vertices_iter` and `normals_iter` but `no two closures, even if identical, have the same type`. So it seemed easier to return the mesh itself.

src/interpreter_funcs/transform.rs
```diff
@@ -74,6 +76,13 @@ impl Func for FuncTransform {
                 }),
                 optional: false,
             },
+            ParamInfo {
+                name: "Rotate and scale around object center",
```
yanchith: Maybe a shorter name?

src/interpreter_funcs/transform.rs
```diff
@@ -91,25 +100,56 @@ impl Func for FuncTransform {
...
             rotate[1].to_radians(),
             rotate[2].to_radians(),
         );
-        let scaling = Matrix4::new_nonuniform_scaling(&scale);
+        let user_scaling = Matrix4::new_nonuniform_scaling(&scale);
+        let user_translation = Matrix4::new_translation(&translate);
 
-        let t = translation * Matrix4::from(rotation) * scaling;
+        let value = if transform_around_local_center {
+            // Move to the origin, scale and rotate, then mve back and finally
```
yanchith: mve -> move

yanchith (dismissed):
👍

PR #116, opened 2020-01-14
Add ability to screenshot the viewport

yanchith:
The functionality is triggered by either the P key, or by clicking on
the "screenshot" button in the "viewport" window. That brings up a
modal popup allowing the user to set the screenshot resolution (either
in absolute numbers, or scale) and whether the background should be
transparent. Screenshots are saved to the platform specific "pictures"
directory.

The renderer had undergone a serious refactoring to make this
happen. The scene is now always rendered to an intermediate texture
and just then blitted to the backbuffer. This will also allow us to do
postprocessing effects later on. The UI is still drawn directly to the
backbuffer.

Creation of offscreen (as opposed to primary) render
targets (i.e. collection of textures and buffers needed for rendering)
is now possible in the renderer. Each time a screenshot is supposed to
be captured, a new offscreen render target is created, rendered to,
and has the sceenshot picture extracted from it before it is
destroyed. Offscreen render targets are currently multisampled with
the exact same number of samples that the primary render target is
sampled with. This restriction can be lifted later.

Currently not working:

- Vulkan on macOS hangs when trying to render a screenshot (but metal
works)
- Vulkan on Linux segfaults when trying to read the buffer containing
the screenshot if the screenshot has any realistic
dimensions (i.e. greater than 200x200)


src/renderer/scene_renderer.rs
```diff
@@ -672,11 +691,6 @@ impl SceneRenderer {
             }),
         });
 
-        // This needs to be set for vulkan, oterwise the validation
-        // layers complain about the stencil reference not being
-        // set... Not sure if this is a bug or not.
-        rpass.set_stencil_reference(0);
```
yanchith, line 678: This doesn't seem to be a problem anymore on the current version of wgpu-rs.

janper (changes requested):
_I had to write this entire review twice, because when I tried to submit it for the first time I lost the internet connection and it all went away #^%&&#%^#%@#_

Thank you for this feature - it was high on the feature list and this implementation works incredibly well! I managed to take an 8K screenshot on my XPS and it didn't even take too long (about 2s or so). Even the quality is great and there are no artifacts on the verge of transparency.

I'd like to add a warning to the modal window, saying something like "Attempting to take large screenshots may cause program crash and loss of any unsaved work" - or anything similar that sounds natural in English.

Also I'd love to see some visual feedback when the screenshot is successfully created. Any of the following options works for me (let's do the easiest one):

1. Another modal window with an obvious green check mark or green OK that appears after success saying something like "Screenshot saved as C:\users\nicolascage\Pictures\hurban_selector-2020-01-16-07-45-35.png". This window should automatically disappear after some time or after mouse move.
2. Automatically opening the image in the system default app (this is what virtual PDF printers do).
3. Automatically opening the containing folder in system default file browser and highlighting the newly created file (this is what web browsers do after downloading a file).
4. Version 1, just not automatically disappearing. Instead with buttons for closing the modal window and for version 2 and 3.

yanchith:
> there are no artifacts on the verge of transparency.

We currently don't solve transparency at all, just the background is cleared to transparent black. I believe if we ever do transparent objects in the viewport, there will be artifacts. We will solve that once we need to.

> I'd like to add a warning to the modal window, saying something like "Attempting to take large screenshots may cause program crash and loss of any unsaved work" - or anything similar that sounds natural in English.

Ok.

Of the rest, 1 and 4 should cheap to implement. 2 and 3 are a nontrivial amount of work even if we supported this on just windows. I'll go for 1.

I will get to implementing the requested changes earliest on Tuesday.

janper:
> We currently don't solve transparency at all, just the background is cleared to transparent black.

Of course we do! We do have a transparent background and antialiased edges. That's where the artifacts usually occur. But we are good, cheers!

janper:
> I'll go for 1

Agree

janper:
Is the German spelling intentional? :)) german spelling

yanchith:
Yes :))

Still working on the notifications

yanchith:
Screenshots now produce notifications. Notifications have autoscroll (in case they overflow their window), and ttl of 5 seconds.

yanchith:
Ready for another review

janper (dismissed):

janper:
Lovely. Just thinking if the notification box needs to be that tall. Up to your consideration though.

yanchith:
Thanks, I'll make it a bit smaller and merge.

PR #118, opened 2020-01-19
Implement resampling, moving, rotating and scaling of voxel clouds

janper:
This commit:
- implements resampling of an existing voxel cloud to a new one with different voxel dimensions.
- implements transformation (move, rotate volume locally, scale volume locally) and resampling (into cubic voxels) of a voxel cloud into a new one.
- renames coordinate conversion function to follow the same convention.
- implements a UI operation for voxelizing and consequent transformation of the voxel cloud. This operation serves fr testing only and should be (along with some other operations) removed before release.

**Question:** Should we make tests and if so, what kind?

_**Note:** I wasn't able to implement the transformation for non-cubic voxels (with varying width, height and depth). Everything worked until a specific (yet common) situation occurred: the volume center was not at the world origin, the target voxel dimensions were varying, rotation angles were non-zero. Scaling and rotation seemed to be correct, the resampling as well but the final position of the mesh was significantly off. I suspect the varying voxel dimensions and the rotation to cause a non-trivial calculation of the translation offset. My solution was to forbid varying voxel dimensions and restrict the functionality to cubic voxels. I'm quite convinced this is sufficient mainly because the varying voxels in general are not very common and only make sense as an optimization if the voxelized object is less detailed in one of the orthogonal directions (say the shape is an extrusion in the vertical direction, then the resolution in the Z direction can be lower and thus the respective voxel dimension larger). Once we allow for free rotation of the voxel cloud, the lower resolution in one direction doesn't apply anymore and we should strive for cubic voxels with side equal to the lowest of the source voxel dimensions._


yanchith:
> Scaling and rotation seemed to be correct, the resampling as well but the final position of the mesh was significantly off.

I'll investigate.

yanchith:
I agree with you that dropping noncubic voxels would be cool. If anything else there would be less complexity.

src/mesh/voxel_cloud.rs
```diff
@@ -141,6 +142,169 @@ impl VoxelCloud {
...
+        voxel_dimensions: &Vector3<f32>,
+    ) -> Option<Self> {
+        assert!(
+            voxel_dimensions.x > 0.0 && voxel_dimensions.y > 0.0 && voxel_dimensions.z > 0.0,
+            "One mor more voxel dimensions is below or equal to zero"
+        );
+
+        source_voxel_cloud
+            .mesh_volume_bounding_box_cartesian()
+            .map(|current_vc_bounding_box| {
```
yanchith: `source_vc_bounding_box`?

src/mesh/voxel_cloud.rs
```diff
@@ -141,6 +142,169 @@ impl VoxelCloud {
         voxel_cloud
     }
 
+    /// Creates a new voxel cloud with arbitrary voxel dimensions from another
+    /// voxel cloud.
+    /// # Panics
```
yanchith, line 148: Missing newline before title

src/mesh/voxel_cloud.rs
```diff
@@ -141,6 +142,169 @@ impl VoxelCloud {
...
+                        .unwrap_or(false);
+                }
+
+                target_voxel_cloud
+            })
+    }
+
+    /// Creates a new voxel cloud from another voxel cloud transformed (scaled,
+    /// rotated, moved) in a cartesian space.
+    /// # Panics
```
yanchith, line 202: Missing newline before title

src/mesh/voxel_cloud.rs
```diff
@@ -141,6 +142,169 @@ impl VoxelCloud {
...
+    /// Panics if any of the voxel dimensions is below or equal to zero.
+    pub fn from_voxel_cloud_transformed(
+        source_voxel_cloud: &VoxelCloud,
+        voxel_dimension: f32,
+        cartesian_translation: &Vector3<f32>,
+        rotation_degrees: &(f32, f32, f32),
```
yanchith: I'd send radians here unless you strongly object. I'd say we use radians everywhere internally, and degrees on the UI. Also, since we are embracing `nalgebra`, we can perhaps take `&Rotation` instead?

src/mesh/voxel_cloud.rs
```diff
@@ -141,6 +142,169 @@ impl VoxelCloud {
...
+        source_voxel_cloud
+            .mesh_volume_bounding_box_cartesian()
+            .map(|current_vc_bounding_box| {
+                // FIXME: Heterogenous voxels require non-trivial compensation
+                // of final voxel cloud position after rotating if the volume
+                // equilibrium is not at the world origin.
+                let voxel_dimensions =
+                    Vector3::new(voxel_dimension, voxel_dimension, voxel_dimension);
+                // Make a bounding box of the source bounding box's mesh volume.
+                // This will be the volume to be scanned for any voxels.
```
yanchith: This comment doesn't seem to talk about the following lines of code, but rather the previous. Same for the one in the other method.

src/mesh/voxel_cloud.rs
```diff
@@ -141,6 +142,169 @@ impl VoxelCloud {
...
+                    &transformed_bounding_box,
+                    &voxel_dimensions,
+                );
+
+                // Transform the target voxels inverse to the user
+                // transformation so that it is possible to sample the source
+                // voxel cloud.
+                let reversed_user_transformation = compound_transformation
+                    .pseudo_inverse(f32::EPSILON)
+                    .expect("No pseudo inverse matrix");
```
yanchith: Hmm, I don't pretend to understand the math behind pseudo inverse or when exactly it can fail (negative `eps` is one case, but that's alright), but don't we want to return `None`, just in case?

Or alternatively, we can still `expect` and handle this only if we run into the panic.

src/mesh/voxel_cloud.rs
```diff
@@ -141,6 +142,169 @@ impl VoxelCloud {
...
+
+                    // Set the new voxel state according to a sampled value from
+                    // the source voxel cloud.
+                    *voxel = source_voxel_cloud
+                        .voxel_at_cartesian_coords(&transformed_voxel_center_cartesian)
+                        .unwrap_or(false);
+                }
+
+                let cartesian_final_translation_vector =
+                    (-1.0 * vector_to_origin) + cartesian_translation;
```
yanchith: `-vector_to_origin`?

src/mesh/voxel_cloud.rs
```diff
@@ -141,6 +142,169 @@ impl VoxelCloud {
...
+        }
+
+        assert!(
+            voxel_dimension > 0.0,
+            "Voxel dimension is below or equal to zero"
+        );
+
+        source_voxel_cloud
+            .mesh_volume_bounding_box_cartesian()
+            .map(|current_vc_bounding_box| {
```
yanchith: `source_vc_bounding_box`?

src/mesh/voxel_cloud.rs
```diff
@@ -141,6 +142,169 @@ impl VoxelCloud {
...
+                // FIXME: Heterogenous voxels require non-trivial compensation
+                // of final voxel cloud position after rotating if the volume
+                // equilibrium is not at the world origin.
```
yanchith: I'd be alright with having just cubic voxels. Let's leave the FIXME for now. We can disallow noncubic voxels in another PR.

src/mesh/voxel_cloud.rs
```diff
@@ -141,6 +142,169 @@ impl VoxelCloud {
...
+    /// Creates a new voxel cloud from another voxel cloud transformed (scaled,
+    /// rotated, moved) in a cartesian space.
+    /// # Panics
+    /// Panics if any of the voxel dimensions is below or equal to zero.
+    pub fn from_voxel_cloud_transformed(
+        source_voxel_cloud: &VoxelCloud,
+        voxel_dimension: f32,
+        cartesian_translation: &Vector3<f32>,
+        rotation_degrees: &(f32, f32, f32),
+        scale: &(f32, f32, f32),
```
yanchith: I know this might sound arbitrary, but let `scale`be `(f32, f32, f32)`. The type is quite small. It's more like we made an exception for the nalgebra types.

Or even better, let it be a `&Vector`. You construct one anyway later in the code.

yanchith (commented):
In general looks good :+1: 

Except the minor technicalities I expressed inline, I think we can add a couple of snapshot tests on voxel clouds. We can maybe use our own software to visually verify them?

Also, it don't we also want to add nonlocal transforms?

janper:
> Also, it don't we also want to add nonlocal transforms?

Short answer: No
Long answer: Most of the 3D software let you define a pivot point for each separate mesh, which becomes part of the mesh as one of its many descriptors. Others ask you ad hoc. We don't suppoert any of these options and rather expect the world origin (which is also the local model origin at the beginning) to be the pivot. I can think of only a single scenario when this would be an useful setup: when you are importing a model that consists of more separate meshes / files. This way you can rotate, move or scale the separate models the same way and their mutual spatial configuration remains the same. Also if a part of the model is in different units (say inches vs mm) and you need to scale it so that it also fits in the right place. Because these rare occasions might occur, I've left an option to use the global origin in the transform operation. For most case we, however, follow the usual practice of simple 3D tools (3D printing slicers or viewers), when you transform the object around its center. 
When it comes to transforming a voxel cloud, the situation is different. We won't expose such function in the UI and the only reason we have it so far is the alignment of two models. There we require transformations around the object center because using the world origin will bring certain level unpredictability to the search (I'm planning for genetic algorithms).

This, however, shows we lack some other features:

- grouping from UI
- in-viewport transformation gizmo
- automatical snapping to the ground plane / world origin

src/mesh/voxel_cloud.rs
janper: I'd keep the non-cubic voxel wherever we can and only drop them here.

janper: i'd keep it as is - cubic voxels here, non-cubic anywhere else

janper:
Ready for another review.
All snapshot tests were visually tested with the very same parameters in our software.
I also found an error and made a couple of optimizations. Although they could probably have been done better, I decided to leave just a FIXME because it doesn't seem to be worth the trouble. Also, I'm not sure if there would be a better way anyway.

janper:
no idea why the CI fails though...

janper:
A massive refactor is underway with all the change requests from this PR incorporated. It would be a waste of time to keep reviewing this PR.

PR #119, opened 2020-01-19
Add basic saving and loading of project files

ondrowan:
This contains just bare save and load. There's a ton of features that
will be done later:

* Autosave
* Prompt if project is unsaved upon exit
* Errors while saving / loading of project
* Keyboard shortcuts
* Save/load being handled in separate thread


janper:
it seems to work, however we never spoke about the GUI. first of all, it's currently interfering with the other GUI elements. Second - do we want to have the top menu bar? shouldn't we stick to floating windows with various functionalities? i'm open to discussion.

image

ondrowan:
This is exactly why I just implemented the menu bar without moving anything. I wasn't sure if this is the way we want to go in the end. It however made sense for me because it's standard and it'd make sense to move more functionality there (save as, autosave, some other options). Floating windows would probably take way too much space in the end and IMO it's not something that people would be used to (not sure about 3D software though).

Cargo.toml
```diff
@@ -27,12 +27,13 @@ imgui-winit-support = { version = "0.2.0", default-features = false, features =
...
 nalgebra = { version = "0.19.0", features = ["serde-serialize"] }
 png = "0.15.0"
-serde = "1.0.102"
+serde = { version = "1.0.102", features = ["derive", "rc"] }
 smallvec = "0.6.10"
 tinyfiledialogs = "3.3.5"
 tobj = { version = "0.1.10", features = ["log"] }
 wgpu = { version = "0.4.0", features = ["vulkan"] }
 winit = "0.20.0-alpha5"
+ron = "0.5.1"
```
yanchith, line 39: you rebel :joy_cat:

src/interpreter/ast.rs
```diff
@@ -1,11 +1,13 @@
 use std::fmt;
 use std::sync::Arc;
 
+use serde::{Deserialize, Serialize};
```
yanchith: Just a matter of style, but elsewhere we don't import the derives, and just derive `serde::Serialize`. Seems a bit cleaner to me.

src/ui.rs
```diff
@@ -612,6 +617,49 @@ impl<'a> UiFrame<'a> {
         }
     }
 
+    pub fn draw_menu_bar(&self, project_path: Option<String>) -> MainMenuBarStatus {
```
yanchith: What's wrong with `Option<&str>`? This forces a per-frame allocation.

src/project.rs
```diff
@@ -0,0 +1,29 @@
+use std::fs::File;
+use std::io::prelude::*;
+use std::io::BufReader;
+use std::path::Path;
+
+use ron;
+use serde::{Deserialize, Serialize};
```
yanchith: Same as above, I'd just use this directly.

src/ui.rs
```diff
@@ -43,6 +43,11 @@ struct ConsoleState {
     message_count: usize,
 }
 
+pub struct MainMenuBarStatus {
```
yanchith: :+1:  we should probably be using these return types instead of `&mut` parameters for the gui when we want multiple return values.

Cargo.toml
```diff
@@ -27,12 +27,13 @@ imgui-winit-support = { version = "0.2.0", default-features = false, features =
 log = { version = "0.4.8" }
 nalgebra = { version = "0.19.0", features = ["serde-serialize"] }
 png = "0.15.0"
-serde = "1.0.102"
+serde = { version = "1.0.102", features = ["derive", "rc"] }
```
yanchith: Meh. The `rc` feature may come to bite us later. I suggest the following:
- Remove the `rc` feature
- Change `LitExpr::String` to contain just `String`, not `Arc<String>`
- Fix the resulting compile errors by:
  - When evaluating a `LitExpr` to a `Value`, just clone the string in `interpreter/mod.rs`.
  - Remove the completely needless 3 `Arc::new` calls in `ui.rs`

Btw even if the docs say otherwise, the `derive` features seems to be a enabled by default. We already have some `derive(serde::Serialize)` in the codebase and they work just fine.

yanchith (changes requested):
I am not against the menu bar, but @janper should make the decision. If we do keep the menu bar, making the UI elements non-interfering would be nice.

See inline for code review.

yanchith:
Also, can we export formatted savefiles? Should be good for our debugging and the file size will be ok anyway.

EDIT: would also be nice to have struct names in the output. You can achieve that by using the `Serializer`. (https://docs.rs/ron/0.5.1/ron/ser/struct.Serializer.html)

ondrowan:
What do you mean by struct names not being in output? I can see them in mine.

src/interpreter/value.rs
```diff
@@ -50,7 +50,7 @@ pub enum Value {
     Float(f32),
     Float2([f32; 2]),
     Float3([f32; 3]),
-    String(Arc<String>),
+    String(String),
```
yanchith: I'd keep this one `Arc<String>`

yanchith:
> What do you mean by struct names not being in output? I can see them in mine.

In the link I sent (https://docs.rs/ron/0.5.1/ron/ser/struct.Serializer.html#method.new), there is a `struct_names` parameter to `Serializer::new`. It will serialize the struct names to the `ron` file. They are optional, but having them should make the save file more human readable.

yanchith:
@ondrowan as we talked on chat, I did the `Option<String>` -> `Option<&str>`. Have a look at the last commit.

janper:
For the time being, let's add the `save`, `save as` and `open` buttons to the **Viewport settings** window. Also let's rename the window to something else, say **Utilities** or something generic.

ondrowan:
@janper It is moved now.

src/lib.rs
```diff
@@ -255,11 +258,40 @@ pub fn init_and_run(options: Options) -> ! {
...
+                }
+
+                if let Some(open_path) = utilities_status.open_path {
+                    let project = project::open(&open_path);
+
+                    scene_meshes.clear();
+
+                    for (_, gpu_mesh_handle) in scene_gpu_mesh_handles.drain() {
+                        renderer.remove_scene_mesh(gpu_mesh_handle);
+                    }
```
yanchith, line 328: Since this was last reviewed, there has been more work added that needs to be done here:

- the scene bounding box needs to be recmputed
- the ground plane needs to be recomputed (and reinserted to the renderer)

The snippet to do this is pretty much:

```rust
                            scene_bounding_box = BoundingBox::union(
                                scene_meshes.values().map(|mesh| mesh.bounding_box()),
                            )
                            .unwrap_or_else(BoundingBox::unit);

                            ground_plane_mesh = compute_ground_plane_mesh(&scene_bounding_box);
                            ground_plane_mesh_bounding_box = ground_plane_mesh.bounding_box();
                            renderer.remove_scene_mesh(
                                ground_plane_gpu_mesh_handle
                                    .take()
                                    .expect("Ground plane must always be present"),
                            );
                            ground_plane_gpu_mesh_handle = Some(
                                renderer
                                    .add_scene_mesh(&GpuMesh::from_mesh(&ground_plane_mesh), true)
                                    .expect("Failed to add ground plane mesh"),
                            );
```

It is already present around 4 times in lib.rs (on the `master` branch), it happens when we poll a response from the interpreter thread.

src/ui.rs
```diff
@@ -433,22 +438,26 @@ impl<'a> UiFrame<'a> {
...
         let ui = &self.imgui_ui;
+        let mut status = UtilitiesStatus {
+            reset_viewport: false,
+            save_path: None,
+            open_path: None,
+        };
 
         const VIEWPORT_WINDOW_WIDTH: f32 = 150.0;
-        const VIEWPORT_WINDOW_HEIGHT: f32 = 170.0;
+        const VIEWPORT_WINDOW_HEIGHT: f32 = 210.0;
```
yanchith: Isn't this now called the utilities window?

yanchith (changes requested):
Almost there, there's been some changes in `master` that now mean that more work needs to be done in `lib.rs` though.

yanchith (approved):
LGTM. @janper ?

janper:
> LGTM. @janper ?
@yanchith, not LGBTI but tolerant.

janper (approved):
The menu looks good now, I'm okay if it merges.

However, I've just checked the `.hs` file extension - it's already used by **Haskell script**. What if we used `.hurban` instead?

ondrowan:
> The menu looks good now, I'm okay if it merges.
> 
> However, I've just checked the `.hs` file extension - it's already used by **Haskell script**. What if we used `.hurban` instead?

Will do it in other PR.

PR #121, opened 2020-01-26
Scalar field

janper:
**This is a huge refactor, which replaces PR #118 and implements all change requests raised there.**

The original **voxel cloud** has been entirely replaced by a more generic **scalar field**, which in concept is the same but rather than holding binary boolean values it can contain any bounded signed numerical values. This fact changes several features of the original voxel cloud, such as:
- the grow and fill functions are replaced by a single computation of a distance field (each voxel's value represents its distance from the surface of a volume)
- meshing of a scalar field can be done with any interval of values
- it will be easy to compute deviations of two scalar fields by just requesting a distance value of one field where is a volume voxel in the other - this is the main reason we need the scalar field
- it is possible to write and read a scalar field in various ways, i.e. as a force field and visualize it (not implemented yet)

The scalar field and all its functions now prefer absolute voxel coordinates to relative. This is because the scalar field should seem infinite as much as possible to the caller. And it also looks more tidy.

_**Question:** When requesting a value of a voxel I was considering returning (and ideally also storing in the_ `values: Vec<Option<T>>` _) an option, where None would mean there is no voxel even if the coordinate is within the bounds. Currently there is a single specific value reserved for an empty voxel. Having an Option, None would be returned in both cases, when the voxel is outside of the bounds and when it is empty. This would make the bounds even more irrelevant. Also there would be no value reserved for an empty voxel. Is it smart to store_ `values: Vec<Option<T>>` _?_

This PR also implements an **interval**, which is similar to what we called a domain in previous projects.

Some improvements were made to the **bounding box** as well.

_Sorry for not squashing the commits into a single one but git diff failed like this so I just rebased the branch onto Master._
```
error: patch failed: ci/nightly.yml:32
error: ci/nightly.yml: patch does not apply
error: patch failed: ci/pr_and_master.yml:29
error: ci/pr_and_master.yml: patch does not apply
error: patch failed: src/interpreter/mod.rs:20
error: src/interpreter/mod.rs: patch does not apply
error: patch failed: src/session.rs:1
error: src/session.rs: patch does not apply
error: patch failed: src/ui.rs:36
error: src/ui.rs: patch does not apply
```


yanchith:
> Sorry for not squashing the commits into a single one but git diff failed like this so I just rebased the branch onto Master.

Was the branch updated?

yanchith:
> Is it smart to store values: Vec<Option<T>> ?

Yes, it is ok to store it. Depending on the `T`, it usually holds that `mem::size_of::<Option<T>>()` is greater than `mem::size_of::<T>()`, but sometimes there is no way around it. Those sizes are equal only if the compiler proves `T` can never have a zero bit pattern (e.g. this is true for `&` and `&mut`, or `NonZero<T>`).

Does this mean I should hold off review and you'll perform more simplifications? Let me know once this is ready.

yanchith:
Also don't worry about the failing macOS CI. That seems to be an azure problem.

janper:
Benchmarked with `auticko.obj` on XPS13

## voxel cloud
grow = 0, voxel size = 0.1, fill = no
took **2341ms**, 2444ms, 2389ms, 2715ms

grow = 0, voxel size = 0.1, fill = yes
took 2375ms, **2166ms**, 2310ms, 2230ms

grow = 2, voxel size = 0.1, fill = yes
took **2865ms**, 2866ms, 2874ms, 2897ms

## scalar field i8
_(insufficient precision)_

grow = 0, voxel size = 0.1, fill = no
took **5300ms**, 5919ms, 6217ms, 5976ms

grow = 0, voxel size = 0.1, fill = yes
took 5700ms, **5180ms**, 5298ms, 5499ms

grow = 2, voxel size = 0.1, fill = yes
took 5641ms, 5815ms, 6136ms, **5532ms**

## scalar field i16
grow = 0, voxel size = 0.1, fill = no
took **5397ms**, 6373ms, 6674ms, 5977ms

grow = 0, voxel size = 0.1, fill = yes
took 5547ms, 5384ms, 5294ms, **5262ms**

grow = 2, voxel size = 0.1, fill = yes
took 6039ms, **5773ms**, 5922ms, 6087ms

janper:
Ready for review

src/interval.rs
```diff
@@ -0,0 +1,785 @@
...
+/// Interval is a set of real numbers lying between two numbers, the extremities
+/// (left and right) of the interval.
+/// # Source
+/// https://en.wikipedia.org/wiki/Interval_(mathematics)
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct Interval<T> {
+    left: T,
+    right: T,
+    left_infinite: bool,
+    right_infinite: bool,
```
janper: I'm considering a refactor: instead of storing these two flags, this could be handled by the constructors - if the interval should be infinite, then the infinite side can be set to the `min_value` or `max_value` of `T`. Because `T` implements `Bounded` trait, there always are specific numeric extremes. 
It would make the implementation simpler (no infinity checks anymore) and it also makes an entirely infinite (left and right) interval possible. 

It will be then also possible to remap a finite interval into an infinite interval, which might be mathematically incorrect but it could be quite handy in our case. 

We currently use the Interval only to specify a value range of voxelized scalar field, that should be considered to be a volume. The scalar field can be constructed from a mesh - imagine a sphere. The voxels representing the sphere surface will be set to value **0** (this is customizable) and the rest will be set to None. Converting interval **0to0** into a mesh materializes the voxelized sphere surface and generates two surfaces - outer and inner. Then it's possible to compute the distance field. Each voxel will be set a value representing its distance from the original mesh surface. The voxels that used to be **0** will remain **0**, the ones next tot them will be **1**, next layer will be **2** etc. But not only that - the voxels inside closed mesh surfaces will be set to negative values, so the sphere surface will be **0**, the concentric sphere bigger than the original will be set to **1** and the concentric sphere smaller than the original will be set to **-1** etc. Now it's possible to materialize mesh from **0to0**, which will be the same as before. Interval **-1to1** will be 3 voxels thick. Interval **infinityTo0** will be the sphere including its inside, so there won't be any inner surface materialized into mesh. Similarly, **1toInfinity** is the entire outer space.
When performing boolean or other operations on two scalar fields, it is valid to make i.e. an union of field voxels within interval **-1to1** (thicker hollow surface) and **infinityTo0** (filled volumes) and remapping values from **infinityTo0** to **-1to1**. The result will be valid and most probably will look good. 

To the best visual outputs, it would be best to calculate the actual value of infinity (the lowest or highest actually used value) and set the numerical values of interval extremes accordingly. This would be computationally more demanding, it will require the flags I'm suggesting to remove and either the scalar field needs to know about intervals or intervals about scalar fields. None of these options seems to be smart though.

src/bounding_box.rs
```diff
@@ -14,7 +16,7 @@ where
     maximum_point: Point3<T>,
 }
 
-impl<T: Bounded + Scalar + Zero + PartialOrd> BoundingBox<T> {
+impl<T: Add<Output = T> + Bounded + Scalar + Sub<Output = T> + PartialOrd + Zero> BoundingBox<T> {
```
yanchith: Just use `Num` as a bound. It includes `Add` and `Sub`. I did so in my branch.

src/interval.rs
```diff
@@ -0,0 +1,785 @@
+use std::ops::{Add, Div, Mul, Sub};
+
+use num_traits::{Bounded, FromPrimitive, ToPrimitive};
+
+/// Interval is a set of real numbers lying between two numbers, the extremities
+/// (left and right) of the interval.
+/// # Source
```
yanchith: newline

src/interval.rs
```diff
@@ -0,0 +1,785 @@
...
+/// https://en.wikipedia.org/wiki/Interval_(mathematics)
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct Interval<T> {
+    left: T,
+    right: T,
+    left_infinite: bool,
+    right_infinite: bool,
+}
+
+impl<
```
yanchith: As per our discussion on slack, if this is to support all numbers, we can do better here. Let's try to be less granular about the bounds here. Use `num_traits::Num` and remove the bounds we get automatically by `Num` as a bound.

If this were for real numbers only, we could use `num_traits::Real`, which defines even more things.

src/interval.rs
yanchith: We would need to talk more about this. I don't think it is a good idea to use sentinel values for infinities, unless the interval were restricted to floating point values, where sentinel values for infinities have reserved bit patterns.

See my comment about using `Option<T>`.

As for the utility of being able to remap between finite and infinite intervals, it doesn't seem very correct indeed.

yanchith: After talking to you on slack for clarification, I think the best way to represent inifinities is by using `Option<T>`.

src/interval.rs
```diff
@@ -0,0 +1,785 @@
...
+    pub fn centre(&self) -> Option<T> {
+        if self.left_infinite || self.right_infinite {
+            None
+        } else {
+            let (min, max) = if self.right > self.left {
+                (self.left, self.right)
+            } else {
+                (self.right, self.left)
+            };
+            Some((min + max) / T::from_f32(2_f32).unwrap())
```
yanchith: Just in case, let's use `T::from_i32` since we are not utilizing the realness here. Also `unwrap` -> `expect`.

src/interval.rs
```diff
@@ -0,0 +1,785 @@
...
+        if self.right_infinite {
+            return value >= self.left;
+        }
+        let (min, max) = if self.right > self.left {
+            (self.left, self.right)
+        } else {
+            (self.right, self.left)
+        };
+        value >= min && value <= max
+    }
```
yanchith: In maths, the open/closed is usually a property of the interval, not the querying. How often will we want to query the interval multiple ways? See https://doc.rust-lang.org/stable/std/ops/trait.RangeBounds.html and https://doc.rust-lang.org/stable/std/ops/enum.Bound.html for inspiration on how this can be represented.

src/interval.rs
```diff
@@ -0,0 +1,785 @@
...
+        if self.right_infinite {
+            None
+        } else {
+            Some(self.right)
+        }
+    }
+
+    /// Computes the length of the interval. Returns none if any of the bounds
+    /// are infinite.
+    pub fn centre(&self) -> Option<T> {
```
yanchith: You use "center" in comments elsewhere in the PR

src/mesh/scalar_field.rs
```diff
@@ -0,0 +1,1899 @@
...
+/// the block.
+#[derive(Debug, Clone, PartialEq, serde::Serialize)]
+pub struct ScalarField<T> {
+    block_start: Point3<i32>,
+    block_dimensions: Vector3<u32>,
+    voxel_dimensions: Vector3<f32>,
+    voxels: Vec<Option<T>>,
+}
+
+impl<
```
yanchith: Again. The set of traits defined here is pretty much `Num`. Let's not be this granular.

src/mesh/scalar_field.rs
```diff
@@ -0,0 +1,1899 @@
...
+        assert!(
+            voxel_dimensions.x > 0.0 && voxel_dimensions.y > 0.0 && voxel_dimensions.z > 0.0,
+            "One mor more voxel dimensions is below or equal to zero"
+        );
+
+        // Make a bounding box of the source bounding box's mesh volume. This
+        // will be the volume to be scanned for any voxels.
+        source_scalar_field
+            .mesh_volume_bounding_box_cartesian(volume_value_interval)
+            .map(|current_sf_bounding_box| {
```
yanchith: source

src/mesh/scalar_field.rs
```diff
@@ -0,0 +1,1899 @@
...
+        let block_dimensions = Vector3::new(
+            cast_u32(diagonal.x),
+            cast_u32(diagonal.y),
+            cast_u32(diagonal.z),
+        );
+        self.resize(&bounding_box.minimum_point(), &block_dimensions);
+    }
+
+    /// Resize the scalar field block to exactly fit the volumetric geometry.
+    /// Returns None for empty the scalar field.
```
yanchith: Comment does not describe what `volume_value_interval` does. Also everywhere.

src/mesh/scalar_field.rs
```diff
@@ -0,0 +1,1899 @@
...
+    }
+
+    /// Creates a new scalar field with arbitrary voxel dimensions from another
+    /// scalar field.
+    ///
+    /// # Panics
+    /// Panics if any of the voxel dimensions is below or equal to zero.
+    pub fn from_scalar_field(
+        source_scalar_field: &ScalarField<T>,
+        volume_value_interval: Interval<T>,
```
yanchith: Let's document what `volume_value_interval` means.

src/interval.rs
```diff
@@ -0,0 +1,785 @@
+use std::ops::{Add, Div, Mul, Sub};
+
+use num_traits::{Bounded, FromPrimitive, ToPrimitive};
+
+/// Interval is a set of real numbers lying between two numbers, the extremities
```
yanchith: As we talked on slack, this documentation needs to be updated. You said the interval can also contain integers.

src/mesh/scalar_field.rs
```diff
@@ -0,0 +1,1899 @@
...
+
+        assert!(
+            voxel_dimension > 0.0,
+            "Voxel dimension is below or equal to zero"
+        );
+
+        // Make a bounding box of the source scalar field's mesh volume. This
+        // will be the volume to be scanned for any voxels.
+        if let Some(source_sf_bounding_box) =
+            source_scalar_field.mesh_volume_bounding_box_cartesian(volume_value_interval)
```
yanchith: Why mesh volume instead of volume?

src/mesh/scalar_field.rs
```diff
@@ -0,0 +1,1899 @@
...
+    ) -> Option<Self> {
+        assert!(
+            voxel_dimensions.x > 0.0 && voxel_dimensions.y > 0.0 && voxel_dimensions.z > 0.0,
+            "One mor more voxel dimensions is below or equal to zero"
+        );
+
+        // Make a bounding box of the source bounding box's mesh volume. This
+        // will be the volume to be scanned for any voxels.
+        source_scalar_field
+            .mesh_volume_bounding_box_cartesian(volume_value_interval)
```
yanchith: Why mesh volume instead of volume?

src/mesh/scalar_field.rs
```diff
@@ -0,0 +1,1899 @@
...
+    }
+
+    /// Computes boolean union (logical OR operation) of two scalar fields. The
+    /// current scalar field will be mutated and resized to contain both input
+    /// scalar fields' volumes. The values from the other scalar field which are
+    /// considered a volume, will be remapped to the volume value interval of
+    /// the source scalar field.
+    ///
+    /// # Panics
+    /// Panics if one of the volume voxel intervals is infinite.
```
yanchith: Did you mean volume *value* interval? Why panic? Isn't it better to disallow infinite intervals? Also, why does only union remap values, but intersection and difference don't?

src/mesh/scalar_field.rs
```diff
@@ -0,0 +1,1899 @@
...
+    T: Add<Output = T>
+        + Bounded
+        + Copy
+        + Div<Output = T>
+        + FromPrimitive
+        + Mul<Output = T>
+        + Neg<Output = T>
+        + PartialOrd
+        + Sub<Output = T>
+        + ToPrimitive,
```
yanchith: See trait bound comments above. Also, isn't it better to inline this to places where it is used?

yanchith (commented):
This is the first part of the review. I managed to review `interval.rs` (without tests, because am not sure we should keep it), `bounding_box.rs` and about the first half of `scalar_field.rs`.

Please don't apply any changes yet.

Had a look at `interval.rs` and I am inclined to think it doesn't bring much value beyond the ranges in `std`, especially if you use them generically with https://doc.rust-lang.org/stable/std/ops/trait.RangeBounds.html. Wouldn't it be better if we used ranges and implemented remapping as a normal function on ranges (using the `RangeBounds` trait)?

I still left comments in the `interval.rs` file in case we decide to keep it.

src/interval.rs
janper: I was hesitating.
I'm not sure how often will we read the same interval in various ways. Maybe never.
Regarding Interval vs. Range, this was my thinking (but I don't insist on keeping the Interval):

- Range allows for iterating over the values. This is semantically not possible with Interval.
- Range doesn't allow for left-open intervals such as `=a..b` (or am I missing something?)
- We'd have to implement custom functions, such as `length`, `centre` and `remap`

src/interval.rs
janper: As for my defense, integers are a subset of real numbers, so strictly speaking this is a valid comment: https://en.wikipedia.org/wiki/Real_number

When we talked on Slack, I wasn't sure about this but when I wrote the comment, this was my thinking.

For clarification, I'll update the comment to reflect the actual use in this software - _Interval is a set of float or integer numbers_.

src/interval.rs
janper: I'd stick to _centre_: _The centre (midpoint) of bounded interval with endpoints a and b is (a + b)/2_ from https://en.wikipedia.org/wiki/Interval_(mathematics)

src/mesh/scalar_field.rs
janper: Because the new scalar field should be able to hit all the materialized voxels, which means it needs to scan the volume occupied by the meshed voxels.

src/mesh/scalar_field.rs
janper: Yes, I mean `volume value interval`.
~It panics if the intervals are infinite on both ends due to remap.~ True, it doesn't need to panic.

This is the only boolean operation where voxels from the other scalar field are being transferred to self. Therefore it's the only one that remaps. During intersection or difference some voxels can be removed, but no voxels are being added.

src/interval.rs
yanchith: > Range allows for iterating over the values. This is semantically not possible with Interval.

I don't think it should trouble us if we don't use it. No need to prevent it.

> Range doesn't allow for left-open intervals such as =a..b (or am I missing something?)

Looks like there is no literal for that indeed. The `RangeBounds` trait does support it though, and this impl exists: `impl<T> RangeBounds<T> for (Bound<T>, Bound<T>)`

> We'd have to implement custom functions, such as length, centre and remap

From what I remember seeing in the PR, only `remap`. `centre` and `length` were just used to power remap. (I might remember incorrectly, or this fact could have changed since the last review).

src/interval.rs
yanchith: Kk, integers are real numbers, true - my bad.

Btw floating point numbers are not synonymous with real numbers. They are a subset of real numbers. Other subsets are rational numbers, e.g.

src/mesh/scalar_field.rs
yanchith: > This is the only boolean operation where voxels from the other scalar field are being transferred to self. Therefore it's the only one that remaps. During intersection or difference some voxels can be removed, but no voxels are being added.

This is true, but the remap in union affects also the voxels that already existed in `self`, doesn't it?

src/interval.rs
janper: Exactly! That's why I'm about to change the description to match the reality - this expects Floats and Integers

src/mesh/scalar_field.rs
janper: > This is true, but the remap in union affects also the voxels that already existed in `self`, doesn't it?

No, only the ones that are being added are remapped:
`558 if !is_voxel_within_closed_interval(voxel, volume_value_interval_self) {`

yanchith:
Can we close this pull request since it is coming in smaller batches? (I ask because of notifications)

PR #124, opened 2020-01-29
Add configurable shadow rendering

yanchith:
The renderer can now draw shadows via shadow mapping. This is used to collect
shadows from rendered meshes and project them onto the ground plane (which is a
plane mesh primitive).

The `CommandBuffer::draw_meshes_*` family of functions now has a flag, whether
the drawing should cast shadows. The `DrawMeshMode` enum has a new
`FlatWithShadows` variant that can collect shadows. The structure is much more
flexible internally in the renderer, but not yet exposed. Also, the renderer now internally supports transparency, but the only way to render a transparent object is to provide a transparent color to the `flat_shading_color` option - that color is used for `DrawMeshMode::FlatWithshadows`.

Known problems:

- Vulkan validation errors for image layout transitions (this is probably not
  our fault, but it looks scary),
- Screenshots with transparent background can now see ground, because
  it is not transparent. We can alternatively not render the ground plane for
  screenshots.


janper (dismissed):
Nice and it does all that we discussed.
Let's disable shadows for transparent screenshots.

yanchith:
Thanks! I'll have a fresh look at the code later and clean it up before merging.

yanchith:
I've given it some thought and think that proper (or at least not terribly incorrect) transparency is within our grasp. Probably less than a day of work. I'll update the PR with that next week.

If anyone wants to review the code, I can merge this first and add transparency with another PR to make the reviewer's job easier.

yanchith:
Ready for review.

janper:
Nice. A couple of notes and questions:
1. ~There seems to be this glitch when looking from below~ It was happening also before the shadows, so let's ignore it. FYI, the sphere is set to 32x32 segments. glitch 
2. You mentioned the shadows can't be more blurry. Is it still the case and why?
3. Once we have a transparency, it would be nice to have some very transparent endless ground because it's confusing when there is no visible shadow because it's inside of the model. If truly endless is a problem, it can be a disk/plane that gradually fades out into full transparency. 
image
Untitled2

yanchith:
> There seems to be this glitch when looking from below It was happening also before the shadows, so let's ignore it. FYI, the sphere is set to 32x32 segments.

Yeah, I believe that is a matcap rendering glitch, I've seen it before too. Don't exactly know what's causing it tbh.

> You mentioned the shadows can't be more blurry. Is it still the case and why?

Yes. For the same reason as before. We are already smoothing the shadows by doing PCF (percentage closer filtering) https://developer.nvidia.com/gpugems/gpugems/part-ii-lighting-and-shadows/chapter-11-shadow-map-antialiasing

We could increase the number of samples to get marginally better results, but my machines already struggle with it when running on intel GPUs - this would make us require capable hardware for everyone, unless we had a flag to turn it down. Not sure if that's worth it.

I can show you how to up the shadow quality so that you can experiment with different values.

Currently, we treat the ground plane as just another mesh in the renderer that just uses a different `ShadingMode`, one that can catch shadows. We could treat it specially and implement some kind of postprocess based shadow blurring just for the ground plane, but that sounds like both a lot of work and also will have performance implications.

> Once we have a transparency, it would be nice to have some very transparent endless ground because it's confusing when there is no visible shadow because it's inside of the model. If truly endless is a problem, it can be a disk/plane that gradually fades out into full transparency.

Not sure what you mean. We already have a fully transparent (not endless, but very large) ground. If the objects were transparent, you would see the shadow there, and you already can in the `Edges` rendering mode. Are you saying that you don't want it to be fully transparent? We can already make it not fully transparent and it behaves as on your pictures around the models, but then the edges of the ground are visible. To solve that, we could blend it into full transparency on the edges using a texture mask (or some clever shader trick). I'd keep that for a separate PR though.

janper (dismissed):
As long as it is what I saw at your computer, we are good to merge. Have you considered setting the ground plane size according to scene's bounding box though?

yanchith:
> Have you considered setting the ground plane size according to scene's bounding box though?

Ok, I'll do it

yanchith:
I still discovered some screenshot transparency issues with the new transparent ground. I'll need to investigate. Will merge after that is fixed.

PR #125, opened 2020-02-04
Fix minimize crash

yanchith:


janper (approved):
Works great on my Windows 10!

PR #126, opened 2020-02-04
Add optional analytic messages to operations

janper:
All operations now allow to compute and display analytic information for resulting meshes or groups. All reports are logged to the respective operation's console. This is useful when working with an unknown geometry, i.e. when importing an OBJ. Some analytics may be computationally intense, therefore all analytic reports are set off by default.


src/interpreter/analytics.rs
```diff
@@ -0,0 +1,154 @@
+use std::collections::HashSet;
```
yanchith: I wouldn't put in in the `interpreter` module. For now, top-level module is fine.

src/interpreter_funcs/create_box.rs
```diff
@@ -95,6 +103,12 @@ impl Func for FuncCreateBox {
             Vector3::from(scale),
         );
 
+        if analyze {
+            analytics::report_mesh_analysis(&value)
+                .iter()
+                .for_each(|line| log(line.clone()));
```
yanchith: Shouldn't need to clone. Can move. Try `into_iter()`. Same goes for all the other reports.

src/interpreter_funcs/extract.rs
```diff
@@ -2,19 +2,25 @@ use std::error;
...
 pub enum FuncExtractError {
     Empty,
+    IndexOutOfBounds(u32, u32),
 }
 
 impl fmt::Display for FuncExtractError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
             Self::Empty => write!(f, "No mesh geometry contained in group"),
+            Self::IndexOutOfBounds(index, length) => write!(
```
yanchith, line 20: Nice!

src/interpreter/analytics.rs
```diff
@@ -0,0 +1,154 @@
...
+
+use crate::bounding_box::BoundingBox;
+use crate::interpreter::{LogMessage, MeshArrayValue};
+use crate::mesh::{analysis, Mesh, UnorientedEdge};
+
+pub fn report_mesh_analysis(mesh: &Mesh) -> Vec<LogMessage> {
```
yanchith: This will work, but it would be slightly better if propagated the logger as a parameter instead. We wouldn't need to allocate the temporary vector in this case. I think the log function can possibly even have the same signature.

I know you asked me about this IRL and I said something else, but the difference here is that these 2 are ad-hoc functions that will probably always live in the application. I would still be against passing the logger to library functions.

src/interpreter_funcs/join_meshes.rs
```diff
@@ -41,11 +49,19 @@ impl Func for FuncJoinMeshes {
     fn call(
         &mut self,
         args: &[Value],
-        _log: &mut dyn FnMut(LogMessage),
+        log: &mut dyn FnMut(LogMessage),
     ) -> Result<Value, FuncError> {
-        let meshes = args.iter().map(|a| a.unwrap_mesh());
+        let meshes = vec![args[0].unwrap_mesh(), args[1].unwrap_mesh()];
```
yanchith: Let's not allocate a vector. You can use an array or a slice instead.

yanchith (changes requested):
Nice! See a couple of inline suggestions.

src/interpreter/analytics.rs
yanchith: Or maybe let's stuff them in `interpreter_funcs/mod.rs`

janper: u sure? `interpreter_funcs/mod.rs` is quite nice and clean now, should we pollute it?

src/interpreter_funcs/create_box.rs
janper: obsolete

src/interpreter_funcs/join_meshes.rs
janper: help please?

```rust
let meshes = &[args[0].unwrap_mesh(), args[1].unwrap_mesh()];
let value = tools::join_multiple_meshes(meshes);
```

```
error[E0271]: type mismatch resolving `<&[&mesh::Mesh; 2] as std::iter::IntoIterator>::Item == &mesh::Mesh`
   --> src\interpreter_funcs\join_meshes.rs:58:21
    |
58  |         let value = tools::join_multiple_meshes(meshes);     
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected &mesh::Mesh, found struct `mesh::Mesh`
    | 
   ::: src\mesh\tools.rs:333:8
    |
333 | pub fn join_multiple_meshes<'a, I>(meshes: I) -> Mesh        
    |        --------------------
334 | where
335 |     I: IntoIterator<Item = &'a Mesh>,
    |                     --------------- required by this bound in `mesh::tools::join_multiple_meshes`
    |
    = note: expected type `&&mesh::Mesh`
               found type `&mesh::Mesh`
```

yanchith (approved):
Thank you

PR #127, opened 2020-02-05
Add option to snap geometry to the origin and ground

janper:
Align objects to the origin or/and ground:
- optionally when importing an OBJ group
- as a separate operation


src/interpreter_funcs/snap_to_ground.rs
```diff
@@ -0,0 +1,89 @@
...
+impl Func for FuncSnapToGround {
+    fn info(&self) -> &FuncInfo {
+        &FuncInfo {
+            name: "Snap To Ground",
+            return_value_name: "Mesh on origin",
+        }
+    }
+
+    fn flags(&self) -> FuncFlags {
+        FuncFlags::empty()
```
yanchith: Isn't this pure?

yanchith (commented):
Looks good, but snap to ground should probably be marked pure.

Also, depending on when #126 gets merged, you might need to add those messages here as well

yanchith (dismissed):
Thank you. I'll merge soon, once the builds pass.

yanchith:
Actually, you'll need to re-apply the changes from the analytics. Or do you want to merge this first and do that in a different PR?

janper:
I can do the merge if you are okay with the PR. Sure, I'll resolve the discrepancies.

yanchith:
Kk, resolve conflicts and then I'll review again and hopefully merge.

janper:
Ready to be checked

src/interpreter_funcs/import_obj_mesh.rs
```diff
@@ -82,18 +102,54 @@ impl<C: ObjCache> Func for FuncImportObjMesh<C> {
...
+
+                        let bboxes = meshes_iter.clone().map(|mesh| mesh.bounding_box());
+                        let union_box = BoundingBox::union(bboxes).expect("No valid meshes");
+
+                        let translation_vector = if move_to_origin {
+                            Point3::origin() - union_box.center()
+                        } else {
+                            Vector3::zeros()
+                        } + if snap_to_ground {
+                            Vector3::new(0.0, 0.0, union_box.diagonal().z / 2.0)
```
yanchith: - `move_to_origin` + `snap_to_ground` work
- standalone `move_to_origin` works
- standalone `snap_to_ground` does not work

I think we need to do different work for `snap_to_ground` depending on the state of `move_to_origin`. I suggest a match statement similar to:

```rust
let translation_vector = match (move_to_origin, snap_to_ground) {
    (true, true) => { /*...*/ }
    (true, false) => { /*...*/ }
    (false, true) => { /*...*/ }
    (false, false) => { /*...*/ }
}
```

You will need to apply the same fix to the other func.

src/interpreter_funcs/snap_to_ground.rs
```diff
@@ -0,0 +1,97 @@
...
+            },
+            ParamInfo {
+                name: "Snap to ground",
+                description: "",
+                refinement: ParamRefinement::Boolean(BooleanParamRefinement {
+                    default_value: true,
+                }),
+                optional: false,
+            },
+        ]
```
yanchith, line 59: You didn't add the analysis option here. Is that on purpose?

yanchith (changes requested):
Sorry to pester you, but I didn't notice this earlier. When only `snap_to_ground` is true, the object does not end up on the ground this way.

src/interpreter_funcs/import_obj_mesh.rs
janper: Are you sure? I doublechecked the code and tested the behavior and it seems to works well.

yanchith: Unless it behaves different on my machine, I am pretty sure. I tested with `bandaska.obj` (and others), both import and the snap to ground.

EDIT: tested again on and reproduced.

janper: Of course! I used `tatra` for testing and that one coincidentally works.

yanchith (approved):
Thank you

PR #128, opened 2020-02-07
Tooltips

yanchith:
Add description fields to funcs and params. Draw these descriptions on hover in the UI when present.

@janper No need to add the description texts themselves in this PR, you can do so at another time. If the description is empty, it doesn't render. I did leave one lorem impsum in there to show how to format paragraphs, though.


janper (approved):
Looks great!

I'll make a new PR with all the strings filled-in and the existing ones rephrased to match the rest.

You've also removed the shrink wrap operation. Please add a note to the commit message or description.

PR #131, opened 2020-02-10
Add a note about icons to installer readme

ondrowan:
@janper Will this be clear in the future or should I add it somewhere more visible?

Also note that I'm not entirely sure what sizes the different icons should be. There seem to be a lot to take into consideration, from user settings to DPI. I haven't found some unified guidelines regarding this and even Microsoft docs seem to have all kinds of different (outdated) information. Also I'm not entirely sure if just one size of icon is saved as window icon and desktop icon or `winit` / `rcedit` do some resizing. Do you want to spend more time on this or get back to it later if it's a problem?

### DIFF
--- a/installer/README.md
+++ b/installer/README.md
@@ -18,6 +18,10 @@ To create a distribution follow these steps:
 - Compile `installer/setup.iss` from Inno Setup IDE,
 - Output file will be located at `installer/bin`.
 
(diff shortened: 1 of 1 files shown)


yanchith (approved):

janper (approved):
This is perfectly good. No need to spend more time. 
AFAIK, icons now can be any BMP or PNG images, therefore there may not be any clear guidelines.

PR #133, opened 2020-02-12
Add description to operations, parameters and renames labels

janper:
_Note: The trailing white spaces are being added by `cargo fmt`._

### DIFF
--- a/src/interpreter_funcs/create_box.rs
+++ b/src/interpreter_funcs/create_box.rs
@@ -15,17 +15,13 @@ impl Func for FuncCreateBox {
     fn info(&self) -> &FuncInfo {
         &FuncInfo {
             name: "Create Box",
-            description: "\
-Lorem Ipsum is simply dummy text of the printing and typesetting industry. \n\
-\n\
-
-Lorem Ipsum has been the industry's standard dummy text ever since the \
-1500s, when an unknown printer took a galley of type and scrambled it to make \
-a type specimen book. It has survived not only five centuries, but also the leap \
-into electronic typesetting, remaining essentially unchanged. It was popularised \
-in the 1960s with the release of Letraset sheets containing Lorem Ipsum \
-passages, and more recently with desktop publishing software like Aldus \
-PageMaker including versions of Lorem Ipsum.",
+            description: "CREATE MESH BOX\n\
+                          \n\
+                          Creates a new mesh box made of 12 welded triangles \
+                          and 8 vertices.\n\
+                          The default size of the box is 1x1x1 model units.\n\
+                          \n\
+                          The resulting mesh geometry will be named 'Box'.",
             return_value_name: "Box",
         }
     }
@@ -38,7 +34,7 @@ PageMaker including versions of Lorem Ipsum.",
         &[
             ParamInfo {
                 name: "Center",
-                description: "",
+                description: "Center of the box in absolute model units.",
                 refinement: ParamRefinement::Float3(Float3ParamRefinement {
                     default_value_x: Some(0.0),
                     min_value_x: None,
@@ -54,7 +50,7 @@ PageMaker including versions of Lorem Ipsum.",
             },
             ParamInfo {
                 name: "Rotate (deg)",
-                description: "",
+                description: "Rotation of the box in degrees.",
                 refinement: ParamRefinement::Float3(Float3ParamRefinement {
                     default_value_x: Some(0.0),
                     min_value_x: None,
@@ -70,7 +66,8 @@ PageMaker including versions of Lorem Ipsum.",
             },
             ParamInfo {
                 name: "Scale",
-                description: "",
+                description: "Scale of the box as a relative factor.\n\
+                              The original size of the box is 1x1x1 model units.",
                 refinement: ParamRefinement::Float3(Float3ParamRefinement {
                     default_value_x: Some(1.0),
                     min_value_x: None,
@@ -86,7 +83,8 @@ PageMaker including versions of Lorem Ipsum.",
             },
             ParamInfo {
                 name: "Analyze resulting mesh",
-                description: "",
+                description: "Reports detailed analytic information on the created mesh.\n\
+                              The analysis may be slow, therefore it is by default off.",
                 refinement: ParamRefinement::Boolean(BooleanParamRefinement {
                     default_value: false,
                 }),
--- a/src/interpreter_funcs/create_box.rs
+++ b/src/interpreter_funcs/create_box.rs
@@ -18,7 +18,7 @@ impl Func for FuncCreateBox {
             description: "CREATE MESH BOX\n\
                           \n\
                           Creates a new mesh box made of 12 welded triangles \
-                          and 8 vertices.\n\
+                          and 8 vertices. \
                           The default size of the box is 1x1x1 model units.\n\
                           \n\
                           The resulting mesh geometry will be named 'Box'.",
--- a/src/interpreter_funcs/create_plane.rs
+++ b/src/interpreter_funcs/create_plane.rs
@@ -17,14 +17,12 @@ impl Func for FuncCreatePlane {
         &FuncInfo {
             name: "Create Plane",
             description: "CREATE MESH PLANE\n\
-            \n\
-                        
-            Creates a new mesh plane made of two welded triangles \
-            and four vertices.\n\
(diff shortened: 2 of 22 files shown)


yanchith:
> Note: The trailing white spaces are being added by cargo fmt.

Hmmm interesting. It looks like any whitespace after the `/` escape is insignificant, which is very nice for formatting indeed.

src/interpreter_funcs/create_box.rs
```diff
@@ -15,17 +15,13 @@ impl Func for FuncCreateBox {
...
-1500s, when an unknown printer took a galley of type and scrambled it to make \
-a type specimen book. It has survived not only five centuries, but also the leap \
-into electronic typesetting, remaining essentially unchanged. It was popularised \
-in the 1960s with the release of Letraset sheets containing Lorem Ipsum \
-passages, and more recently with desktop publishing software like Aldus \
-PageMaker including versions of Lorem Ipsum.",
+            description: "CREATE MESH BOX\n\
+                          \n\
+                          Creates a new mesh box made of 12 welded triangles \
+                          and 8 vertices.\n\
```
yanchith: General note: Is this new paragraph on purpose? Don't we want to let the text wrapping to take care of this? I looked at how this looks rendered and I am not sure this is on purpose.

src/interpreter_funcs/create_plane.rs
```diff
@@ -16,7 +16,15 @@ impl Func for FuncCreatePlane {
     fn info(&self) -> &FuncInfo {
         &FuncInfo {
             name: "Create Plane",
-            description: "",
+            description: "CREATE MESH PLANE\n\
+            \n\
+
```
yanchith: This line is quite a lot of whitespace. If you remove it, even `rustfmt` formats this similarly to the other things. Also everywhere.

src/interpreter_funcs/laplacian_smoothing.rs
```diff
@@ -14,7 +14,23 @@ impl Func for FuncLaplacianSmoothing {
...
+            For each vertex in a mesh, a new position is chosen based on local \
+            information (such as the position of neighbors) and the vertex is moved there.\n\
+            The vertex and face count will remain unchanged. \n\
+            \n\
+            Laplacian smoothing removes small details, grain and kinks of the original model. \
+            Too many iterations may reduce the mesh volume. \
+            The output mesh will be recomputed with smooth normals.\n\
+            \n\
+            The input mesh will be marked used and thus invisible in the viewport. \
+            It can still be used in subsequent operations.\n\
```
yanchith: Do we want to talk about this? This might change before the exhibition.

src/interpreter_funcs/loop_subdivision.rs
```diff
@@ -51,13 +73,17 @@ impl Func for FuncLoopSubdivision {
...
                 refinement: ParamRefinement::Mesh,
                 optional: false,
             },
             ParamInfo {
                 name: "Iterations",
-                description: "",
+                description:
+                    "Number of iterations (repetitions) of the Loop subdivision algorithm.\n\
+                     The number of iterations is limited to 3. \
+                     Too many iterations may take long time and produce a heavy dense mesh.",
```
yanchith, line 87: Isn't this missing a comma between "heavy" and "dense"?

src/interpreter_funcs/revert_mesh_faces.rs
```diff
@@ -13,7 +13,23 @@ impl Func for FuncRevertMeshFaces {
...
+                          Changes winding (vertex order) of all mesh faces.\n\
+                          This is a maintenance operation that helps to \
+                          synchronize properties of multiple meshes before \
+                          hybridization. Some hybridization algorithms require \
+                          the faces of input meshes to face the same direction \
+                          (inwards or outwards). This operation is also useful \
+                          after Synchronizing mesh faces, which can result in \
+                          a mesh facing inwards.\n\
+                          \n\
+                          This operation does not affect normals and mesh rendering.\n\
```
yanchith, line 27: It is low prio for us, but we will eventually render backfaces with a different material (for nontransparent meshes). Maybe let's append "in the current version of the software".

src/interpreter_funcs/synchronize_mesh_faces.rs
```diff
@@ -13,7 +13,26 @@ impl Func for FuncSynchronizeMeshFaces {
...
+            This is a maintenance operation that helps to \
+            fix meshes with faces randomly facing inwards and outwards. \
+            The resulting mesh may face entirely inwards. In such case it \
+            is necessary to Revert mesh faces so that they face outwards. \
+            The operation is useful for synchronizing properties of multiple \
+            meshes before hybridization. Some hybridization algorithms \
+            require the faces of input meshes to face the same direction \
+            (inwards or outwards).\n\
+            \n\
+            This operation does not affect normals and mesh rendering.\n\
```
yanchith: Same as above. It might, in the future.

src/interpreter_funcs/transform.rs
```diff
@@ -83,16 +91,17 @@ impl Func for FuncTransform {
             ParamInfo {
                 name: "Transform around object center",
                 description: "\
```
yanchith: Let's be consistent and remove the leading backslash

src/lib.rs
```diff
@@ -308,7 +308,7 @@ pub fn init_and_run(options: Options) -> ! {
                 camera.zoom(input_state.camera_zoom);
                 camera.zoom_step(input_state.camera_zoom_steps);
 
-                let utilities_status = ui_frame.draw_utilities_window(
+                let utilities_status = ui_frame.draw_menu_window(
```
yanchith: Might as well rename the variable here too.

src/ui.rs
```diff
@@ -1165,7 +1165,7 @@ impl<'a> UiFrame<'a> {
                     ty == Ty::MeshArray,
                 )
             })
-            .unwrap_or_else(|| imgui::ImString::new("<Nil>"));
+            .unwrap_or_else(|| imgui::ImString::new("<Select one option>"));
```
yanchith, line 1168: 😢

yanchith (commented):
Amazing job with the texts 🎉

Hmm, I see `rustfmt` is having a hard time being consistent, *sigh*.

General notes:

- I see both single and double `\n` used. Is this on purpose? The rendered effect of the single `\n` paragraphs is rather subtle. I'd personally just use double `\n` and rely on text wrapping elsewhere.

- There's a couple of empty whitespace lines in the descriptions. These are escaped from the rendered output, but they prevented `rustfmt` from formatting the description. Compare how `create_box` and `create_plane` are formatted, e.g. I'd prefer the empty lines weren't there.

- We mention in some places that used geometries become invisible. Does that mean that don't want to do transparent used values? If we want to make transparent used values, I'd remove these comments.

janper:
> We mention in some places that used geometries become invisible. Does that mean that don't want to do transparent used values? If we want to make transparent used values, I'd remove these comments.

Let's keep the note for now and once we have the transparent geometries, we'll remove the description as a part of the PR.
Also, the dead geometry's visibility should be optional.

src/interpreter_funcs/create_box.rs
janper: Yes, all formatting is on purpose, including the single/double newline but i'll go through it one more time and simplify the formatting.

src/interpreter_funcs/laplacian_smoothing.rs
janper: If it does, we'll resolve the description in the respective PR. Also the visibility of the dead geometry should be optional.

src/interpreter_funcs/loop_subdivision.rs
janper: I don't think so. They are not equivalent but rather complementary.

src/interpreter_funcs/revert_mesh_faces.rs
janper: Isn't the "in the current version of the software" by default? Again, I'd resolve the description once it's relevant. Should i leave FIXME everywhere?

yanchith: Nah, we'll address it when it happens.

yanchith (approved):

PR #134, opened 2020-02-12
Implement simple align operation

janper:
Moves and scales a mesh to match another mesh. 'Mesh to align' will be translated so that its center matches the center of 'Align to mesh'.  'Mesh to align' will be scaled so that the diagonal of its bounding box will have the same length as the diagonal of the bounding box of 'Align to mesh'.

This is the first step towards a more sophisticated alignment method using scalar fields.


yanchith (approved):
👍

PR #135, opened 2020-02-12
Add error report for Synchronize mesh faces operation

janper:
This adds explicit error message if impossible to synchronize mesh faces due to non-manifold input mesh.


yanchith (approved):
👍

PR #137, opened 2020-02-13
Implement autorun in Session and add it in UI

yanchith:
This adds a new checkbox to the operations window that controls the interpreter
session "autorun" mode. When enabled the session automatically runs the
interpreter as new edits are made to the program.

Implementation-wise, the `time` variable queried at the beginning of frame
processing is passed down everywhere program edits or polling are made. We could
just use `Instant::now()` internally in those functions, but that does a syscall
at best, and a mutex lock at worst each time (see docs for
`std::time::Instant::now()`). We can potentially simplify later.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,7 +24,7 @@ use crate::plane::Plane;
 use crate::renderer::{
     DirectionalLight, DrawMeshMode, GpuMesh, GpuMeshHandle, Options as RendererOptions, Renderer,
 };
-use crate::session::{PollInterpreterResponseNotification, Session};
+use crate::session::{PollNotification, Session};
 use crate::ui::{ScreenshotOptions, Ui};
 
 pub mod geometry;
@@ -49,7 +49,9 @@ mod pull;
 mod session;
 mod ui;
 
-const CAMERA_INTERPOLATION_DURATION: Duration = Duration::from_millis(1000);
+const DURATION_CAMERA_INTERPOLATION: Duration = Duration::from_millis(1000);
+const DURATION_NOTIFICATION: Duration = Duration::from_millis(5000);
+const DURATION_AUTORUN_DELAY: Duration = Duration::from_millis(500);
 
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct Options {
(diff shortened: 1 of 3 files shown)


janper (changes requested):
I like it very much! This makes the software way easier to use. I have a couple of requests or rather questions if we can manage to implement the following things before the exhibition:
- having sliders as numerical inputs will now make a lot of sense https://trello.com/c/8gesm5FU/201-integer-input-should-be-a-dragint
- clipboard support? https://trello.com/c/9iOX2GXt/125-add-clipboard-support-for-imgui
- the delay before the recomputation makes it feel laggy. can you please make it shorter?
- can we pick a different and more integrated UI component for the *Run automatically* switch?
- would it be difficult to add tooltips for hardcoded UI components, such as the run and remove buttons, the switch and also the components in the right menu?

if the answer to all of the above is *No*, then consider this PR approved and ready to merge.

yanchith:
If you want me to do anything else before the exhibition, please assign it to me in trello and sort by priority, but know that I probably won't have enough time to do everything.

> having sliders as numerical inputs will now make a lot of sense https://trello.com/c/8gesm5FU/201-integer-input-should-be-a-dragint

Can be another PR. Should be easy. Please prioritize in trello,

> clipboard support? https://trello.com/c/9iOX2GXt/125-add-clipboard-support-for-imgui

Can be another PR. Don't know how hard that is. Please prioritize in trello.

> the delay before the recomputation makes it feel laggy. can you please make it shorter?

Sure! It is currently 500ms. Do we want it to feel completely instantaneous? We can try to go even below 33 or 16 ms if that's what we want. I'd suggest to not do that, though.

> can we pick a different and more integrated UI component for the Run automatically switch?

Please elaborate. Not sure what you mean here.

> would it be difficult to add tooltips for hardcoded UI components, such as the run and remove buttons, the switch and also the components in the right menu?

As discussed in person, it should be simple. We already have a task for that in trello (https://trello.com/c/fMCD73Te/285-report-the-most-important-events-into-the-main-console). It is currently assigned to you.

yanchith:
I tried 100ms and it feels quite snappy to me with the current kinds of inputs we can have. We can re-adjust the delay after we have sliders and drag inputs.

PR #140, opened 2020-02-15
Add console logs to all major UI and Interpreter events

janper:


src/ui.rs
```diff
@@ -480,21 +486,56 @@ impl<'a> UiFrame<'a> {
...
 
                 ui.radio_button(imgui::im_str!("Shaded"), draw_mode, DrawMeshMode::Shaded);
-                ui.radio_button(imgui::im_str!("Edges"), draw_mode, DrawMeshMode::Edges);
+                if ui.is_item_clicked(imgui::MouseButton::Left) {
+                    notifications.push(
+                        current_time,
+                        NotificationLevel::Info,
+                        "Viewport mode changed to Shaded.",
+                    );
+                }
```
yanchith: No need to do it this way. `Ui::radio_button` returns `true` if it was clicked.

https://docs.rs/imgui/0.2.1/imgui/struct.Ui.html#method.radio_button

Same for all below.

src/ui.rs
```diff
@@ -1131,6 +1183,11 @@ impl<'a> UiFrame<'a> {
         }
 
         if interpret_clicked {
+            notifications.push(
+                current_time,
+                NotificationLevel::Info,
+                "Execution of the Sequence of operations has started...",
+            );
```
yanchith, line 1422: Note that this won't notify us when autorun starts execution. But maybe that's ok.

src/session.rs
```diff
@@ -16,6 +16,8 @@ use crate::interpreter_server::{
 pub enum PollNotification {
     Add(VarIdent, Value),
     Remove(VarIdent, Value),
+    FinishedSuccessfully,
+    FinishedWithError(String),
```
yanchith, line 21: Add a FIXME here:

`// FIXME: Replace String with InterpretError`

This is easier said then done, because `InterpretError` can possibly contain a dynamic boxed runtime error, which is not currently clone-able, and I don't want to bother you with it.

src/session.rs
```diff
@@ -443,12 +445,15 @@ impl Session {
...
                                 Err(interpret_error) => {
-                                    log::error!(
-                                        "Interpreter failed with error: {}",
-                                        interpret_error,
-                                    );
+                                    let error_message = format!("{}", interpret_error);
+                                    callback(PollNotification::FinishedWithError(
+                                        error_message.clone(),
+                                    ));
+                                    log::error!("Interpreter failed with error: {}", error_message);
```
yanchith: Let's allocate less here. Currently this does 3 allocations (`format!`, `log::error!` and `.clone()`), but two will suffice:

```rust
log::error!(
    "Interpreter failed with error: {}",
    interpret_error,
);
callback(PollNotification::FinishedWithError(format!("{}", interpret_error)));
```

src/session.rs
```diff
@@ -16,6 +16,8 @@ use crate::interpreter_server::{
 pub enum PollNotification {
     Add(VarIdent, Value),
     Remove(VarIdent, Value),
```
yanchith: Please rename the existing variants to `ValueAdded` and `ValueRemoved`.

src/lib.rs
```diff
@@ -310,10 +310,13 @@ pub fn init_and_run(options: Options) -> ! {
                 camera.zoom(input_state.camera_zoom);
                 camera.zoom_step(input_state.camera_zoom_steps);
 
+                let menu_notifications = Rc::clone(&notifications);
```
yanchith: This `Rc::clone` is unnecessary, just use the original value.

src/lib.rs
```diff
@@ -325,6 +328,13 @@ pub fn init_and_run(options: Options) -> ! {
 
                     project::save(&save_path, project);
 
+                    let save_project_notification = Rc::clone(&notifications);
```
yanchith: This `Rc::clone` is unnecessary, just use the original value.

src/lib.rs
```diff
@@ -363,6 +373,13 @@ pub fn init_and_run(options: Options) -> ! {
                         session.push_prog_stmt(time, stmt);
                     }
 
+                    let open_project_notification = Rc::clone(&notifications);
```
yanchith: This `Rc::clone` is unnecessary, just use the original value.

src/lib.rs
```diff
@@ -376,7 +393,13 @@ pub fn init_and_run(options: Options) -> ! {
                 ui_frame.draw_notifications_window(&notifications.borrow());
 
                 ui_frame.draw_pipeline_window(time, &mut session);
-                ui_frame.draw_operations_window(time, &mut session, DURATION_AUTORUN_DELAY);
+                let operations_notifications = Rc::clone(&notifications);
```
yanchith: This `Rc::clone` is unnecessary, just use the original value.

src/lib.rs
```diff
@@ -547,6 +570,25 @@ pub fn init_and_run(options: Options) -> ! {
                         }
                         _ => (/* Ignore other values, we don't display them in the viewport */),
                     },
+                    PollNotification::FinishedSuccessfully => {
+                        let interpreter_notification = Rc::clone(&notifications);
```
yanchith: This `Rc::clone` is unnecessary, just use the original value.

src/lib.rs
```diff
@@ -547,6 +570,25 @@ pub fn init_and_run(options: Options) -> ! {
...
+                        interpreter_notification.borrow_mut().push(
+                            time,
+                            NotificationLevel::Info,
+                            "Execution of the Sequence of operations finished successfully.",
+                        );
+                    }
+                    PollNotification::FinishedWithError(error_message) => {
+                        let interpreter_notification = Rc::clone(&notifications);
```
yanchith: This `Rc::clone` is unnecessary, just use the original value.

yanchith (changes requested):
Nice! Got a few improvement suggestions though.

Mainly I noticed you were cloning the `Rc` even where not necessary. The `Rc` pointer only needs to be cloned if it is about to be moved into a closure, such as the one that handles screenshot data (note that it is `FnOnce`, not `FnMut`). If that one clone didn't happen, we would move the notifications into that closure and loose the ability to use them again. For all other uses the clones are unnecessary.

src/ui.rs
yanchith: If we did want to do this... we could maybe add another notification to `poll` ~and remove this one~.

src/session.rs
janper: That was my first try, until adding all the Clone traits brought me to the Box and there Rust happened.

src/ui.rs
janper: I'm hesitant. What do you think?

yanchith: Up to you, but note that the console will really be spammed by these messages if we change this. Let's delay this maybe.

src/ui.rs
```diff
@@ -488,22 +494,53 @@ impl<'a> UiFrame<'a> {
...
+                    notifications.push(
+                        current_time,
+                        NotificationLevel::Info,
+                        "Viewport mode changed to X-Ray: Shaded with internal Edges (Wireframes).",
+                    );
+                }
                 status.reset_viewport =
                     ui.button(imgui::im_str!("Reset Viewport"), [-f32::MIN_POSITIVE, 0.0]);
 
+                if ui.is_item_clicked(imgui::MouseButton::Left) {
```
yanchith: You missed this one. The button also tells you whether it was clicked directly.

yanchith (changes requested):

yanchith:
Cool, now just please resolve the conflict and we can shipit.

yanchith (approved):

PR #142, opened 2020-02-17
Reorder operations and buttons, tweak UI constants, add draggable inputs

janper:
I've decided not to remove per-component extremes for inputs because we might need it some day and our own routine propagates to the UI and restricts the values even when IMGUI doesn't.


src/ui.rs
```diff
@@ -931,9 +983,28 @@ impl<'a> UiFrame<'a> {
...
 
-                                            if ui
-                                                .input_float2(&input_label, &mut float2_lit)
-                                                .read_only(interpreter_busy)
+                                            let mut drag_float2 = ui
+                                                                .drag_float2(&input_label, &mut float2_lit)
+                                                                .speed(0.05)
+                                                                .power(2.0);
+
+                                            if let (Some(min_value_x), Some(min_value_y)) =
```
yanchith: I think this won't work if only one of the component limits is set.

I'd personally really remove per-component limits in our param refinements so we don't have to do mental gymnastics here. That is unless you found some very compelling reason to keep them.

Also below for float3.

yanchith (changes requested):
Very nice, but see inline comment.

src/ui.rs
janper: hm...
i'm thinking what would be the easiest thing to do now. what if i remove setting the limits for `Float2` and `Float3` entirely for now? the UI allows to set the wrong value only temporarily anyway. this would give us a chance to discuss the issue in more details.

if you don't like that, I'd prefer to rewrite this condition so that it works and still keep the individual limits.

yanchith: I still don't know why you'd like to keep the ability to define limits per vector component. I don't think removing that would be hard (but it would be a bit of typing, true).

If you really want to, let's remove these imgui limits from all the drag components, since we do the clamping manually ourselves anyway.

yanchith (approved):
Thank you

PR #144, opened 2020-02-18
Split in-operation analytics, set voxel count limit

janper:
Based on real-life use, the software sometimes tries to make heavy calculations. With autorun turned on, it may cause the program to crash. Therefore safety switches have been implemented to prevent computation of heavy geometry.

The analytic operation has been split into bounding box analytics (quick and turned on by default) and mesh analytics (slow and turned off by default).

Voxel based operations are by default checking the resulting voxel count and if it is above the threshold, the operation returns an error, including a suggestion for a new voxel size. The calculation is imprecise.

All these settings are optional and can be toggled.


src/interpreter_funcs/voxel_boolean_intersection.rs
```diff
@@ -128,9 +136,25 @@ impl Func for FuncBooleanIntersection {
                 optional: false,
             },
             ParamInfo {
-                name: "Analyze resulting mesh",
+                name: "Prevent Unsafe Settings",
+                description: "Stop computation and throw warning if the calculation may be too slow.",
```
yanchith: This seems inconsistent. If you stop computation, this should no longer be a warning, but rather an error. In general warnings probably should not stop computation.

yanchith (commented):
Please resolve conflicts before I review the code. Also, however:

>  it may cause the program to crash

Does this happen or are we just worried it might?

> Therefore safety switches have been implemented to prevent computation of heavy geometry.

I am not sure this is the direction we should be taking. There will always be a way to create heavy geometry, not sure we should be preventing that. If we take this road, where will we stop? Each added parameter adds usability friction.

- Long term I suggest we look into cancellation.
- Short term I suggest we do not tackle this problem at all (except maybe adding warning text to tooltips).

Let's discuss!

src/interpreter_funcs/voxel_boolean_difference.rs
```diff
@@ -154,7 +178,48 @@ impl Func for FuncBooleanDifference {
         let voxel_dimensions = args[2].unwrap_float3();
         let growth_iterations = args[3].unwrap_uint();
         let fill = args[4].unwrap_boolean();
-        let analyze = args[5].unwrap_boolean();
+        let warn_if_slow = args[5].unwrap_boolean();
```
yanchith: `error_if_slow` or `error_if_large`?

src/interpreter_funcs/voxel_boolean_difference.rs
```diff
@@ -154,7 +178,48 @@ impl Func for FuncBooleanDifference {
...
+        let (voxel_count, bbox_diagonal) = if voxel_count1 > voxel_count2 {
+            (voxel_count1, bbox_diagonal1)
+        } else {
+            (voxel_count2, bbox_diagonal2)
+        };
+
+        log(LogMessage::info(format!("Voxel count = {}", voxel_count)));
+
+        if warn_if_slow && voxel_count > VOXEL_COUNT_THRESHOLD {
+            let vy_to_vx = voxel_dimensions[1] / voxel_dimensions[0];
```
yanchith: Correct me if I misunderstand, but if this is supposed to be a transformation, then it transforms from x to y, hence the name should be `vx_to_vy`. Same everywhere this was introduced.

src/interpreter_funcs/voxel_boolean_difference.rs
```diff
@@ -28,6 +31,11 @@ impl fmt::Display for FuncBooleanDifferenceError {
             FuncBooleanDifferenceError::EmptyVoxelCloud => {
                 write!(f, "The resulting voxel cloud is empty")
             }
+            FuncBooleanDifferenceError::TooManyVoxels(max_count, x, y, z) => write!(
+                f,
+                "Too many voxels. Limit set to {}. Try setting voxel size to [{}, {}, {}] or more.",
+                max_count, x, y, z
+            ),
```
yanchith: Note that you are formatting these floating point numbers to have many fractional digits, but the UI is only formatted to take in 3, meaning the user can not actually adhere to the suggestion this error gives.

This would format the output to have 5 fractional digits: `println!("{:.5}", 0.123456);`

Have a look at the `fmt` module for formatting options https://doc.rust-lang.org/stable/std/fmt/index.html

yanchith (commented):
Almost good to go, I do have a couple of small improvement suggestions.

src/interpreter_funcs/voxel_boolean_difference.rs
janper: it's `y / x`, so ti's an *y to x ratio*. not sure how to call it correctly

yanchith: Haha, I have lived with the mstrices for too long. Ok, ignore this.

janper: renamed to `vy_over_vx`

yanchith (approved):
Thank you

PR #145, opened 2020-02-21
Improve saving

ondrowan:
This patch contains all changes needed before the exhibition. There's:

* "new project"
* "save as..."
* handling of errors while saving / opening
* prompt if user wants to exit / open / create new project while the current project is unsaved

One thing that's still not done is layout of buttons in "prevent exit" dialog (save, don't save, cancel). I checked VSCode and they use "don't save" on the left side, then a gap and then "cancel" and "save" next to each other. Is this desirable?


src/lib.rs
```diff
@@ -52,6 +53,7 @@ mod ui;
 const DURATION_CAMERA_INTERPOLATION: Duration = Duration::from_millis(1000);
 const DURATION_NOTIFICATION: Duration = Duration::from_millis(5000);
 const DURATION_AUTORUN_DELAY: Duration = Duration::from_millis(100);
+const BASE_WINDOW_TITLE: &str = "H.U.R.B.A.N. Selector";
```
yanchith: Just note that this has been renamed to "H.U.R.B.A.N. selector" meanwhile in master (notice the lowercase "s"). Also everywhere in this PR.

src/ui.rs
```diff
@@ -911,7 +1034,7 @@ impl<'a> UiFrame<'a> {
         // imgui components can be made read-only, so we can not trust
         // it.
         if !interpreter_busy {
-            if let Some((stmt_index, arg_index, expr)) = change {
+            if let Some((stmt_index, arg_index, expr)) = change.clone() {
```
yanchith: Let's do `let changed = change.is_some()` prior to this and then return that from the function, so that we don't have to clone.

src/ui.rs
```diff
@@ -512,8 +528,23 @@ impl<'a> UiFrame<'a> {
...
+                        status.new_project = true;
+                    }
+
+                    project_status.prevent_overwrite_status = None;
+                    project_status.new_requested = false;
+                }
+
                 if ui.button(imgui::im_str!("Save project"), [-f32::MIN_POSITIVE, 0.0]) {
-                    match project_path {
+                    match project_status.path.clone() {
```
yanchith: Can we still try to match on a reference here and not clone?

src/project.rs
```diff
@@ -10,29 +11,106 @@ use crate::interpreter::ast;
...
+
+impl ProjectStatus {
+    pub fn save(&mut self, path: &str) {
+        self.path = Some(path.to_string());
+        self.changed_since_last_save = false;
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum ProjectError {
```
yanchith, line 40: Let's add `impl error::Error for ProjectError {}`

src/lib.rs
```diff
@@ -359,11 +362,84 @@ pub fn init_and_run(options: Options) -> ! {
...
+                            renderer.remove_scene_mesh(
+                                ground_plane_gpu_mesh_handle
+                                    .take()
+                                    .expect("Ground plane must always be present"),
+                            );
+                            ground_plane_gpu_mesh_handle = Some(
+                                renderer
+                                    .add_scene_mesh(&GpuMesh::from_mesh(&ground_plane_mesh), true)
+                                    .expect("Failed to add ground plane mesh"),
+                            );
```
yanchith, line 426: This probably shouldn't be here. The recomputing should happen when new results are produced by the interpreter. It will either happen automatically (when autorun is enabled), or manually. Either way, we probably shouldn't do it here, before the geometries in the scene actually change.

This is also true for new project. (I am sorry mislead you here in your previous PR)

src/lib.rs
```diff
@@ -375,8 +451,73 @@ pub fn init_and_run(options: Options) -> ! {
...
+                    change_window_title(&window, &project_status);
+                }
+
+                if ui_frame.draw_operations_window(time, &mut session, DURATION_AUTORUN_DELAY) {
+                    project_status.changed_since_last_save = true;
+
+                    change_window_title(&window, &project_status);
+                }
+
+                if let Some(prevent_override_status2) =
```
yanchith: Shadowing is fine. Also one is `override` and one is `overwrite`. Maybe this should be unified across the PR.

src/project.rs
```diff
@@ -10,29 +11,106 @@ use crate::interpreter::ast;
 
 pub const PROJECT_EXTENSION: &str = "hurban";
 
+#[derive(Debug, Clone)]
+pub enum NextAction {
```
yanchith, line 16: Let's implement `Copy` here. Clippy should afterwards point out places where this is cloned (there's a few cases in lib.rs)

yanchith (commented):
Functionality seems to be working well :tada: 

Note that there's conflicts with master, they shouldn't be hard to resolve, but I will have to review again. @janper should probably review only once conflicts have been resolved. Also the code seems to be a bit more complex than what one would think it needs to be, but we can probably refactor later, since an obvious simplification didn't come to mind.

Nit: we'll have to remove the uses of `format!` macro from the UI eventually (per frame allocations are not very perf friendly and fragment the heap if the allocator isn't very good at de-fragmenting). All the strings there could be defined statically. Can wait for later though (I'll create a maintenance task in trello)

janper (changes requested):
Consider approved, however these are my suggestions:
- **Unsaved changes** popup could put the three buttons next to each other?
- **Unsaved changes** popup should implement changes I made to Screenshot popup in PR#146 if that one gets merged first
- **Unsaved changes** popup - please rename the button from _Don't save_ to _Discard changes_
- **Unsaved changes** popup - please change button order to _Save_, _Discard changes_, _Cancel_
- Drop the word _project_ from all the buttons
-  (May not belong to this PR, but where else?) Please remove _Screenshot_ button to _Save Screenshot_
Thanks!

src/lib.rs
yanchith: Hmm, maybe we actually do want to keep this. Looks like a UX decision, so perhaps @janper should decide that. Do we want to clear the scene immediately when opening/newing a project? Or do we want to wait for the interpreter to run the new (potentially empty) program?

ondrowan: I have kept this as is until you two decide what to do with it.

ondrowan:
This is reviewable again. Merge with `master` wasn't very smooth, so please check if I haven't accidentally removed something.

@janper There are some missing tooltips and notifications now that I've merged with `master` branch. Let's wait until everything else is approved and then we can add those as the last thing. Either tell me what texts to use or commit it yourself afterwards.

@yanchith I've removed the most offending `format!` that allocated `String` each frame and used `[].join()` instead. AFAIK it still allocates `String` but with pre-computed length, making it slightly faster. I guess you had something else in mind, so just let me know in chat and we can refactor it later. Since I had no idea what are the possibilities, I made some research and found [this juicy thread](https://www.reddit.com/r/rust/comments/48fmta/seven_ways_to_concatenate_strings_in_rust_the/).

janper (approved):

janper:
Good enough. I'll need to review all the texts again (the printed description uses different words), so I'll take care of the hints as well.

One little thing, if possible - can the buttons be aligned to the window center? 
image
I can also do that in my upcoming PR.

src/lib.rs
yanchith: @janper Didn't reply. Let's keep it.

janper: Sorry, I didn't get any notification.
Also I don't know what does this code do.

I believe when a project is loaded, the scene should be wiped and set to default state. The geometry should be generated only when the pipeline is executed (either by a button or automatically).

yanchith:
@janper 

> I can also do that in my upcoming PR.

Please do.

@ondrowan 

> I've removed the most offending format! that allocated String each frame and used [].join() instead.

That one didn't allocate per frame. The `change_window_title` function was only called conditionally. There are many other invcations of `format!` macro in UI that do it uncoditionally. Let's address that in other PR though.

If there are no objections, I am merging this soon.

yanchith (approved):

src/lib.rs
yanchith: Alright. I'll create a task to ensure this

PR #147, opened 2020-02-21
Display used values as transparent

yanchith:
The feature is published as a check box in the main menu. Enabled by default,
now all used values/geometries are visible in the viewport, rendered with a
transparent(er) version of the currently selected viewport draw mode.

The renderer is refactored to divorce the concepts of a material and a viewport
draw mode. A viewport draw mode is what the application knows about and maps to
renderer materials. The renderer is now able to render each object within a
batch with a different material. Internally, a material is represented as a set
of values contained in a uniform buffer. Depending on a material's transparency
behavior (opaque/transparent/xray), the renderer automatically selects a correct
render pipeline to draw with - some pipelines read-write depth, some only read
depth, and some always pass the depth test without writing - and sorts the
objects in the three transparency category render lists (opaque, transparent,
xray) based on the object's centroid z value after projection with the camera
view matrix.

The session now notifies the caller with changes to used values. Internal value
diffing algorithm has been simplified, optimized and expanded to support
transitioning values between used and unused value sets.

The "Center viewport" feature now takes into consideration, whether used values
are visible, and always centers only around visible values.

Please test extensively. A lot of untested code has been changed and refactored.

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -4,7 +4,7 @@ use std::fmt;
 use std::time::{Duration, Instant};
 
 use crate::interpreter::ast::{FuncIdent, Prog, Stmt, VarIdent};
-use crate::interpreter::{Func, InterpretError, LogMessage, Ty, Value};
+use crate::interpreter::{Func, InterpretError, InterpretValue, LogMessage, Ty, Value};
 use crate::interpreter_funcs;
 use crate::interpreter_server::{
     InterpreterRequest, InterpreterResponse, InterpreterServer, PollResponseError, RequestId,
@@ -14,11 +14,25 @@ use crate::interpreter_server::{
 /// about what values have been added since the last poll, and what
 /// values have been removed are no longer required.
 pub enum PollNotification {
-    ValueAdded(VarIdent, Value),
-    ValueRemoved(VarIdent, Value),
+    UsedValueAdded(VarIdent, Value),
+    UsedValueRemoved(VarIdent, Value),
+    UnusedValueAdded(VarIdent, Value),
+    UnusedValueRemoved(VarIdent, Value),
     FinishedSuccessfully,
     // FIXME: Replace String with InterpretError
     FinishedWithError(String),
+}
+
+#[derive(Debug, Clone, PartialEq)]
+enum DiffEvent {
+    AddUsed(VarIdent, Value),
+    RemoveUsed(VarIdent),
+    VerifyUsed(VarIdent, Value),
+    TransitionUsedToUnused(VarIdent, Value),
+    AddUnused(VarIdent, Value),
+    RemoveUnused(VarIdent),
+    VerifyUnused(VarIdent, Value),
+    TransitionUnusedToUsed(VarIdent, Value),
 }
 
 /// An editing session.
@@ -42,7 +56,12 @@ pub struct Session {
     log_messages: Vec<Vec<LogMessage>>,
     error: Option<InterpretError>,
 
+    used_values: HashMap<VarIdent, Value>,
     unused_values: HashMap<VarIdent, Value>,
+
+    // Working memory for diffing interpreter responses
+    diff_events: Vec<DiffEvent>,
+    diff_processed_idents: HashSet<VarIdent>,
 
     // Auxiliary side-arrays for prog. Determine mesh and mesh-array
     // vars visible from a stmt. The value is read by producing a
@@ -66,10 +85,14 @@ impl Session {
             interpreter_interpret_request_in_flight: None,
             interpreter_edit_prog_requests_in_flight: HashSet::new(),
 
+            diff_events: Vec::with_capacity(64),
+            diff_processed_idents: HashSet::with_capacity(64),
+
             prog: Prog::new(Vec::new()),
             log_messages: Vec::new(),
             error: None,
 
+            used_values: HashMap::new(),
             unused_values: HashMap::new(),
 
             var_visibility_mesh: Vec::new(),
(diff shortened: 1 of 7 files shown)


janper (changes requested):
The transparency level of shaded geometry is set way too high for the dark theme 
image

The wireframe zombie is indistinguishable form the live geometry 
image

The shaded zombie looks good in the funky theme 
image

The wireframe zombie is still bad even in the funky theme 
image

My proposal for the wireframe issue: can we render zombies as shaded without edges in all viewport modes?

Also there is a fundamental problem with the groups: however the group is created now (import or disjoint), it's impossible to do anything with it, but extract. When you extract, the group becomes a zombie an at the same place there is a new live geometry. Therefore you don't see the zombie group (sure, the group may be larger, so  you may see the outstanding bits). Only then you can do something with the mesh, say move it, which renders the extracted mesh a zombie, which is overlapping the zombie group. As a result, you'll almost never see a zombie group! My proposed solution: https://trello.com/c/35MgDAAI/303-for-purposes-of-the-mak-exhibition-replace-import-obj-as-group-with-import-obj-and-join-all-its-meshes-and-remove-both-extract-o

src/ui.rs
```diff
@@ -609,6 +629,33 @@ impl<'a> UiFrame<'a> {
...
+                    viewport_draw_used_values,
+                );
+                if status.viewport_draw_used_values_changed {
+                    notifications.push(
+                        current_time,
+                        NotificationLevel::Info,
+                        if *viewport_draw_used_values {
+                            "Viewport now draws used geometry."
+                        } else {
+                            "Viewport now doesn't draw used geomery."
```
janper: geometry

src/ui.rs
```diff
@@ -609,6 +629,33 @@ impl<'a> UiFrame<'a> {
...
+                        }
+                    );
+                }
+                if ui.is_item_hovered() {
+                    ui.tooltip(|| {
+                        let wrap_token = ui.push_text_wrap_pos(WRAP_POS_TOOLTIP_TEXT_PIXELS);
+                        ui.text("DRAW USED GEOMETRY IN VIEWPORT\n\
+                        \n\
+                        Used geometry is geometry that has been referenced as a parameter \
```
janper, line 704:
```
DRAW USED GEOMETRY IN VIEWPORT\n\
                        \n\
                        When enabled, used geometry will be drawn with \
                        a transparent material.\n\
                        \n\
                        Used geometry is geometry that has been referenced as a parameter \
                        in an operation. 
```

janper (changes requested):

yanchith:
> The transparency level of shaded geometry is set way too high for the dark theme

Ok, I'll make it less transparent for the dark theme.

> The wireframe zombie is indistinguishable form the live geometry

Yes, it was not meant to be different. I can not change it cheaply. This will be addressable in the future, though.

> The shaded zombie looks good in the funky theme

Funnily enough, this was certainly not intentional on my part. I think I do have a blending bug somewhere. Not sure I can fix it quickly, so I'll just pick the transparency value based on the selected theme and make sure it looks good on both.

> The wireframe zombie is still bad even in the funky theme

Same as in dark theme. Can't fix this quickly.

> My proposal for the wireframe issue: can we render zombies as shaded without edges in all viewport modes?

Yes, I'll do that.

> Also there is a fundamental problem with the groups: however the group is created now (import or disjoint), it's impossible to do anything with it, but extract. When you extract, the group becomes a zombie an at the same place there is a new live geometry. Therefore you don't see the zombie group (sure, the group may be larger, so you may see the outstanding bits). Only then you can do something with the mesh, say move it, which renders the extracted mesh a zombie, which is overlapping the zombie group. As a result, you'll almost never see a zombie group! My proposed solution: 

This can happen for non-group values as well (imagine transform with default parameters). This is correct behavior, not a defect. I'd do nothing to change this, both for MAK, and for later.

janper:
> > Also there is a fundamental problem with the groups: however the group is created now (import or disjoint), it's impossible to do anything with it, but extract. When you extract, the group becomes a zombie an at the same place there is a new live geometry. Therefore you don't see the zombie group (sure, the group may be larger, so you may see the outstanding bits). Only then you can do something with the mesh, say move it, which renders the extracted mesh a zombie, which is overlapping the zombie group. As a result, you'll almost never see a zombie group! My proposed solution:
> 
> This can happen for non-group values as well (imagine transform with default parameters). This is correct behavior, not a defect. I'd do nothing to change this, both for MAK, and for later.

The problem is, that with the current toolset there is no possibility for seeing a zombie group. It simply can't happen, except for the rare cases when the group actually contains more than one geometry but we haven't come across such case with the hundreds of sample models we've got from SCD.

src/session.rs
```diff
@@ -488,6 +431,179 @@ impl Session {
...
+    /// We diff the old and new sets and perform the following operations:
+    ///
+    /// - Detect which values should be added and add them (for both used and unused sets),
+    /// - Detect which values should be removed and remove them (for both used
+    ///   and unused sets),
+    /// - Detect which values should be retained, but they changed, we both
+    ///   remove and add them to notify the outside of the change (for both used
+    ///   and unused sets),
+    /// - Detect which values should be transitioned from the used set to unused
+    ///   and miami vice versa.
```
ondrowan: Seriously? 😄

src/ui.rs
```diff
@@ -40,6 +39,16 @@ pub enum Theme {
     Funky,
 }
 
+/// The draw mode applied to a group of objects in the viewport. Not always a
+/// 1:1 mapping with renderer materials.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum ViewportDrawMode {
```
ondrowan, line 45: This doesn't need to be `Eq` or `Hash`.

ondrowan (dismissed):
I give this a quick look and found just 2 tiny things.

janper (changes requested):
1. Would it be possible to disable shadows for zombies?

2. Came across this (see the shadows) 
without zombies
image

with zombies
image

janper:
I see the self-transparency issue now 
image
nevermind, i can live with it

yanchith:
> Would it be possible to disable shadows for zombies?

Yes. I'll do so.

> Came across this (see the shadows)

This probably wasn't introduced in this branch. It seems to be related to how we compute the bounding box that for what gets rendered into the shadow map. I'll create an issue in trello to solve that.

src/session.rs
yanchith: https://thumbs.gfycat.com/RemarkableTediousDwarfmongoose-small.gif

src/ui.rs
yanchith: Hmmm, I can remove `Hash`, but I usually prefer to keep `Eq` to state future intentions, that this enum will always be `Eq`.

janper (approved):

janper:
Absolutely perfect now! And it also makes sense and feels intuitive. Thank you!

PR #150, opened 2020-02-23
Adjust texts to match printed manual, add styling to new elements

janper:
Some texts are now returned to their former version, some parts are a compromise between old and new versions (mainly words "sequence" and "pipeline").
The official name of the pipeline is now "Operation pipeline" with a subtitle "sequence of operations".

This is mainly due to a printed manual, that hasn't been updated and still uses the name "Operation pipeline". This compromise can remain in the master.

### DIFF
--- a/src/ui.rs
+++ b/src/ui.rs
@@ -744,7 +744,7 @@ impl<'a> UiFrame<'a> {
                 if ui.is_item_hovered() {
                     ui.tooltip(|| {
                         let wrap_token = ui.push_text_wrap_pos(WRAP_POS_TOOLTIP_TEXT_PIXELS);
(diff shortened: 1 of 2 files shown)


src/ui.rs
```diff
@@ -732,12 +744,12 @@ impl<'a> UiFrame<'a> {
                 if ui.is_item_hovered() {
                     ui.tooltip(|| {
                         let wrap_token = ui.push_text_wrap_pos(WRAP_POS_TOOLTIP_TEXT_PIXELS);
-                        ui.text_colored(self.colors.tooltip_text, "SAVE PROJECT INTO A .hurban FILE\n\
+                        ui.text_colored(self.colors.tooltip_text, "SAVE PROJECT INTO THE CURENT A .hurban FILE\n\
```
yanchith: CURRENT

yanchith (approved):

PR #154, opened 2020-03-05
First part of Scalar Filed merging process

janper:
This is the bare minimum of the actual Scalar field feature.

Please, let me know if you need more info or if we should meet in person prior to the review. I believe everything is written in the comments though and there is not much extra insight you wouldn't already know.


src/interpreter_funcs/voxelize.rs
```diff
@@ -102,7 +107,7 @@ impl Func for FuncVoxelize {
                 In some cases not growing the volume at all may result in \
                 a non manifold voxelized mesh.",
                 refinement: ParamRefinement::Uint(UintParamRefinement {
-                    default_value: Some(2),
+                    default_value: Some(0),
```
yanchith, line 109: Thank god.

src/interpreter_funcs/voxelize.rs
```diff
@@ -158,57 +163,63 @@ impl Func for FuncVoxelize {
...
-        let mut voxel_cloud = VoxelCloud::from_mesh(mesh, &Vector3::from(voxel_dimensions));
-        for _ in 0..growth_iterations {
-            voxel_cloud.grow_volume();
-        }
+        let mut scalar_field = ScalarField::from_mesh(mesh, &voxel_dimensions, 0_i16, growth_u32);
 
-        if fill {
-            voxel_cloud.fill_volumes();
-        }
+        scalar_field.compute_distance_filed(&(0..=0));
```
yanchith: Probably makes sense to force the `Copy` bound on the parameter and take it by value for better ergonomics, since the ranges will always be small copyable structs.

src/mesh/voxel_cloud.rs
```diff
@@ -1,38 +1,58 @@
...
     block_start: Point3<i32>,
     block_dimensions: Vector3<u32>,
     voxel_dimensions: Vector3<f32>,
-    // FIXME: @Optimization Change this to a bit vector.
-    voxel_map: Vec<bool>,
+    voxels: Vec<Option<T>>,
 }
 
-impl VoxelCloud {
+impl<T: Bounded + Copy + FromPrimitive + Neg<Output = T> + Num + PartialOrd + ToPrimitive>
```
yanchith: Maybe a where clause would be a bit more readable?

TODO @yanchith : check type params for redundancies

src/mesh/voxel_cloud.rs
```diff
@@ -1,38 +1,58 @@
...
+{
     /// Define a new empty block of voxel space, which begins at
-    /// `block_start`(in discrete voxel units), has dimensions
+    /// `block_start`(in discrete absolute voxel units), has dimensions
     /// `block_dimensions` (in discrete voxel units) and contains voxels sized
-    /// `voxel_dimensions` (in model space units).
+    /// `voxel_dimensions` (in cartesian model-space units).
+    ///
+    /// # Panics
+    /// Panics if any of the voxel dimensions is below or equal to zero.
```
yanchith, line 52: Newline

(some markdown renderers require them after or before certain elements, otherwise, this won't count as a title)

src/mesh/voxel_cloud.rs
```diff
@@ -43,16 +63,21 @@ impl VoxelCloud {
...
-            voxel_map: vec![false; cast_usize(map_length)],
+            voxels,
         }
     }
 
     /// Creates a new empty voxel space from a bounding box defined in cartesian
     /// units.
+    ///
+    /// # Panics
+    /// Panics if any of the voxel dimensions is below or equal to zero.
```
yanchith, line 78: Newline

src/mesh/voxel_cloud.rs
```diff
@@ -77,23 +105,46 @@ impl VoxelCloud {
...
         );
 
-        VoxelCloud::new(&block_start, &block_dimensions, voxel_dimensions)
+        ScalarField::new(&block_start, &block_dimensions, voxel_dimensions)
     }
 
-    /// Creates a voxel cloud from an existing mesh with computed
-    /// occupied voxels.
-    pub fn from_mesh(mesh: &Mesh, voxel_dimensions: &Vector3<f32>) -> Self {
+    /// Creates a scalar field from an existing mesh. The voxels intersecting
```
yanchith: Newline after first sentence

src/mesh/voxel_cloud.rs
```diff
@@ -77,23 +105,46 @@ impl VoxelCloud {
...
+    /// the mesh (volume voxels) will be set to `value_on_mesh_surface`, the
+    /// empty voxels (void voxels) will be set to None. The `growth_offset`
+    /// defines how much bigger the scalar field be when initialized. This is
+    /// useful if the distance field is about to be calculated for purposes of
+    /// voxel growth.
+    ///
+    /// # Panics
+    /// Panics if any of the voxel dimensions is below or equal to zero.
```
yanchith, line 119: Newline

src/mesh/voxel_cloud.rs
```diff
@@ -129,386 +180,145 @@ impl VoxelCloud {
...
+    /// `ScalarField::from_mesh` generates a scalar field, which marks volume
+    /// voxels with value `0`. `compute_distance_field` marks each voxel with a
+    /// value representing the voxel's distance from the original volume,
+    /// therefore the voxels right at the shell of the volume are marked 0, the
+    /// layer around them is marked 1 or -1 (inside closed volumes) etc. Once
+    /// the scalar field is populated with meaningful values, it is possible to
+    /// treat (perform boolean operations or materialize into mesh) on various
+    /// numerical ranges. Such range is specified ad-hoc by parameter
+    /// `volume_value_range`.
+    pub fn contains_voxels_within_range<U>(&self, volume_value_range: &U) -> bool
```
yanchith, line 217: Let's also require `U: Copy` and pass by value. We expect these values to be small and can get a tiny bit of ergonomics back this way.

Same everywhere.

src/mesh/voxel_cloud.rs
```diff
@@ -758,303 +491,316 @@ impl VoxelCloud {
...
+    let voxel_scaling_ratio_1d = voxel_scaling_ratio_3d.cbrt();
+    // When changing the voxel dimensions, also the bounding box dimensions
+    // change, therefore the equation is not simple. Therefore a safe buffer of
+    // 1.1 is a quick fix.
+    // FIXME: Come up with a precise equation
+    current_voxel_dimensions * voxel_scaling_ratio_1d * 1.1
+}
+
+/// Returns true if the value of a voxel is within given value range. Returns
+/// `false` if the voxel value is not within the `value_range` or if the voxel
```
yanchith: "`false`" is quoted but "true" isn't

src/mesh/voxel_cloud.rs
```diff
@@ -648,107 +458,30 @@ impl VoxelCloud {
...
-                        cast_i32(volume_dimensions.z),
-                    );
-                BoundingBox::new(&volume_start, &volume_end)
-            })
-    }
-
-    /// Fill hollow volumes in voxel cloud. The original voxel cloud will be
-    /// mutated. The method flood-fills the outer space with void voxels,
-    /// leaving everything inside volume voxels filled.
+    /// Compute discrete distance field. Each voxel will be set a value equal to
```
yanchith: Newline after first sentence.

src/mesh/voxel_cloud.rs
```diff
@@ -758,303 +491,316 @@ impl VoxelCloud {
...
     }
 }
 
-/// Computes a voxel position relative to the block start (relative
-/// coordinate) from an index to the linear representation of the voxel
-/// block.
+/// Computes a voxel position relative to the block start (relative coordinate)
+/// from an index to the linear representation of the voxel block.
 ///
 /// Returns None if out of bounds.
```
yanchith: No it doesn't

src/mesh/voxel_cloud.rs
```diff
@@ -758,303 +491,316 @@ impl VoxelCloud {
...
+    let x = one_dimensional_i32 % x_dimension_i32;
+    Point3::new(x, y, z)
 }
 
-/// Gets the index to the voxel map from absolute voxel coordinates
-/// (relative to the voxel space origin).
+/// Computes a voxel position relative to the model space origin (absolute
+/// coordinate) from an index to the linear representation of the voxel block.
 ///
 /// Returns None if out of bounds.
```
yanchith: No it doesn't

src/mesh/voxel_cloud.rs
```diff
@@ -758,303 +491,316 @@ impl VoxelCloud {
...
+    relative_voxel_to_absolute_voxel_coordinate(&relative, block_start)
 }
 
-/// Computes a voxel position relative to the model space origin (absolute
-/// coordinate) from an index to the linear representation of the voxel
-/// block.
+/// Computes a voxel position in world space cartesian units from an index to
+/// the linear representation of the voxel block.
 ///
 /// Returns None if out of bounds.
```
yanchith: No it doesn't

src/mesh/voxel_cloud.rs
```diff
@@ -758,303 +491,316 @@ impl VoxelCloud {
...
+    block_start: &Point3<i32>,
 ) -> Point3<i32> {
+    relative_coordinate + block_start.coords
+}
+
+/// Computes the center of a voxel in worlds space cartesian units from voxel
+/// coordinates relative to the voxel block start.
+///
+/// # Panics
+/// Panics if any of the voxel dimensions is zero.
```
yanchith: Newline

src/mesh/voxel_cloud.rs
```diff
@@ -758,303 +491,316 @@ impl VoxelCloud {
...
 
-/// Computes the voxel-space coordinates of a voxel containing the
-/// input point
-#[allow(dead_code)]
-fn cartesian_to_relative_voxel_coords(
+/// Computes the absolute voxel space coordinate of a voxel containing the input
+/// point.
+///
+/// # Panics
+/// Panics if any of the voxel dimensions is equal or below zero.
```
yanchith, line 748: Newline

src/mesh/voxel_cloud.rs
yanchith: The next function requires voxel dimension components to not be equal to or below zero, which seems more correct. Let's update this one,

src/mesh/voxel_cloud.rs
```diff
@@ -758,303 +491,316 @@ impl VoxelCloud {
...
-    );
-    Point3::new(
-        absolute_coords.x as f32 * voxel_dimensions.x,
-        absolute_coords.y as f32 * voxel_dimensions.y,
-        absolute_coords.z as f32 * voxel_dimensions.z,
-    )
+/// Computes an index to the linear representation of the voxel block from
+/// voxel coordinates relative to the voxel space block start.
+///
+/// Returns None if out of bounds.
```
yanchith, line 767: Now only the `to_one_dimensional_coordinate` fns return `None` if out of bounds. I wonder if this shouldn't be unified with the rest somehow.

src/mesh/voxel_cloud.rs
```diff
@@ -1126,325 +872,70 @@ mod tests {
...
 
-        insta::assert_json_snapshot!("sphere_after_voxelization", &voxel_cloud);
+        insta::assert_json_snapshot!("sphere_after_voxelization_into_scalar_field", &scalar_field);
     }
 
     #[test]
-    fn test_voxel_cloud_three_dimensional_to_one_dimensional_and_back_relative() {
-        let voxel_cloud = VoxelCloud::new(
+    fn test_scalar_field_three_dimensional_to_one_dimensional_and_back_relative() {
+        let scalar_field: ScalarField<i16> = ScalarField::new(
```
yanchith: Why does this test need an instance of the scalar field? Wouldn't block dimensions be enough?

yanchith (changes requested):
Very nice overall.

I reviewed everything except for `compute_distance_field`, which I will review soon. Meanwhile we can get started.

src/interpreter_funcs/voxelize.rs
janper: We've got better ways now.

src/interpreter_funcs/voxelize.rs
janper: I need to check if it is possible - I believe this was enforced by Clippy. Let's see

src/mesh/voxel_cloud.rs
janper: Sure, I'll refactor this to `where`.

There shouldn't be any redundancy - I kept adding traits as requested by the compiler.

src/mesh/voxel_cloud.rs
janper: i find it correct. Only one when the dimensional coordinate is out of bounds, the related functions panic.

yanchith (commented):
Finished the whole review now. `compute_distance_field` seems fine, but I can't find a reason in this whole PR why we would want `ScalarField` to be parametric over `T`. What are the future plans?

A couple of thoughts:

- fortunately, `T` cannot be unsigned, because unsigned integers don't implement `Neg` (this would behave very weirdly, if it could)
- I am not exactly sure what is the benefit of using floating point values for the current implementation of `compute_distance_field`, but I do see downsides, such as loss of precision for large numbers and more expensive computation.

Don't we want to just use `i32`?

src/interpreter_funcs/voxelize.rs
janper: image

janper:
> Don't we want to just use `i32`?

No, we don't. We're already using `i16`, `i32` and `f64`. The scalar field offers great opportunities in the future. We already implement simplex noise, which yields values from -1.0 to 1.0
Let's not reduce our opportunities here!

src/mesh/voxel_cloud.rs
```diff
@@ -648,107 +464,32 @@ impl VoxelCloud {
...
+    /// `ScalarField::from_mesh` generates a scalar field, which marks volume
+    /// voxels with value `0`. `compute_distance_field` marks each voxel with a
+    /// value representing the voxel's distance from the original volume,
+    /// therefore the voxels right at the shell of the volume are marked 0, the
+    /// layer around them is marked 1 or -1 (inside closed volumes) etc. Once
+    /// the scalar field is populated with meaningful values, it is possible to
+    /// treat (perform boolean operations or materialize into mesh) on various
+    /// numerical ranges. Such range is specified ad-hoc by parameter
+    /// `volume_value_range`.
+    pub fn compute_distance_filed<U>(&mut self, volume_value_range: &U)
```
yanchith: Typo: `compute_distance_field`

src/interpreter_funcs/voxelize.rs
yanchith: Hmm, very weird, but ranges really do not implement `Copy`, even if `T` is `Copy`. Too bad.

src/mesh/voxel_cloud.rs
yanchith: kk, ranges do not implement `Copy`, we can't do this

src/mesh/voxel_cloud.rs
yanchith: Okay, it's internal functionality anyway

janper:
Ready for another round of review

src/mesh/voxel_cloud.rs
```diff
@@ -760,269 +496,238 @@ impl VoxelCloud {
...
+///
+/// # Panics
+///
+/// Panics if any of the voxel dimensions is equal or below zero.
+fn relative_voxel_to_cartesian_coordinate(
+    relative_coordinate: &Point3<i32>,
+    block_start: &Point3<i32>,
+    voxel_dimensions: &Vector3<f32>,
 ) -> Point3<f32> {
     assert!(
```
yanchith, line 732: My previous comment on this is not resolved. We should be checking for a positive number here instead of just non-zero. The doc comment says we are. But we aren't.

yanchith (changes requested):
Almost there. You just forgot to implement one of the previous comments.

src/mesh/voxel_cloud.rs
```diff
@@ -648,107 +459,32 @@ impl VoxelCloud {
...
+    /// `ScalarField::from_mesh` generates a scalar field, which marks volume
+    /// voxels with value `0`. `compute_distance_field` marks each voxel with a
+    /// value representing the voxel's distance from the original volume,
+    /// therefore the voxels right at the shell of the volume are marked 0, the
+    /// layer around them is marked 1 or -1 (inside closed volumes) etc. Once
+    /// the scalar field is populated with meaningful values, it is possible to
+    /// treat (perform boolean operations or materialize into mesh) on various
+    /// numerical ranges. Such range is specified ad-hoc by parameter
+    /// `volume_value_range`.
+    pub fn compute_distance_fieled<U>(&mut self, volume_value_range: &U)
```
yanchith: Lol. Another typo here.

yanchith:
Also see the clippy error (only triggered now because the generic bounds were removed). I should have noticed that, but didn't. To explain: You almost never want `&Option<Foo>`, because `Option` is a model example of type you never want to pass by reference. Use instead `Option<&Foo>`. There is a very easy to transpose between the two.

janper:
> There is a very easy to transpose between the two.

Can you please be more cryptic?

image

yanchith:
> Use instead Option<&Foo>. There is a very easy to transpose between the two.

I am sorry I mislead you. The current situation doesn't require `Option<&Foo>` - I was speaking in general terms. The transposition I mentioned is `Option::as_ref` (https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.as_ref), which takes a `&Option<T>` and gives you `Option<&T>`.

janper:
ready again

yanchith (approved):

PR #155, opened 2020-03-05
Misc tweaks that should survive after the MAK exhibition

janper:
See inline comments for more details about the individual tweaks

### DIFF
--- a/src/interpreter/ast.rs
+++ b/src/interpreter/ast.rs
@@ -132,7 +132,6 @@ pub enum Expr {
     // arg would be the direct result of another func call. We use
     // `VarExpr` instead to refer to a previous result.
     Lit(LitExpr),
-    #[allow(dead_code)]
     Var(VarExpr),
 }
 
@@ -158,9 +157,7 @@ impl fmt::Display for Expr {
 #[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
 pub enum LitExpr {
     Nil,
-    #[allow(dead_code)]
     Boolean(bool),
-    #[allow(dead_code)]
     Int(i32),
     Uint(u32),
     Float(f32),
--- a/src/interpreter_funcs/voxelize.rs
+++ b/src/interpreter_funcs/voxelize.rs
@@ -12,12 +12,13 @@ use crate::interpreter::{
 };
 use crate::mesh::voxel_cloud::VoxelCloud;
 
-const VOXEL_COUNT_THRESHOLD: u32 = 50000;
+const VOXEL_COUNT_THRESHOLD: u32 = 100_000;
 
 #[derive(Debug, PartialEq)]
 pub enum FuncVoxelizeError {
     WeldFailed,
     EmptyVoxelCloud,
+    VoxelDimensionsZero,
     TooManyVoxels(u32, f32, f32, f32),
 }
 
@@ -29,6 +30,7 @@ impl fmt::Display for FuncVoxelizeError {
                 "Welding of separate voxels failed due to high welding proximity tolerance"
             ),
             FuncVoxelizeError::EmptyVoxelCloud => write!(f, "The resulting voxel cloud is empty"),
+            FuncVoxelizeError::VoxelDimensionsZero => write!(f, "One or more voxel dimensions are zero"),
             FuncVoxelizeError::TooManyVoxels(max_count, x, y, z) => write!(
                 f,
                 "Too many voxels. Limit set to {}. Try setting voxel size to [{:.3}, {:.3}, {:.3}] or more.",
@@ -84,7 +86,7 @@ impl Func for FuncVoxelize {
                 heavier geometry that significantly affect performance. Too high values produce \
                 single large voxel, too low values may generate holes in the resulting geometry.",
                 refinement: ParamRefinement::Float3(Float3ParamRefinement {
-                    min_value: Some(f32::MIN_POSITIVE),
+                    min_value: Some(0.005),
                     max_value: None,
                     default_value_x: Some(1.0),
                     default_value_y: Some(1.0),
@@ -158,12 +160,21 @@ impl Func for FuncVoxelize {
         log: &mut dyn FnMut(LogMessage),
     ) -> Result<Value, FuncError> {
         let mesh = args[0].unwrap_mesh();
-        let voxel_dimensions = args[1].unwrap_float3();
+        let voxel_dimensions = Vector3::from(args[1].unwrap_float3());
         let growth_iterations = args[2].unwrap_uint();
         let fill = args[3].unwrap_boolean();
         let error_if_large = args[4].unwrap_boolean();
         let analyze_bbox = args[5].unwrap_boolean();
         let analyze_mesh = args[6].unwrap_boolean();
+
+        if voxel_dimensions
+            .iter()
+            .any(|dimension| approx::relative_eq!(*dimension, 0.0))
+        {
+            let error = FuncError::new(FuncVoxelizeError::VoxelDimensionsZero);
+            log(LogMessage::error(format!("Error: {}", error)));
+            return Err(error);
+        }
 
         let bbox_diagonal = mesh.bounding_box().diagonal();
         let voxel_count = (bbox_diagonal.x / voxel_dimensions[0]).ceil() as u32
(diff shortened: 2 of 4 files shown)


src/interpreter/ast.rs
```diff
@@ -132,7 +132,6 @@ pub enum Expr {
     // arg would be the direct result of another func call. We use
     // `VarExpr` instead to refer to a previous result.
     Lit(LitExpr),
-    #[allow(dead_code)]
```
janper, line 135: No more dead (or at least Clippy doesn't complain)

src/interpreter/ast.rs
```diff
@@ -158,9 +157,7 @@ impl fmt::Display for Expr {
 #[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
 pub enum LitExpr {
     Nil,
-    #[allow(dead_code)]
```
janper, line 161: No more dead (or at least Clippy doesn't complain)

src/interpreter/ast.rs
```diff
@@ -158,9 +157,7 @@ impl fmt::Display for Expr {
...
     Boolean(bool),
-    #[allow(dead_code)]
```
janper, line 163: No more dead (or at least Clippy doesn't complain)

src/interpreter_funcs/voxelize.rs
```diff
@@ -12,12 +12,13 @@ use crate::interpreter::{
 };
 use crate::mesh::voxel_cloud::VoxelCloud;
 
-const VOXEL_COUNT_THRESHOLD: u32 = 50000;
+const VOXEL_COUNT_THRESHOLD: u32 = 100_000;
```
janper, line 15: We can afford more on the NUC, which is our reference hardware. Actually, we can afford even more than this (10M voxel operations take about 15s).

src/interpreter_funcs/voxelize.rs
```diff
@@ -12,12 +12,13 @@ use crate::interpreter::{
...
 
 #[derive(Debug, PartialEq)]
 pub enum FuncVoxelizeError {
     WeldFailed,
     EmptyVoxelCloud,
+    VoxelDimensionsZero,
```
janper, line 21: Forgotten erroneous scenario.

src/interpreter_funcs/voxelize.rs
```diff
@@ -84,7 +86,7 @@ impl Func for FuncVoxelize {
                 heavier geometry that significantly affect performance. Too high values produce \
                 single large voxel, too low values may generate holes in the resulting geometry.",
                 refinement: ParamRefinement::Float3(Float3ParamRefinement {
-                    min_value: Some(f32::MIN_POSITIVE),
+                    min_value: Some(0.005),
```
janper, line 89: Small values, even though they may not generate too many voxels, cause the calculation to be too small. It's because the `to_mesh` function needs to divide each face into chunks as small as the smallest voxel dimension, which makes the voxelization too slow. `0.005` is an arbitrary number but seems to work well. The minimum value is being set very often because the user drags the slider all the was to the left quite often.

src/interpreter_funcs/voxelize.rs
```diff
@@ -158,13 +160,22 @@ impl Func for FuncVoxelize {
         log: &mut dyn FnMut(LogMessage),
     ) -> Result<Value, FuncError> {
         let mesh = args[0].unwrap_mesh();
-        let voxel_dimensions = args[1].unwrap_float3();
+        let voxel_dimensions = Vector3::from(args[1].unwrap_float3());
```
janper, line 163: The rest of the code is slightly nicer when the voxel dimensions are already stored in a Vector

src/interpreter_funcs/voxelize.rs
```diff
@@ -158,13 +160,22 @@ impl Func for FuncVoxelize {
...
 
+        if voxel_dimensions
+            .iter()
+            .any(|dimension| approx::relative_eq!(*dimension, 0.0))
+        {
+            let error = FuncError::new(FuncVoxelizeError::VoxelDimensionsZero);
+            log(LogMessage::error(format!("Error: {}", error)));
+            return Err(error);
+        }
+
```
janper, line 178: Check if the vector dimensions are above zero. Should never happen thanks to the GUI settings. Maybe can be turned into an assert and panic if it happens. I didn't want to risk it at the exhibition though.

src/interpreter_funcs/voxelize.rs
```diff
@@ -173,8 +184,8 @@ impl Func for FuncVoxelize {
         log(LogMessage::info(format!("Voxel count = {}", voxel_count)));
 
         if error_if_large && voxel_count > VOXEL_COUNT_THRESHOLD {
-            let vy_over_vx = voxel_dimensions[1] / voxel_dimensions[0];
-            let vz_over_vx = voxel_dimensions[2] / voxel_dimensions[0];
+            let vy_over_vx = voxel_dimensions.y / voxel_dimensions.x;
+            let vz_over_vx = voxel_dimensions.z / voxel_dimensions.x;
```
janper, line 188: This is the nicer part. 
However, this is a very temporary piece of code because it's being replaced by a more precise method in PR#154

src/lib.rs
```diff
@@ -182,7 +182,7 @@ pub fn init_and_run(options: Options) -> ! {
 
         winit::window::WindowBuilder::new()
             .with_title(BASE_WINDOW_TITLE)
-            .with_inner_size(winit::dpi::LogicalSize::new(1280.0, 720.0))
+            .with_maximized(true)
```
janper, line 185: Start with a maximized window

src/lib.rs
```diff
@@ -206,6 +206,12 @@ pub fn init_and_run(options: Options) -> ! {
 
     change_window_title(&window, &project_status);
 
+    let initial_window_size = window.inner_size().to_physical(window.hidpi_factor());
+    let initial_window_width = initial_window_size.width.round() as u32;
+    let initial_window_height = initial_window_size.height.round() as u32;
+    let initial_window_aspect_ratio =
+        initial_window_size.width as f32 / initial_window_size.height as f32;
+
```
janper, line 208: We don't know the windows size upfront because fullscreen can be anything.

src/ui.rs
```diff
@@ -37,6 +37,8 @@ const SUBDIGITAL_LOGO_WINDOW_WIDTH: f32 = 100.0;
 
 const ABOUT_WINDOW_WIDTH: f32 = 600.0;
 
+const DRAG_SPEED: f32 = 0.01;
```
janper, line 40: Make this a global constant so that we can change it easily at one place.

src/ui.rs
```diff
@@ -498,7 +500,7 @@ impl<'a> UiFrame<'a> {
             .movable(true)
             .resizable(false)
             .collapsible(false)
-            .always_auto_resize(true)
+            .always_auto_resize(false)
```
janper, line 503: This makes the about window fixed height and therefore it doesn't extend the window boundaries (so easily).

src/ui.rs
```diff
@@ -528,46 +530,25 @@ impl<'a> UiFrame<'a> {
...
-                    space for selection and algorithmic seeking of new forms. \n\
-                    \n\
-                    H.U.R.B.A.N. Selector is currently under development. It utilizes the Rust \
-                    programming language, and its source code is freely available to the community."));
+                    compositions.\n\
+                     \n\
+                     The software is currently in very early development stages. It strives to be \
+                     a tool for simple parametric modeling using various \
+                     hybridization strategies for mesh and voxel models, allowing designers to \
+                     smoothly interpolate between multiple source models."));
```
janper: The description has changed. This is what we exhibit.

src/ui.rs
```diff
@@ -2077,6 +2057,10 @@ impl<'a> UiFrame<'a> {
                 imgui::StyleColor::HeaderActive,
                 self.colors.combo_box_selected_item_active,
             ),
+            (
+                imgui::StyleColor::PopupBg,
+                self.colors.popup_window_background,
+            ),
```
janper: Fixes the orange background of the dropdown menus.

src/interpreter_funcs/voxelize.rs
yanchith: Interesting thought, but let's pretend the UI can't validate anything. This will make us future-proof for other frontends.

src/lib.rs
yanchith: We'll need to test this on macOS before merging. If this works well, I think we can get rid of the fullscreen (`HS_FULLSCREEN=1`) option completely. The removal can be a separate PR.

src/ui.rs
yanchith: formatting

yanchith: Should we also update the initial README section to this text?

yanchith (commented):
Nice! Almost no objections, see inline comments for questions.

src/interpreter_funcs/voxelize.rs
janper: That's a relief! Better safe than sorry!

src/lib.rs
janper: I'd like to keep the fullscreen option too. May be useful for some showcases.

src/ui.rs
janper: Not yet. We'll come back to it before the end of the project. 
We'll have to mention the sponsors too... Let's postpone.

src/lib.rs
yanchith: Just tested on linux (X11) :heavy_check_mark:

yanchith (approved):
I'll merge this in the evening once I test on macOS.

janper:
Please, merge #151 and #152 first. There may be conflicts.

PR #156, opened 2020-03-14
Add Boolean operations and their supplemental code

janper:


src/math.rs
```diff
@@ -1,3 +1,8 @@
+use std::ops::Bound::*;
```
yanchith: Let's not import the variants globally. Import them just in the function that uses them.

src/math.rs
```diff
@@ -29,6 +34,39 @@ pub fn lerp(source: f32, target: f32, weight: f32) -> f32 {
...
+        if let Included(source_end) | Excluded(source_end) = source_range.end_bound() {
+            if let Included(target_start) | Excluded(target_start) = target_range.start_bound() {
+                if let Included(target_end) | Excluded(target_end) = target_range.end_bound() {
+                    let source_start_f64 = source_start.to_f64().expect("Can't convert to f64");
+                    let source_end_f64 = source_end.to_f64().expect("Can't convert to f64");
+                    let target_start_f64 = target_start.to_f64().expect("Can't convert to f64");
+                    let target_end_f64 = target_end.to_f64().expect("Can't convert to f64");
+                    let length_source_f64 = source_end_f64 - source_start_f64;
+                    if approx::relative_eq!(length_source_f64, 0.0) {
+                        return T::from_f64((target_start_f64 + target_end_f64) / 2_f64);
```
yanchith: This is inconsistent with the documentation. It actually returns `None` if the result can't be converted into `T` (but the ranges are finite). I prefer not panicking on failed conversion to `T` though, so let's update the documentation instead.

Note that we can still `expect` conversions to `f64`. To my knowledge, any of the the `to_f64` or `to_f32` implementations currently can't fail, even if the API says otherwise.

src/math.rs
```diff
@@ -29,6 +34,39 @@ pub fn lerp(source: f32, target: f32, weight: f32) -> f32 {
...
+                    }
+                    let length_target_f64 = target_end_f64 - target_start_f64;
+
+                    let value_f64 = value.to_f64().expect("Can't convert to f64");
+                    let remapped = target_start_f64
+                        + ((value_f64 - source_start_f64) / length_source_f64) * length_target_f64;
+
+                    return Some(T::from_f64(remapped).expect("Can't convert from f64"));
```
yanchith: Similarly to above. Let's not panic on failed conversion to `T` but return `None` instead.

src/math.rs
```diff
@@ -29,6 +34,39 @@ pub fn lerp(source: f32, target: f32, weight: f32) -> f32 {
     source + weight * (target - source)
 }
 
+/// Remaps the value from one range to another. Return None if one of the
+/// ranges is infinite.
+pub fn remap<T, U>(value: T, source_range: &U, target_range: &U) -> Option<T>
```
yanchith: I am half thinking we should not be generic here, but just do this for floats.

The reason is we can't specify the conversion strategy: `num_traits` delegates to the `as` operator, which is does truncation (not rounding) towards zero.

```rust
    assert_eq!(3, 3.80 as u32);
    assert_eq!(Some(3), u32::from_f64(3.80));
    assert_eq!(-3, -3.80 as i32);
    assert_eq!(Some(-3), i32::from_f64(-3.80));
```

I don't think this is what we would expect. Thoughts?

If we decide to go the floating point road here, the other comments regarding this function are irrelevant.

src/interpreter_funcs/voxel_boolean_difference.rs
```diff
@@ -0,0 +1,262 @@
...
+                name: "Bounding Box Analysis",
+                description: "Reports basic and quick analytic information on the created mesh.",
+                refinement: ParamRefinement::Boolean(BooleanParamRefinement {
+                    default_value: true,
+                }),
+                optional: false,
+            },
+            ParamInfo {
+                name: "Detailed Mesh Analysis",
+                description: "Reports detailed analytic information on the created mesh.\n\
```
yanchith, line 155: This "\n" on purpose?

src/interpreter_funcs/voxel_boolean_difference.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    fn call(
+        &mut self,
+        args: &[Value],
+        log: &mut dyn FnMut(LogMessage),
+    ) -> Result<Value, FuncError> {
+        let mesh1 = args[0].unwrap_mesh();
+        let mesh2 = args[1].unwrap_mesh();
+        let voxel_dimensions = Vector3::from(args[1].unwrap_float3());
```
yanchith: `args[2]`, this crashes

src/interpreter_funcs/voxel_boolean_difference.rs
```diff
@@ -0,0 +1,262 @@
...
+        let growth_u32 = args[3].unwrap_uint();
+        let growth_f32 = growth_u32 as f32;
+        let fill = args[4].unwrap_boolean();
+        let error_if_large = args[5].unwrap_boolean();
+        let analyze_bbox = args[6].unwrap_boolean();
+        let analyze_mesh = args[7].unwrap_boolean();
+
+        if voxel_dimensions
+            .iter()
+            .any(|dimension| approx::relative_eq!(*dimension, 0.0))
```
yanchith: Similarly to previous changes of this kind, shouldn't we also be checking for negative numbers?

src/interpreter_funcs/voxel_boolean_difference.rs
```diff
@@ -0,0 +1,262 @@
...
+        let voxel_count1 = voxel_cloud::evaluate_voxel_count(&bbox1, &voxel_dimensions);
+
+        let bbox2 = mesh2.bounding_box();
+        let voxel_count2 = voxel_cloud::evaluate_voxel_count(&bbox2, &voxel_dimensions);
+
+        let voxel_count = if voxel_count1 > voxel_count2 {
+            voxel_count1
+        } else {
+            voxel_count2
+        };
```
yanchith: `let voxel_count = voxel_count1.max(voxel_count2);`

src/interpreter_funcs/voxel_boolean_difference.rs
```diff
@@ -0,0 +1,262 @@
...
+                VOXEL_COUNT_THRESHOLD,
+                suggested_voxel_size.x,
+                suggested_voxel_size.y,
+                suggested_voxel_size.z,
+            ));
+            log(LogMessage::error(format!("Error: {}", error)));
+            return Err(error);
+        }
+
+        let mut voxel_cloud1 = ScalarField::from_mesh(mesh1, &voxel_dimensions, 0.0, growth_u32);
```
yanchith, line 218: Shouldn't we name these `scalar_field`? Also for the other 2 operations.

src/interpreter_funcs/voxel_boolean_intersection.rs
```diff
@@ -0,0 +1,261 @@
...
+pub struct FuncBooleanIntersection;
+
+impl Func for FuncBooleanIntersection {
+    fn info(&self) -> &FuncInfo {
+        &FuncInfo {
+            name: "Intersection",
+            description: "BOOLEAN INTERSECTION OF VOXEL CLOUDS FROM TWO MESH GEOMETRIES\n\
+            \n\
+            Converts the input mesh geometries into voxel clouds, then performs \
+            boolean intersection of the first to second voxel clouds and eventually \
```
yanchith: "first to second" -> "first and second? Unlike difference, this is commutative.

src/interpreter_funcs/voxel_boolean_difference.rs
```diff
@@ -0,0 +1,262 @@
...
+        &FuncInfo {
+            name: "Difference",
+            description: "BOOLEAN DIFFERENCE OF VOXEL CLOUDS FROM TWO MESH GEOMETRIES\n\
+            \n\
+            Converts the input mesh geometries into voxel clouds, then performs \
+            boolean difference of the first to second voxel clouds and eventually \
+            materializes the resulting voxel cloud into a welded mesh. \
+            Boolean difference removes volume of the first geometry where there is \
+            a volume in the second geometry (removes the second volume from the first one). \
+            It is equivalent to arithmetic subtraction.\n\
```
yanchith: Not equivalent, but analogous.

src/interpreter_funcs/voxel_boolean_intersection.rs
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    fn call(
+        &mut self,
+        args: &[Value],
+        log: &mut dyn FnMut(LogMessage),
+    ) -> Result<Value, FuncError> {
+        let mesh1 = args[0].unwrap_mesh();
+        let mesh2 = args[1].unwrap_mesh();
+        let voxel_dimensions = Vector3::from(args[1].unwrap_float3());
```
yanchith: `args[2]`, this crashes.

src/interpreter_funcs/voxel_boolean_intersection.rs
```diff
@@ -0,0 +1,261 @@
...
+        let growth_u32 = args[3].unwrap_uint();
+        let growth_f32 = growth_u32 as f32;
+        let fill = args[4].unwrap_boolean();
+        let error_if_large = args[5].unwrap_boolean();
+        let analyze_bbox = args[6].unwrap_boolean();
+        let analyze_mesh = args[7].unwrap_boolean();
+
+        if voxel_dimensions
+            .iter()
+            .any(|dimension| approx::relative_eq!(*dimension, 0.0))
```
yanchith: Same as in "Difference", let's include the possibility of negative numbers in the check.

src/interpreter_funcs/voxel_boolean_intersection.rs
```diff
@@ -0,0 +1,261 @@
...
+        let voxel_count1 = voxel_cloud::evaluate_voxel_count(&bbox1, &voxel_dimensions);
+
+        let bbox2 = mesh2.bounding_box();
+        let voxel_count2 = voxel_cloud::evaluate_voxel_count(&bbox2, &voxel_dimensions);
+
+        let voxel_count = if voxel_count1 > voxel_count2 {
+            voxel_count1
+        } else {
+            voxel_count2
+        };
```
yanchith: `let voxel_count = voxel_count1.max(voxel_count2)`;

src/interpreter_funcs/voxel_boolean_union.rs
```diff
@@ -0,0 +1,261 @@
...
+pub struct FuncBooleanUnion;
+
+impl Func for FuncBooleanUnion {
+    fn info(&self) -> &FuncInfo {
+        &FuncInfo {
+            name: "Union",
+            description: "BOOLEAN UNION OF VOXEL CLOUDS FROM TWO MESH GEOMETRIES\n\
+            \n\
+            Converts the input mesh geometries into voxel clouds, then performs \
+            boolean union of the first to second voxel clouds and eventually \
```
yanchith: Likewise: maybe we should be saying "first and second" here, because union commutes.

src/interpreter_funcs/voxel_boolean_union.rs
```diff
@@ -0,0 +1,261 @@
...
+                name: "Bounding Box Analysis",
+                description: "Reports basic and quick analytic information on the created mesh.",
+                refinement: ParamRefinement::Boolean(BooleanParamRefinement {
+                    default_value: true,
+                }),
+                optional: false,
+            },
+            ParamInfo {
+                name: "Detailed Mesh Analysis",
+                description: "Reports detailed analytic information on the created mesh.\n\
```
yanchith, line 158: This "\n" on purpose?

src/interpreter_funcs/voxel_boolean_intersection.rs
```diff
@@ -0,0 +1,261 @@
...
+                name: "Bounding Box Analysis",
+                description: "Reports basic and quick analytic information on the created mesh.",
+                refinement: ParamRefinement::Boolean(BooleanParamRefinement {
+                    default_value: true,
+                }),
+                optional: false,
+            },
+            ParamInfo {
+                name: "Detailed Mesh Analysis",
+                description: "Reports detailed analytic information on the created mesh.\n\
```
yanchith, line 154: This "\n" on purpose?

src/interpreter_funcs/voxel_boolean_union.rs
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    fn call(
+        &mut self,
+        args: &[Value],
+        log: &mut dyn FnMut(LogMessage),
+    ) -> Result<Value, FuncError> {
+        let mesh1 = args[0].unwrap_mesh();
+        let mesh2 = args[1].unwrap_mesh();
+        let voxel_dimensions = Vector3::from(args[1].unwrap_float3());
```
yanchith: Nice copy pasta. Use `args[2]`. This crashes.

src/interpreter_funcs/voxel_boolean_union.rs
```diff
@@ -0,0 +1,261 @@
...
+        let growth_f32 = growth_u32 as f32;
+        let fill = args[4].unwrap_boolean();
+        let error_if_large = args[5].unwrap_boolean();
+        let analyze_bbox = args[6].unwrap_boolean();
+        let analyze_mesh = args[7].unwrap_boolean();
+
+        if voxel_dimensions
+            .iter()
+            .any(|dimension| approx::relative_eq!(*dimension, 0.0))
+        {
```
yanchith: Similarly to the other operations added - let's take into account negative values here.

src/interpreter_funcs/voxel_boolean_union.rs
```diff
@@ -0,0 +1,261 @@
...
+            return Err(error);
+        }
+
+        let mut voxel_cloud1 = ScalarField::from_mesh(mesh1, &voxel_dimensions, 0.0, growth_u32);
+        let mut voxel_cloud2 = ScalarField::from_mesh(mesh2, &voxel_dimensions, 0.0, growth_u32);
+
+        voxel_cloud1.compute_distance_field(&(0.0..=0.0));
+        voxel_cloud2.compute_distance_field(&(0.0..=0.0));
+
+        let boolean_union_range = &(-growth_f32..=growth_f32);
```
yanchith: Maybe I am missing something, but why is this different from the other 2 operations? Shouldn't the `meshing_range` be ok provided that `ScalarField::compute_distance_field` correctly computes what is inside and what is outside?

yanchith (changes requested):
Here comes the first part of the review (everything except `voxel_cloud.rs` changes). You can work on the first batch of issues while the review continues.

src/math.rs
janper: not panicking anymore

src/math.rs
janper: what about returning `f64` all the time? ergonomy+correctness meeting you half way...

src/interpreter_funcs/voxel_boolean_difference.rs
janper: we to that everywhere. i want ti to be in a new line and having an empty line with only two lines altogether looks ugly

src/interpreter_funcs/voxel_boolean_difference.rs
janper: you used to prefer the `if` version, didn't you?

src/interpreter_funcs/voxel_boolean_difference.rs
janper: there will be a separate PR for renaming the voxel cloud to scalar field everywhere

src/interpreter_funcs/voxel_boolean_difference.rs
janper: are you sure?

src/interpreter_funcs/voxel_boolean_union.rs
janper: yes, maybe it's even more correct because using the boolean union range ignores the insides of the second scalar field

src/math.rs
yanchith: You mean taking anything, but returning floating point? Yes, that would work. Don't need `num_traits` for that, though. `std::convert::From` will work well enough. There are implementations for `u8`, `u16`, `u32`, `i8`, `i16`, `i32`, `f32` that can safely convert to `f64`.

https://doc.rust-lang.org/stable/std/convert/trait.From.html

src/interpreter_funcs/voxel_boolean_difference.rs
yanchith: Can't recall that, tbh. We for sure do have some `min` and `max` elsewhere, so why not here as well.

src/interpreter_funcs/voxel_boolean_difference.rs
yanchith: No, I am not. But it feels more correct. Equivalence is quite a strong term.

src/math.rs
```diff
@@ -34,32 +33,43 @@ pub fn lerp(source: f32, target: f32, weight: f32) -> f32 {
     source + weight * (target - source)
 }
 
-/// Remaps the value from one range to another. Return None if one of the
-/// ranges is infinite.
+/// Remaps the value from one range to another. Return None if conversion of the
+/// value type to f64 fails.
```
yanchith: Returning `None` for infinite ranges is still true I think. Let's keep that in the doc comment as well.

src/math.rs
janper: What is the type / traits of T then?

janper: nevermind, it's done

src/mesh/voxel_cloud.rs
```diff
@@ -200,6 +201,23 @@ impl ScalarField {
         scalar_field
     }
 
+    /// Clears the scalar field, sets its block dimensions to zero.
+    pub fn wipe(&mut self) {
+        self.block_start = Point3::origin();
```
yanchith, line 217: Not sure if important, but should this be resetting `block_start`?

src/mesh/voxel_cloud.rs
```diff
@@ -463,6 +481,322 @@ impl ScalarField {
...
+        &mut self,
+        volume_value_range_self: &U,
+        other: &ScalarField,
+        volume_value_range_other: &U,
+    ) where
+        U: RangeBounds<f32>,
+    {
+        // Find volume common to both scalar fields.
+        if let (Some(self_volume_bounding_box), Some(other_volume_bounding_box)) = (
+            self.volume_bounding_box(volume_value_range_self),
```
yanchith: Since we have `resize_to_voxel_space_bounding_box`, shouldn't this be called `voxel_space_bounding_box` or `volume_voxel_space_bouding_box`?

src/mesh/voxel_cloud.rs
```diff
@@ -463,6 +481,322 @@ impl ScalarField {
...
+                &original_block_dimensions,
+            ) {
+                Some(original_one_dimensional) => original_values[original_one_dimensional],
+                None => None,
+            };
+        }
+    }
+
+    /// Resize the scalar field block to exactly fit the volumetric geometry.
+    /// Returns None for empty the scalar field.
```
yanchith: This doesn't return anything.

src/mesh/voxel_cloud.rs
```diff
@@ -463,6 +481,322 @@ impl ScalarField {
...
+    /// scalar field should be considered to be a volume. The
+    /// `ScalarField::from_mesh` generates a scalar field, which marks volume
+    /// voxels with value `0`. `compute_distance_field` marks each voxel with a
+    /// value representing the voxel's distance from the original volume,
+    /// therefore the voxels right at the shell of the volume are marked 0, the
+    /// layer around them is marked 1 or -1 (inside closed volumes) etc. Once
+    /// the scalar field is populated with meaningful values, it is possible to
+    /// treat (perform boolean operations or materialize into mesh) on various
+    /// numerical ranges. Such range is specified ad-hoc by parameter
+    /// `volume_value_range`.
```
yanchith: This comment is admirable, but let's have it in only one place. Maybe the struct definition of `ScalarField` is a good place? You may have to adjust the comment to fit naturally in the new place.

src/mesh/voxel_cloud.rs
```diff
@@ -463,6 +481,322 @@ impl ScalarField {
...
+            return;
+        }
+
+        // Wipe if resizing to an empty scalar field.
+        if resized_block_dimensions == &Vector3::zeros() {
+            self.wipe();
+            return;
+        }
+
+        let original_values = self.voxels.clone();
```
yanchith: Just noticed we could do this with one less allocation. Currently there are two:

- this clone
- `self.voxels.resize` down below.

We could instead create the new voxels only, with correct len from the start, fill them, and then just assign them to replace the original.

You might want to restructure some surrounding lines of code to make that more readable, e.g. instead of having `original_*` and mutating the `self` values, you can have `new_*`, mutate those, and in the end replace the originals in `self`.

src/mesh/voxel_cloud.rs
```diff
@@ -620,6 +954,145 @@ impl ScalarField {
...
+                absolute_max.z,
+                i32::min_value(),
+                "scalar field emptiness check failed"
+            );
+            None
+        } else {
+            let block_dimensions = Vector3::new(
+                clamp_cast_i32_to_u32(absolute_max.x - absolute_min.x + 1),
+                clamp_cast_i32_to_u32(absolute_max.y - absolute_min.y + 1),
+                clamp_cast_i32_to_u32(absolute_max.z - absolute_min.z + 1),
```
yanchith: Clamp casting doesn't seem correct in this case. The logic of this function, if correct, should always ensure that max point's coords are greater than min point's coords, so `absolute_max.x - absolute_min.x + 1` should always be at least 1.

`cast_u32` should be safe to use. If it panics, it means this function has an error, which we'd rather discover than hide.

src/mesh/voxel_cloud.rs
```diff
@@ -620,6 +954,145 @@ impl ScalarField {
...
+    pub fn volume_bounding_box<U>(&self, volume_value_range: &U) -> Option<BoundingBox<i32>>
+    where
+        U: RangeBounds<f32>,
+    {
+        self.compute_volume_boundaries(volume_value_range).map(
+            |(volume_start, volume_dimensions)| {
+                let volume_end = volume_start
+                    + Vector3::new(
+                        // Voxels occupy also the end voxel position in the
+                        // grid, hence +1.
```
yanchith: Hmm I have a sense of dejavu here. Is this a good idea?

Voxel cloud containing a single voxel (0, 0, 0) has volume_start: (0, 0, 0) and volume_dimensions: (1, 1, 1). This, I believe should construct a bounding box [min: (0, 0, 0), max: (1, 1, 1)], not [min: (0, 0, 0), max: (2, 2, 2)]. Or am I missing something?

src/mesh/voxel_cloud.rs
```diff
@@ -463,6 +481,322 @@ impl ScalarField {
...
+    /// value representing the voxel's distance from the original volume,
+    /// therefore the voxels right at the shell of the volume are marked 0, the
+    /// layer around them is marked 1 or -1 (inside closed volumes) etc. Once
+    /// the scalar field is populated with meaningful values, it is possible to
+    /// treat (perform boolean operations or materialize into mesh) on various
+    /// numerical ranges. Such range is specified ad-hoc by parameter
+    /// `volume_value_range`.
+    ///
+    /// # Panics
+    /// Panics if one of the volume value ranges is infinite.
```
yanchith, line 527: Missing newline

src/mesh/voxel_cloud.rs
```diff
@@ -463,6 +481,322 @@ impl ScalarField {
...
+    ///
+    /// # Warning
```
yanchith, line 529: Missing newline

src/mesh/voxel_cloud.rs
```diff
@@ -463,6 +481,322 @@ impl ScalarField {
...
+                                    volume_value_range_self,
+                                )
+                                .expect("One of the ranges is infinite.")
+                                    as f32,
+                            );
+                        }
+                    }
+                }
+            }
+            self.shrink_to_fit(volume_value_range_self);
```
yanchith: Does this ever shrink? The bounding boxes are computed with the value ranges in mind, the values are remapped (so they can not end outside the self-range), and we already did a resize to fit the union bounding box. Is there some other case when this would actually shrink?

yanchith (changes requested):
This is part 2 of the review. Now only the added tests for scalar field remain unreviewed.

Nice work overall. See inline for questions and suggestions.

Also: I probably missed this in previous PR review, but not sure what the value of `is_voxel_within_range` and `is_value_at_absolute_voxel_coordinate_within_range` is. The former can be replaced by `.map().unwrap_or()`, and the latter doesn't seem like it should be a part of `ScalarField`, but rather the caller can just check that themselves. Thoughts?

src/mesh/voxel_cloud.rs
janper: that's a matter of decision. generally, it makes no difference, so I decided to set the scalar field to its most default state.

src/mesh/voxel_cloud.rs
janper: you requested and explanation of the `volume_value_range`, so i've added them whereever it appears. it doesn't appear anywhere in the struct itself and it is basically a way of interpreting the values stored in it. are you sure it belongs to the description of the struct?
i don't mind, just asking

src/mesh/voxel_cloud.rs
janper: true. the same applies for the intersection

janper:
> Also: I probably missed this in previous PR review, but not sure what the value of `is_voxel_within_range` and `is_value_at_absolute_voxel_coordinate_within_range` is. The former can be replaced by `.map().unwrap_or()`, and the latter doesn't seem like it should be a part of `ScalarField`, but rather the caller can just check that themselves. Thoughts?

Not sure if what you propose is semantically equivalent. Have a look at this piece of code from `to_mesh`:
```rust
if is_voxel_within_range(*voxel, volume_value_range) {
    // do something
} else {
    // do nothing
}
```

How would you replace this with `.map().unwrap_or()`? I think the body of `is_voxel_within_range` will have to be explicitly stated here like this:

```rust
if Some(voxel_value) = *voxel {
    if value_range.contains(&voxel_value) {
        // do something
    }
}
```

Am I getting this right?

src/mesh/voxel_cloud.rs
yanchith: I don't mind either way. Just asking.

src/mesh/voxel_cloud.rs
yanchith: It would be good if it were either module level or struct level documentation. Have a look at std or any other crate to see how to write module level documentation.

yanchith:
> How would you replace this with .map().unwrap_or()?

Something akin to:

```rust
    let range = 0_f32..0.5_f32;
    let voxel: Option<f32> = Some(0.2);
    
    if voxel.map(|value| range.contains(&value)).unwrap_or(false) {
        println!("Value");
    }
```

It's just my intuition, but I don't think this warrants having it's own function.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&code=fn%20main()%20%7B%0A%20%20%20%20let%20range%20%3D%200_f32..0.5_f32%3B%0A%20%20%20%20let%20voxel%3A%20Option%3Cf32%3E%20%3D%20Some(0.2)%3B%0A%20%20%20%20%0A%20%20%20%20if%20voxel.map(%7Cvalue%7C%20range.contains(%26value)).unwrap_or(false)%20%7B%0A%20%20%20%20%20%20%20%20println!(%22Value%22)%3B%0A%20%20%20%20%7D%0A%7D%0A

janper:
I see. Do you prefer this? Doesn't seem to be more readable or perform better. Up to you, though.

janper:
Ready for another review

src/mesh/voxel_cloud.rs
```diff
@@ -34,6 +34,18 @@ use super::{primitive, tools, Face, Mesh};
 /// value outside of the block will cause the program to panic. Reading a value
 /// from beyond the bounds returns None, which is also a valid value even inside
 /// the block.
+///
+/// `volume_value_range` is an interval defining which values of the scalar
```
yanchith: Let's preface this comment with something like: "In this struct's method parameters, ...". This is what I meant by "You may have to adjust the comment to fit naturally in the new place." in my original review comment.

src/mesh/voxel_cloud.rs
```diff
@@ -34,6 +34,18 @@ use super::{primitive, tools, Face, Mesh};
...
+/// field should be considered to be a volume. The `ScalarField::from_mesh`
+/// generates a scalar field, which marks volume voxels with value `0`.
+/// `compute_distance_field` marks each voxel with a value representing the
+/// v
+/// oxel's distance from the original volume, therefore the voxels right at the
```
yanchith: Accidental whitespace

src/mesh/voxel_cloud.rs
yanchith: Hmmm, I am not sure it applies to intersection. The intersection bounding box may and up bigger than the actual volume, because voxels might be turned off after computing the intersection bounding box and resizing to it.

yanchith (commented):
Almost there! I am going to review the tests now.

src/mesh/voxel_cloud.rs
```diff
@@ -929,4 +1286,258 @@ mod tests {
...
+        sf_correct.fill_with(Some(0.0));
+        sf_correct.set_value_at_absolute_voxel_coordinate(&Point3::new(2, 2, 2), None);
+
+        sf_a.boolean_intersection(&(0.0..=0.0), &sf_b, &(0.0..=0.0));
+
+        assert_eq!(sf_a, sf_correct);
+    }
+
+    #[test]
+    fn test_scalar_field_boolean_union_shifted() {
```
yanchith, line 1377: What a beautiful shape :heart:

yanchith (commented):
Tests look good! I think we are just missing a test for boolean difference. A very simple one will do.

janper:
ready for final review

src/math.rs
```diff
@@ -36,10 +36,19 @@ pub fn lerp(source: f32, target: f32, weight: f32) -> f32 {
 pub fn remap<T, U>(value: T, source_range: &U, target_range: &U) -> Option<f64>
 where
     f64: std::convert::From<T>,
```
yanchith, line 38: Not saying you should change this, but note that this can also be written as `T: std::convert::Into<f64>`. In general, if there is a `From` implementation, the `Into` implementation gets generated automatically, because the standard library provides this blanket implementation: https://doc.rust-lang.org/stable/src/core/convert/mod.rs.html#553-560

src/mesh/voxel_cloud.rs
```diff
@@ -538,6 +538,36 @@ impl ScalarField {
     ) where
         U: RangeBounds<f32>,
     {
+        use std::ops::Bound::*;
+
+        // FIXME: This discards the `other` scalar field's fine-grain value
```
yanchith, line 543: T_T We'll have to think about this in a more general sense. Thanks for documenting it in the fixme.

yanchith (approved):
Thanks!

src/math.rs
janper: I tried, didn't work for some reason. Maybe I did something differently, but that was my first attempt. Also I kind of like it this way because it describes better what it actually does.

src/mesh/voxel_cloud.rs
janper: I believe this is a philosophical question - what is the validity of a scalar field when exposed to another, in general? Should we take into consideration only the volume value range or also the values around it? What about infinite (very common cases) volume value ranges?
This fix should be enough for the UI operations we have, though.

PR #157, opened 2020-03-16
Add project extension if it's missing while saving

ondrowan:
While Windows can filter by file type in save dialog, it doesn't automatically handle adding of this file type's extension to saved file. This patch fixes it and adds extension to files that don't contain it.

### DIFF
--- a/src/project.rs
+++ b/src/project.rs
@@ -3,7 +3,7 @@ use std::fmt;
 use std::fs::File;
 use std::io::prelude::*;
 use std::io::{self, BufReader};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use ron;
 use serde::Serialize;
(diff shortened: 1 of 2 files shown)


janper (dismissed):
yup, it works

src/project.rs
```diff
@@ -97,17 +97,28 @@ pub struct Project {
...
+pub fn save<P: AsRef<Path>>(path: P, project: Project) -> Result<PathBuf, ProjectError> {
     let pretty_config = ron::ser::PrettyConfig::default();
     let mut serializer = ron::ser::Serializer::new(Some(pretty_config), true);
     project.serialize(&mut serializer)?;
 
+    let mut path_buf = path.as_ref().to_path_buf();
+
+    if path_buf.extension().is_none() {
+        path_buf.set_extension(PROJECT_EXTENSION);
+    }
```
yanchith: Should we be assuming that if there is an extension, it is correct? The doc comment says "If this path does not contain valid project extension", but this code doesn't handle validity.

yanchith (commented):
Looks good but see inline comment.

src/project.rs
ondrowan: It's fixed now. If you try to save `gurban.turban` it actually saves `gurban.turban.hurban`. @janper please test this fix one more time as well.

yanchith (approved):

PR #159, opened 2020-03-19
Implement streaming obj serialization

yanchith:
This implements streaming OBJ serialization into a
`std::io::Writer`. Functionality is triggered by a main menu "Export OBJ"
button. Only unused geometries are exported. For each mesh or element of
mesh-array, a new "o" tag is created in the resulting OBJ file. The object name
is the same as the `return_value_name` that would be displayed in the pipeline
window in most cases. For some edge cases however, only the raw stringified var
ident is outputted. If the exported mesh is a mesh-array element and it's index
is not zero, it is part of the name written after the "o" tag.

The commit also does a bit of minor cleanup elsewhere:

- Ensures the session never generates the same variable identifier
  twice (related to being confident that if a var name is present in the
  session, it does not reference a previously removed variable declaration),
- Removes some per-frame uses of the format! macro from `ui.rs`,
- Inlines one simple call to tinyfiledialogs from `ui.rs` to `lib.rs`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -314,6 +314,7 @@ pub fn init_and_run(options: Options) -> ! {
                     &mut viewport_draw_mode,
                     &mut viewport_draw_used_values,
                     &mut project_status,
+                    &mut session,
                     &mut notifications.borrow_mut(),
                 );
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -601,7 +601,7 @@ pub fn init_and_run(options: Options) -> ! {
 
                 if menu_status.export_obj {
                     if let Some(path) = tinyfiledialogs::save_file_dialog_with_filter(
-                        "Save",
+                        "Export OBJ",
                         "export.obj",
                         &["*.obj"],
                         "Wavefront (.obj)",
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2,9 +2,12 @@ pub use crate::logger::LogLevel;
 pub use crate::renderer::{GpuBackend, GpuPowerPreference, Msaa};
 pub use crate::ui::Theme;
 
+use std::borrow::Cow;
 use std::cell::RefCell;
 use std::collections::HashMap;
+use std::f32;
 use std::fs::File;
+use std::io::BufWriter;
 use std::mem;
 use std::rc::Rc;
 use std::sync::Arc;
@@ -36,6 +39,7 @@ mod analytics;
 mod bounding_box;
 mod camera;
 mod convert;
+mod exporter;
 mod imgui_winit_support;
 mod input;
 mod interpreter;
@@ -341,9 +345,6 @@ pub fn init_and_run(options: Options) -> ! {
                     );
                 }
 
-                let reset_viewport =
-                    input_state.camera_reset_viewport || menu_status.reset_viewport;
-
                 if let Some(prevent_overwrite_modal_trigger) = menu_status.prevent_overwrite_modal {
                     project_status.prevent_overwrite_status = match prevent_overwrite_modal_trigger
                     {
@@ -540,7 +541,12 @@ pub fn init_and_run(options: Options) -> ! {
                         SaveModalResult::Save => {
                             let save_path = match project_status.path.clone() {
                                 Some(project_path) => Some(project_path),
-                                None => ui_frame.draw_save_dialog(),
+                                None => tinyfiledialogs::save_file_dialog_with_filter(
+                                    "Save",
+                                    project::DEFAULT_NEW_FILENAME,
+                                    project::EXTENSION_FILTER,
+                                    project::EXTENSION_DESCRIPTION,
+                                ),
                             };
 
                             if let Some(save_path) = save_path {
@@ -548,7 +554,7 @@ pub fn init_and_run(options: Options) -> ! {
                                 let project = project::Project { version: 1, stmts };
 
                                 match project::save(&save_path, project) {
-                                    Ok(_) => match prevent_overwrite_status {
+                                    Ok(()) => match prevent_overwrite_status {
                                         project::NextAction::Exit => {
                                             *control_flow = winit::event_loop::ControlFlow::Exit
                                         }
@@ -562,7 +568,7 @@ pub fn init_and_run(options: Options) -> ! {
                                         }
                                     },
                                     Err(err) => {
-                                        log::error!("{}", err);
+                                        log::error!("Project save failed: {}", err);
 
                                         project_status.error = Some(err);
                                     }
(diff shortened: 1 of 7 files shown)


yanchith:
While dreaming, I realized a bug I introduced here. Gonna fix it now. Stay tuned.

yanchith:
@ondrowan @janper I'd appreciate a review of commit https://github.com/subdgtl/HURBAN-Selector/pull/159/commits/668c15d97e91db904b8db791fa9fd9107abc7bdb at least. I tried to keep the fix as minimal as possible, delegating further work for FIXMEs. Comments in the commit explain the problem and the "solution".

janper:
> @ondrowan @janper I'd appreciate a review of commit [668c15d](https://github.com/subdgtl/HURBAN-Selector/commit/668c15d97e91db904b8db791fa9fd9107abc7bdb) at least. I tried to keep the fix as minimal as possible, delegating further work for FIXMEs. Comments in the commit explain the problem and the "solution".

I understand the issue, I understand the fix, I don't understand the code though.
Nevertheless, are you really worried we'd run out of  9,223,372,036,854,775,807 idents in one session? And if so, can't you use u128, which gives you 340,282,366,920,938,463,463,374,607,431,768,211,455 idents?

src/ui.rs
```diff
@@ -924,16 +919,26 @@ impl<'a> UiFrame<'a> {
...
                         \n\
-                        Opens the dialog for saving the current viewport into a PNG file.");
+                        Opens a system dialog for saving the current viewport into a PNG file.");
                         wrap_token.pop(ui);
                     });
                 }
 
-                ui.separator();
+                status.export_obj = ui.button(
+                    imgui::im_str!("Export OBJ"),
```
janper: there is a convention, that labels of buttons which open a new dialog end with a triple dot, so
`Export OBJ...`

src/lib.rs
```diff
@@ -584,15 +590,86 @@ pub fn init_and_run(options: Options) -> ! {
...
-                if reset_viewport {
+                if input_state.camera_reset_viewport || menu_status.reset_viewport {
                     camera_interpolation =
                         Some(CameraInterpolation::new(&camera, &scene_bounding_box, time));
                 }
 
+                if menu_status.export_obj {
+                    if let Some(path) = tinyfiledialogs::save_file_dialog_with_filter(
+                        "Save",
+                        "export.obj",
```
janper: This is a good filename when the current project hasn't been saved yet. The export usually suggests the same filename as the current file, just a different extension, so of the project is called `my_model.hurban`, the the suggested export file name should be `my_model.obj`.

src/lib.rs
```diff
@@ -584,15 +590,86 @@ pub fn init_and_run(options: Options) -> ! {
...
 
-                if reset_viewport {
+                if input_state.camera_reset_viewport || menu_status.reset_viewport {
                     camera_interpolation =
                         Some(CameraInterpolation::new(&camera, &scene_bounding_box, time));
                 }
 
+                if menu_status.export_obj {
+                    if let Some(path) = tinyfiledialogs::save_file_dialog_with_filter(
+                        "Save",
```
janper: Please, change to `Export OBJ`

janper (changes requested):
Thanks, It works, it's nice and the exported files are valid. I added some comments inline and some comments here:

- Could you please add a notification, just like in the case of the Screenshot saver?
- Should it be possible to export an empty OBJ? I prefer disabling the button but also displaying a notification that the file hasn't been saved because it would be empty, would be OK.
- I think you already mentioned it in the code comments, but there is quite a nasty behavior that the exporter exports currently computed geometry, not the current pipeline. So if you disable autoexecute then you may very easily export outdated version of the geometry. I suggest either informing about this in the notification area or even better trigger pipeline execution (if needed?) prior to the export.
- EDIT: ~I'd like to change the menu button order. I've created a Trello task: https://trello.com/c/Z7Rp0yLS/325-reorder-menu-buttons~

src/ui.rs
yanchith: Another task. Another PR.

yanchith:
> Nevertheless, are you really worried we'd run out of 9,223,372,036,854,775,807 idents in one session? 

Not in one session. They don't reset. Ever. Meaning if you send a save file to someone and they load it, they don't start from 0. Regardless, I am not worried u64 won't be enough, even then. This is just an edge case that can be handled easily, so I think we should at some point. Hence the FIXME.

> And if so, can't you use u128

Same principle. Practically we won't ever run out, but it would be nice if we were handling this correctly.

yanchith:
> Could you please add a notification, just like in the case of the Screenshot save?

There is one. You probably don't see it because of the system dialog freezing our app though.

yanchith:
> Should it be possible to export an empty OBJ? I prefer disabling the button but also displaying a notification that the file hasn't been saved because it would be empty, would be OK.

Ok, I can disable it.

src/ui.rs
janper: I only want you to add three dots to this string. What is the issue?

yanchith:
> I think you already mentioned it in the code comments, but there is quite a nasty behavior that the exporter exports currently computed geometry, not the current pipeline. So if you disable autoexecute then you may very easily export outdated version of the geometry. I suggest either informing about this in the notification area or even better trigger pipeline execution (if needed?) prior to the export.

It is not only about autoexecute. The same happens with autoexecute and long running operation. Actually, this is kind of the expected behavior, but I agree it is not intuitive :smile: Where do you want to notify the user of this? I don't think the notification area is the right place, because the system dialog vs notifications interaction I mentioned above.

I also don't think triggering an execution without at least a prompt is a good idea. Also, it won't solve the situation with a long running operation.

I'd perhaps just mention it in the button's description.

UPDATE: Discussed offline. The export button will be disabled (and have a tooltip explaining why), if the interpreter is either busy, or hasn't been sent the latest program to execute.

src/ui.rs
yanchith: Nevermind, I'll do it in here, it is a small thing. I'll also do https://trello.com/c/Z7Rp0yLS/325-reorder-menu-buttons

yanchith:
@janper I believe I addressed all of your points. Each point is addressed by a separate commit, should you want to review in detail. Let me know if you have more requests.

janper:
I've tested all scenarios I could think of and it seems to work great! Thank you.

There is still one issue though - the export button is clicable and functional even when disabled.

janper:
I see you even made the warning appear when relevant. Nice!

yanchith:
> There is still one issue though - the export button is clicable and functional even when disabled.

Ooops. One sec :smile:

janper (approved):

PR #160, opened 2020-03-20
Add shortcuts for running the program and deleting last statement

yanchith:


janper (commented):
I'm not sure what this is  ¯\_(ツ)_/¯

janper (dismissed):
I SEE!!! The name of this PR is just so unclear...

Great, it works really well! 

We should inform the user about this as well. What about this for now?
image

yanchith:
> We should inform the user about this as well. What about this for now?

Sure. I'll add this

yanchith:
> The name of this PR is just so unclear...

We have different internal and external terminology. I tend to use the internal one.

PR #161, opened 2020-03-20
Add voxel-based interpolated union (voxel morph) and supplemental code

janper:


src/mesh/voxel_cloud.rs
```diff
@@ -685,6 +705,100 @@ impl ScalarField {
...
+    /// scalar field should be considered to be a volume. The
+    /// `ScalarField::from_mesh` generates a scalar field, which marks volume
+    /// voxels with value `0`. `compute_distance_field` marks each voxel with a
+    /// value representing the voxel's distance from the original volume,
+    /// therefore the voxels right at the shell of the volume are marked 0, the
+    /// layer around them is marked 1 or -1 (inside closed volumes) etc. Once
+    /// the scalar field is populated with meaningful values, it is possible to
+    /// treat (perform boolean operations or materialize into mesh) on various
+    /// numerical ranges. Such range is specified ad-hoc by parameter
+    /// `volume_value_range`.
```
yanchith: I thought we agreed to move this comment elsewhere in previous PR

src/mesh/voxel_cloud.rs
```diff
@@ -685,6 +705,100 @@ impl ScalarField {
...
+    ///
+    /// # Panics
+    /// Panics if one of the volume value ranges is infinite.
```
yanchith: Newline

src/mesh/voxel_cloud.rs
```diff
@@ -685,6 +705,100 @@ impl ScalarField {
...
+    ///
+    /// # Warning
```
yanchith: Newline

src/interpreter_funcs/voxel_interpolated_union.rs
```diff
@@ -0,0 +1,260 @@
...
+pub struct FuncInterpolatedUnion;
+
+impl Func for FuncInterpolatedUnion {
+    fn info(&self) -> &FuncInfo {
+        &FuncInfo {
+            name: "Voxel Morph",
+            description: "VOXEL INTERPOLATION MORPH FROM TWO MESH GEOMETRIES\n\
+            \n\
+            Converts the input mesh geometries into voxel clouds, resizes both voxel \
+            clouds to bel large enough to contain volumes from both of them \
```
yanchith: "to be large enough"

src/interpreter_funcs/voxel_interpolated_union.rs
```diff
@@ -0,0 +1,260 @@
...
+                \n\
+                High values produce coarser results, low values may increase precision but produce \
+                heavier geometry that significantly affect performance. Too high values produce \
+                single large voxel, too low values may generate holes in the resulting geometry.",
+                refinement: ParamRefinement::Float3(Float3ParamRefinement {
+                    min_value: Some(0.005),
+                    max_value: None,
+                    default_value_x: Some(0.25),
+                    default_value_y: Some(0.25),
+                    default_value_z: Some(0.25),
```
yanchith: These defaults are different from other voxel based operations. Is this on purpose? Should we try to adjust the default here or for other operations?

src/interpreter_funcs/voxel_interpolated_union.rs
```diff
@@ -0,0 +1,260 @@
...
+            },
+            ParamInfo {
+                name: "Factor",
+                description: "Proportional interpolation factor between the first and the second mesh.\n\
+                \n\
+                Factor = 0.0: the result is equal to the first mesh.\n\
+                Factor = 0.5: the result is half way between the first and the second mesh.\n\
+                Factor = 1.0: the result is equal to the second mesh.",
+                refinement: ParamRefinement::Float(FloatParamRefinement {
+                    default_value: Some(0.0),
```
yanchith: Don't we want the default to be 0.5?

src/interpreter_funcs/voxel_interpolated_union.rs
```diff
@@ -0,0 +1,260 @@
...
+        let voxel_dimensions = Vector3::from(args[2].unwrap_float3());
+        let fill = args[3].unwrap_boolean();
+        let factor = args[4].unwrap_float();
+        let error_if_large = args[5].unwrap_boolean();
+        let analyze_bbox = args[6].unwrap_boolean();
+        let analyze_mesh = args[7].unwrap_boolean();
+
+        if voxel_dimensions
+            .iter()
+            .any(|dimension| approx::relative_eq!(*dimension, 0.0))
```
yanchith: Let's check for negative values here.

Dejavu again. Would be good if you checked all your future PRs for this, so we don't have to do this every time.

src/mesh/voxel_cloud.rs
```diff
@@ -448,6 +448,26 @@ impl ScalarField {
         tools::weld(&joined_voxel_mesh, (min_voxel_dimension as f32) / 4.0)
     }
 
+    /// Compute bounding box from scalar field. The bounding box will contain
+    /// the entire block of the voxel space described by the current scalar
+    /// field, regardless if it contains any voxels.
+    ///
+    /// The Bounding box will be defined in cartesian units.
+    pub fn bounding_box_cartesian(&self) -> BoundingBox<f32> {
```
yanchith: The other function returning a bounding box is called. `volume_voxel_space_bounding_box`. Either rename this to `cartesian_bounding_box` or the other to `bounding_box_volume_voxel_space`.

I usually slightly prefer the C-style naming of things (`object_verb` or `module_object_verb`), but decide what you think is more pleasing to the eye for you :slightly_smiling_face:

yanchith: Also. Do you think we should be using `cartesian_space` instead of `cartesian` to make the duality with `voxel_space` more obvious? I am undecided.

src/mesh/voxel_cloud.rs
```diff
@@ -685,6 +705,100 @@ impl ScalarField {
         }
         self.shrink_to_fit(volume_value_range_self)
     }
+    /// Computes interpolated values of two scalar fields. Prior to
+    /// interpolation, the scalar fields will be recomputed to distance fields.
```
yanchith: I suggest we don't do it this way. The current approach leaves little control for the caller to optimize and does unnecessary work.

Instead I suggest:

- Make the user compute the distance fields manually before passing the scalar fields in,
- Document that this function assumes they have been computed and that otherwise it is garbage-in, garbage-out
- Don't clone the other scalar field

I will review the body of the function again once this has been changed.

src/mesh/voxel_cloud.rs
```diff
@@ -685,6 +705,100 @@ impl ScalarField {
...
+    /// The current scalar field will be mutated and resized to encompass values
+    /// from both scalar fields. The two scalar fields do not have to contain
+    /// voxels of the same size.  The interpolation is linear.
+    ///
+    /// `interpolation_factor` is a value between 0.0 and 1.0 defining the ratio
+    /// between the current value to the other value.
```
yanchith: Should we document what happens if the factor exceeds the 0..1 range?

src/mesh/voxel_cloud.rs
```diff
@@ -685,6 +705,100 @@ impl ScalarField {
...
+            // Wipe the current scalar field if none of the scalar fields
+            // contained any volume voxels.
+            self.wipe();
+        }
+    }
+
+    /// Interpolate values of the current scalar field to the values of the
+    /// other scalar field. The current scalar field will not be resized, so
+    /// only the overlapping areas will be will be interpolated. If the voxel in
+    /// the current or the other scalar field is Non (does not contain any voxel
```
yanchith: Non -> None

src/mesh/voxel_cloud.rs
```diff
@@ -685,6 +705,100 @@ impl ScalarField {
...
+    /// or does not exist), the output value will be also None. The two scalar
+    /// fields do not have to contain voxels of the same size. The interpolation
+    /// is linear.
+    ///
+    /// `interpolation_factor` is a value between 0.0 and 1.0 defining the ratio
+    /// between the current value to the other value.
```
yanchith: Same as above, let's document the border conditions.

src/mesh/voxel_cloud.rs
```diff
@@ -685,6 +705,100 @@ impl ScalarField {
...
+            other_clone.resize_to_cartesian_bounding_box(&bounding_box);
+            self.compute_distance_field(volume_value_range_self);
+            other_clone.compute_distance_field(volume_value_range_other);
+
+            self.interpolate_to(&other_clone, interpolation_factor);
+        } else {
+            // Wipe the current scalar field if none of the scalar fields
+            // contained any volume voxels.
+            self.wipe();
+        }
```
yanchith: This is a bounding box union, meaning it should never happen if you provide two bounding boxes. Instead I'd `expect` here.

yanchith (changes requested):
Nice work! I only have optimization, documentation and readability comments, but no semantic objections.

src/interpreter_funcs/voxel_interpolated_union.rs
janper: we need to decide. if the user plays with the default primitives, which are always around 1 unit big, it may make sense to set the default voxel size to somewhere around 0.1 (I know it's not 0.25).
on the other hand, when the primitives are by default set to the unit size (1), then also the voxels could be set to 1. in real life though, many users will get frustrated when playing with the software and learning, that they get visually unpleasing results with the default values.
on the third hand, most of the imported models are made in mm (products) or cm (german  world) or m (architecture and urban planning) and the objects are usually several thousands of units big. in such a case, even voxel size 1 is too small to perform well. 

any suggestion?

src/mesh/voxel_cloud.rs
janper: Are you sure about this? There is another function called `interpolate_to` doing exactly this. The one you refer to is what can be extracted to the UI operation (unless there is some private field use).

src/mesh/voxel_cloud.rs
janper: it will be a linear extrapolation, as expected. i believe there is no need to explicitly document that. i can, though

src/mesh/voxel_cloud.rs
janper: it can! the values are being interpolated, so you can't know if the resulting number fall into the meshing range.

src/interpreter_funcs/voxel_interpolated_union.rs
yanchith: Tough question... Maybe let's optimize for our current imported models (meaning real units that come in the hundreds and thousands) and keep this at 1 for now? We will have to increase the safe voxel limit for this so that this doesn't always error.

src/mesh/voxel_cloud.rs
```diff
@@ -685,6 +705,90 @@ impl ScalarField {
...
+        other: &mut ScalarField,
+        volume_value_range_other: &U,
+        interpolation_factor: f32,
+    ) where
+        U: RangeBounds<f32>,
+    {
+        let bounding_box_self = self.bounding_box_cartesian_space();
+        let bounding_box_other = other.bounding_box_cartesian_space();
+
+        if let Some(bounding_box) = BoundingBox::union(vec![bounding_box_self, bounding_box_other])
```
yanchith: please don't allocate a vector here

src/mesh/voxel_cloud.rs
yanchith: True, but the condition is on the result of `BoundingBox::union` of two cartesian bounding boxes. IIRC those boxes are just computed from the dimensions of the scalar field without looking at voxel values. Maybe that wasn't supposed to be like that?

src/mesh/voxel_cloud.rs
yanchith: Hmm, then maybe let's extract it to the operation, in which case we don't need this function at all?

src/interpreter_funcs/voxel_interpolated_union.rs
janper: regarding the voxel limit - can I put it somewhere central/global?

janper: also, what about adjusting the primitives to roughly match the scale of our sample OBJs?

like setting them to 10, 100 or 1000 even by default

src/mesh/voxel_cloud.rs
janper: you're right. this can't be an empty bounding box, unless both of the input bounding boxes are empty. this is rather an UI operation error

src/interpreter_funcs/voxel_interpolated_union.rs
yanchith: > like setting them to 10, 100 or 1000 even by default

Yeah, let's do that (in another PR)

src/mesh/voxel_cloud.rs
```diff
@@ -88,7 +88,7 @@ impl ScalarField {
     /// # Panics
     ///
     /// Panics if any of the voxel dimensions is below or equal to zero.
-    pub fn from_cartesian_bounding_box(
+    pub fn from_bounding_box_cartesian_space_space(
```
yanchith: space_space :slightly_smiling_face:

src/interpreter_funcs/voxel_interpolated_union.rs
```diff
@@ -217,30 +214,40 @@ impl Func for FuncInterpolatedUnion {
...
+            voxel_cloud1.resize_to_bounding_box_cartesian_space(&bounding_box);
+            voxel_cloud1.compute_distance_field(&volume_value_range);
+            voxel_cloud2.compute_distance_field(&volume_value_range);
+
+            voxel_cloud1.interpolate_to(&voxel_cloud2, interpolation_factor);
+        } else {
+            let error = FuncError::new(FuncInterpolatedUnionError::EmptyScalarField);
+            log(LogMessage::error(format!("Error: {}", error)));
+            return Err(error);
+        }
```
yanchith: Similar to my previous comment. The union of **2** bounding boxes can not be `None` (`None` is not the same as empty). This branch will never happen.

src/mesh/voxel_cloud.rs
yanchith: Yeah, but empty bounding box is still not `None`. This is not even a UI operation error.

yanchith (commented):
Almost there

src/interpreter_funcs/voxel_interpolated_union.rs
janper: So, should `BoundingBox::union` even return an `Option`? `None` can't happen at all.

yanchith: It can. If you give it an empty iterator.

janper: true

yanchith (approved):

PR #162, opened 2020-03-20
Add marching cubes meshing routine to existing UI operations

janper:


src/mesh/voxel_cloud.rs
```diff
@@ -1022,6 +1181,294 @@ impl ScalarField {
     }
 }
 
+/// Lookup table of triangle faces suitable to the sample situation. Each vector
+/// item contains a vector of vertex indices. The vector indices correspond with
+/// the edge index.
+///
+///  ## Corner and edge indices scheme
```
yanchith: Any reason why this is a level 2 title?

src/mesh/voxel_cloud.rs
```diff
@@ -448,6 +448,165 @@ impl ScalarField {
...
+                        Point3::new(x, y, z + 1),
+                    ];
+                    let corners_cartesian_coordinates: Vec<_> = corners_absolute_coordinates
+                        .iter()
+                        .map(|a| absolute_voxel_to_cartesian_coordinate(a, &self.voxel_dimensions))
+                        .collect();
+
+                    // These points will become vertices of mesh patches forming
+                    // the outer envelope of the scalar field.
+                    #[rustfmt::skip]
```
yanchith: I have to say the autoformatted version of this looks much nicer than what I see here. Any reason for this skip?

src/mesh/voxel_cloud.rs
```diff
@@ -1022,6 +1181,294 @@ impl ScalarField {
...
+///              |  /                   |  /
+///              | /e3                  | /e1
+///              |/_____________________|/
+///              v3         e2          v2
+/// ```
+///
+/// # Source
+///
+/// http://paulbourke.net/geometry/polygonise/
+fn marching_cubes_lookup_table() -> Vec<Vec<(u32, u32, u32)>> {
```
yanchith: This is very wasteful. We can do much better.

We'd like to allocate this data in a contiguous block of memory. Also, there is no reason to create this data in runtime. We can store the table in the static memory. For inspiration:

```rust
/// Row in the marching cubes lookup table.
///
/// Fields are u8 so that more rows fit in an x86 cache line (64 bytes).
/// With the current layout optimization, a row is 16 bytes, meaning 4 of them fit.
struct Row {
    /// Lenght of valid data slice.
    len: u8,
    /// Row data.
    data: [(u8, u8, u8); 5],
}

// Use https://crates.io/crates/static_assertions to ensure the size
static_assertions::assert_eq_size!(Row, [16; u8]);


impl Row {
    fn as_slice(&self) -> &[(u8, u8, u8)] {
        &self.data[0..usize::from(self.len)]
    }
}

const EMPTY: (u8, u8, u8) = (0, 0, 0);

/// Marching cubes lookup table. The size is just for demonstration - real table is bigger.
static MARCHING_CUBES_LOOKUP_TABLE: [Row; 3] = [
    Row { len: 0, data: [EMPTY, EMPTY, EMPTY, EMPTY, EMPTY] },
    Row { len: 1, data: [(0, 1, 2), EMPTY, EMPTY, EMPTY, EMPTY] },
    Row { len: 2, data: [(0, 1, 2), (3, 4, 5), EMPTY, EMPTY, EMPTY] },
];

fn main() {
    let row_data = MARCHING_CUBES_LOOKUP_TABLE[2].as_slice();
    println!("{:?}", row_data);
    println!("{}", std::mem::size_of::<Row>());
    println!("{}", std::mem::size_of::<[Row; 2]>());
}
```

src/mesh/voxel_cloud.rs
```diff
@@ -448,6 +448,165 @@ impl ScalarField {
...
+                        Point3::new(x, y + 1, z),
+                        Point3::new(x + 1, y + 1, z),
+                        Point3::new(x + 1, y, z),
+                        Point3::new(x, y, z),
+                        Point3::new(x, y + 1, z + 1),
+                        Point3::new(x + 1, y + 1, z + 1),
+                        Point3::new(x + 1, y, z + 1),
+                        Point3::new(x, y, z + 1),
+                    ];
+                    let corners_cartesian_coordinates: Vec<_> = corners_absolute_coordinates
```
yanchith: No need to heap-allocate. Either collect to a `ArrayVec<[Point<f32>; 8]>`, or perhaps just explicitly write out the array.

I prefer the latter, because the only reason we currently use the arrayvec crate is by-value `.into_iter()` for arrays (intead of by-ref), but rust is considering to fix this with a breaking change (https://github.com/rust-lang/rust/pull/65819), meaning we could potentially drop arrayvec as a dependency later.

src/mesh/voxel_cloud.rs
```diff
@@ -448,6 +448,165 @@ impl ScalarField {
...
+                        .iter()
+                        .map(|a| absolute_voxel_to_cartesian_coordinate(a, &self.voxel_dimensions))
+                        .collect();
+
+                    // These points will become vertices of mesh patches forming
+                    // the outer envelope of the scalar field.
+                    #[rustfmt::skip]
+                    let edge_midpoints = vec![
```
yanchith: Use an array, no need to heap-allocate

src/mesh/voxel_cloud.rs
```diff
@@ -448,6 +448,165 @@ impl ScalarField {
...
+                        Point3::from(corners_cartesian_coordinates[7].coords.lerp(&corners_cartesian_coordinates[4].coords, 0.5)), //e7
+                        Point3::from(corners_cartesian_coordinates[0].coords.lerp(&corners_cartesian_coordinates[4].coords, 0.5)), //e8
+                        Point3::from(corners_cartesian_coordinates[1].coords.lerp(&corners_cartesian_coordinates[5].coords, 0.5)), //e9
+                        Point3::from(corners_cartesian_coordinates[2].coords.lerp(&corners_cartesian_coordinates[6].coords, 0.5)), //e10
+                        Point3::from(corners_cartesian_coordinates[3].coords.lerp(&corners_cartesian_coordinates[7].coords, 0.5)), //e11
+                    ];
+
+                    // Boolean vector marking each of the voxels either volume
+                    // (`true`) or void(`false`)
+                    let corners_inside_volume_pattern: Vec<_> = corners_absolute_coordinates
```
yanchith: Use an an array. No need to heap-allocate.

src/mesh/voxel_cloud.rs
```diff
@@ -448,6 +448,165 @@ impl ScalarField {
...
+                    }
+                    if corners_inside_volume_pattern[7] {
+                        sample_index |= 128;
+                    }
+
+                    // The marching cubes lookup table sample, that contains
+                    // triplets of indices to the edge midpoints (currently
+                    // already vertices) that will become mesh patch faces.
+                    let marching_cubes_faces =
+                        marching_cubes_lookup_table[cast_usize(sample_index)].clone();
```
yanchith: `usize::from(sample_index)`

Let's only use our `cast_*` functions for conversions that can fail so that we can pinpoint issues better.

yanchith: The `.clone()` most likely won't be necessary once you incorporate my lookup table suggestion, but a slice would be enough in this case also - no need to have the whole vector if we just need a view.

src/mesh/voxel_cloud.rs
```diff
@@ -448,6 +448,165 @@ impl ScalarField {
...
+            || !self.contains_voxels_within_range(volume_value_range)
+        {
+            return None;
+        }
+
+        let marching_cubes_lookup_table = marching_cubes_lookup_table();
+
+        // A collection of mesh patches defining an outer envelope of volumes
+        // stored in the scalar field
+        let mut marching_cubes: Vec<Mesh> = Vec::new();
```
yanchith, line 473: How likely is it that we get an empty result from the marching cubes table? Don't we want to pre-allocate with capacity here?

src/mesh/voxel_cloud.rs
```diff
@@ -448,6 +448,165 @@ impl ScalarField {
...
+                        marching_cubes_lookup_table[cast_usize(sample_index)].clone();
+
+                    // In some cases (when all the voxels are inside or outside
+                    // a volume), there is no mesh patch to be created.
+                    if !marching_cubes_faces.is_empty() {
+                        marching_cubes.push(
+                        // Generate new mesh path from all edge midpoint
+                        // vertices and only those faces that should be created
+                        // according th the marching cubes lookup table.
+                        Mesh::from_triangle_faces_with_vertices_and_computed_normals_remove_orphans(
```
yanchith, line 671: `remove_orphan_vertices` is kind of wasteful in this situation (called by this mesh constructor). A by hand implementation could be way more efficient. Let's leave an optimization fixme here.

src/mesh/voxel_cloud.rs
```diff
@@ -448,6 +448,165 @@ impl ScalarField {
...
+        }
+
+        // Join separate mesh planes into one mesh
+        let joined_voxel_mesh = tools::join_multiple_meshes(&marching_cubes);
+        let min_voxel_dimension = self
+            .voxel_dimensions
+            .x
+            .min(self.voxel_dimensions.y.min(self.voxel_dimensions.z));
+        // and weld naked edges.
+        tools::weld(&joined_voxel_mesh, (min_voxel_dimension as f32) / 4.0)
```
yanchith, line 696: Hmmm. I believe this should be possible to do without weld. Let's leave a fixme here to construct the mesh in-place.

yanchith (changes requested):
Beautiful :heart: 

I do have a couple of optimization related suggestions though.

src/mesh/voxel_cloud.rs
janper: because it is a less important information. I'll change it to level 1 though

src/mesh/voxel_cloud.rs
janper: because the autoformatted version looks much worse than this

yanchith: I tried formatting it locally and it looked ok. Are you sure?

src/mesh/voxel_cloud.rs
janper:
```
error[E0271]: type mismatch resolving `<std::slice::Iter<'_, nalgebra::geometry::point::Point<f32, nalgebra::base::dimension::U3>> as std::iter::IntoIterator>::Item == nalgebra::geometry::point::Point<f32, nalgebra::base::dimension::U3>`
   --> src\mesh\voxel_cloud.rs:635:25
    |
635 |                         Mesh::from_triangle_faces_with_vertices_and_computed_normals_remove_orphans(
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected reference, found struct `nalgebra::geometry::point::Point`
    | 
   ::: src\mesh\mod.rs:192:12
    |
192 |     pub fn from_triangle_faces_with_vertices_and_computed_normals_remove_orphans<F, V>(
    |            ---------------------------------------------------------------------
...
199 |         V: IntoIterator<Item = Point3<f32>>,
    |                         ------------------ required by this bound in `mesh::Mesh::from_triangle_faces_with_vertices_and_computed_normals_remove_orphans`
    |
    = note: expected type `&nalgebra::geometry::point::Point<f32, nalgebra::base::dimension::U3>`
               found type `nalgebra::geometry::point::Point<f32, 
nalgebra::base::dimension::U3>`
```

yanchith: Can't help you without more context, but it looks like you are just missing a `.copied()` on an iterator

janper: `edge_midpoints.iter().map(|v| *v)` ??

src/mesh/voxel_cloud.rs
janper: ArrayVec, since it's a `bool` structure

src/mesh/voxel_cloud.rs
yanchith: If add the `.copied()` adapter, it should work.

src/mesh/voxel_cloud.rs
janper: I should never be empty, but we also have no clue about the size.

src/mesh/voxel_cloud.rs
janper: This is the optimization I mentioned yesterday. Cleaning up the list of vertices before feeding them into a mesh constructor, is exactly what `remove_orphan_vertices` does. So we either rely on an existing routine or we somehow put a very similar piece of code here. I decided not to do it here.

There is one more option - removing all the orphans all at once when all the mesh patches are joined into a single mesh. This, however, sounds way slower as the orphan removal will have to scan through a much longer list of items.

Am I missing something?

src/mesh/voxel_cloud.rs
janper: Again something I already gave a thought. Yes, it's possible, however you'd have to generate all the possible vertices and then remove the unused. Or you can create some sort of a "hash code" based on the vertex position, collect a list of such "hasehs", then calculate their cartesian position, then put them into some order and reindex the faces. Yes, that would be possible but also too much trouble, wouldn't it?

src/mesh/voxel_cloud.rs
yanchith: I don't follow your argument.

src/mesh/voxel_cloud.rs
yanchith: By empty I meant finding an empty row in the lookup table (there are 2), which means we would skip adding it to the mesh patch to this vec.

src/mesh/voxel_cloud.rs
yanchith: > So we either rely on an existing routine or we somehow put a very similar piece of code here. I decided not to do it here.

`remove_orphan_vertices` is general purpose and allocates. We could pre-filter without allocation with the information we have in this function. Note that I only suggested to add a fixme here, not actually do the change right now. I know it is a lot of work.

src/mesh/voxel_cloud.rs
yanchith: > Yes, that would be possible but also too much trouble, wouldn't it?

That's why I only suggested to add the fixme. It is to document potential future improvements to implementations.

src/mesh/voxel_cloud.rs
janper: we may have different aesthetic standards :)

src/mesh/voxel_cloud.rs
janper: had to use `ArrayVec` because of this:

```
error[E0277]: a collection of type `[bool; 8]` cannot be built from an iterator over elements of type `bool`
    --> src\mesh\voxel_cloud.rs:1671:30
     |
1671 | ...                   .collect();
     |                        ^^^^^^^ a collection of type `[bool; 8]` cannot be built from `std::iter::Iterator<Item=bool>`      
     |
     = help: the trait `std::iter::FromIterator<bool>` is not implemented for `[bool; 8]`
```

src/mesh/voxel_cloud.rs
janper: it's very probable - actually it's way more probable than not. only few voxels actually materialize into a mesh, most of them are outside or inside the volumetric shapes.

src/mesh/voxel_cloud.rs
janper: is this what you meant, though?

src/mesh/voxel_cloud.rs
```diff
@@ -12,6 +14,1082 @@ use crate::plane::Plane;
...
+/// With the current layout optimization, a row is 16 bytes, meaning 4 of them fit.
+struct Row {
+    /// Length of a valid data slice.
+    len: u8,
+    /// Row data.
+    data: [(u32, u32, u32); 5],
+}
+
+// FIXME: explain why this is 512 bits rather than expected 488 bits
+static_assertions::assert_eq_size!(Row, [u8; 64]);
```
yanchith: For this struct to be 16 bytes, you need the data to be `u8` also (exactly as the doc comment says), and later do `u32::from` to convert to a wider value when using. Using `u32` in the table itself defeats some the purpose of this optimization - the way it is currently, one row occupies the whole cache line. Not bad, but not good either.

> // FIXME: explain why this is 512 bits rather than expected 488 bits

These are usually expressed in bytes, not bits. And the reason is alignment. Read up:

- https://doc.rust-lang.org/nomicon/repr-rust.html
- https://doc.rust-lang.org/nomicon/other-reprs.html

src/mesh/voxel_cloud.rs
yanchith: Okay - in this case let's leave it `Vec::new`

src/mesh/voxel_cloud.rs
```diff
@@ -12,6 +14,1082 @@ use crate::plane::Plane;
 
 use super::{primitive, tools, Face, Mesh, NormalStrategy};
 
+/// Row in the marching cubes lookup table.
```
yanchith: Can we please move all the marching cubes implementation details below the public items so they don't clutter the start of the file?

src/mesh/voxel_cloud.rs
```diff
@@ -41,7 +1119,7 @@ use super::{primitive, tools, Face, Mesh, NormalStrategy};
 /// volume voxels with value `0`. `compute_distance_field` marks each voxel with
 /// a value representing the voxel's distance from the original volume,
 /// therefore the voxels right at the shell of the volume are marked 0, the
-/// layer around them is marked 1 or -1 (inside closed volumes) etc. Once the
+/// layer around them is marked 1 or x (inside closed volumes) etc. Once the
```
yanchith: accident?

yanchith:
https://github.com/subdgtl/HURBAN-Selector/pull/162#discussion_r396113690

Of course you did. You cannot `collect` to an array. You would have to write out the array manually - which is what I meant.

yanchith (commented):
Looking better now. I also realized why you couldn't do it without `ArrayVec` - because you were reluctant to give up iterators. Yes - to use arrays, you would have to manually specify their literals and fill each value. I would still prefer t drop arrayvec eventually, but we can do that later...

src/mesh/voxel_cloud.rs
yanchith: See my comment down below (github didn't allow me to reply here)

janper:
> You would have to write out the array manually - which is what I meant.

So that i know how did it feel to write a code for an ENIAC?

src/mesh/voxel_cloud.rs
janper: I intentionally used `u32` to avoid casting for each voxel. Is casting (or generally your version with `u8`) more efficient in this case?

src/mesh/voxel_cloud.rs
janper: possibly yes, i'll have to try

src/mesh/voxel_cloud.rs
janper: kind of

yanchith:
> So that i know how did it feel to write a code for an ENIAC?

:smile: 

Yes, and also so that we can remove dependencies and compile faster.

src/mesh/voxel_cloud.rs
yanchith: If nothing anything else, the table (and therefore our binary) is smaller with u8. I measured locally with both u32 and u8 and it does not seem to be much faster with u8 - most likely, because we don't access the table sequentially.

It for sure won't be slower, if that is what you ask.

src/mesh/voxel_cloud.rs
```diff
@@ -1184,6 +1451,1082 @@ fn absolute_voxel_to_one_dimensional_coordinate(
...
+/// Fields are u8 so that more rows fit in an x86 cache line (64 bytes).
+/// With the current layout optimization, a row is 16 bytes, meaning 4 of them fit.
+struct Row {
+    /// Length of a valid data slice.
+    len: u8,
+    /// Row data.
+    data: [(u8, u8, u8); 5],
+}
+
+// FIXME: explain why this is 512 bits rather than expected 488 bits
```
yanchith: This fixme is obsolete

src/mesh/voxel_cloud.rs
```diff
@@ -448,6 +450,250 @@ impl ScalarField {
...
+                        marching_cubes.push(
+                        // Generate new mesh path from all edge midpoint
+                        // vertices and only those faces that should be created
+                        // according to the marching cubes lookup table.
+                        //
+                        // FIXME: @Optimization remove unused vertices already
+                        // here and generate the mesh without the need of
+                        // removing orphans.
+                        Mesh::from_triangle_faces_with_vertices_and_computed_normals_remove_orphans(
+                            marching_cubes_faces.iter().map(|(a, b, c)| (u32::from(*a), u32::from(*b), u32::from(*c))),
```
yanchith: If you format this manually, does rustfmt return it to this style? It has a tendency to break for large files or large level of nesting.

yanchith (commented):
Last two things and we are done here. We don't have to remove the last use of arrayvec now. Also, please resolve conflicts before the last review.

yanchith (approved):

PR #163, opened 2020-03-20
Update Renderer development readme section

yanchith:


README.md
```diff
@@ -93,18 +93,15 @@ Default is primary GPU backend for the current OS.
 
 ### Renderer development
 
-If working on the renderer, enabling Vulkan validation layers is
-useful for additional validation:
-
-``` shell
-VK_LAYER_PATH=$VULKAN_SDK/bin \
-VK_INSTANCE_LAYERS=VK_LAYER_KHRONOS_VALIDATION \
-cargo run
```
janper, line 102: Is this not valid anymore?
If it still has its place, let's keep it in here.

janper (changes requested):

README.md
yanchith: Removing these lines was the entire point of the PR :smile: 

As the readme change explains (I hope), `gfx` now sets this automatically. This didn't use to be the case when this part of the readme was written.

So no, it is not necessary anymore. You still get the information from the validation layers even if you don't provide those environment variables.

janper (approved):

PR #165, opened 2020-03-22
Review analysis performance

yanchith:
This PR achieves two things:

- First two commits optimize function `edge_sharing`, making it about 2.5x faster in wall time.
- Third commit refactors function `border_edge_loops` and fixes a bug that caused it to loop infinitely for some meshes.

Can be reviewed commit by commit, to see the changes applied one by one.

### DIFF
--- a/src/mesh/analysis.rs
+++ b/src/mesh/analysis.rs
@@ -1,4 +1,4 @@
-use std::collections::HashSet;
+use std::collections::{HashSet, VecDeque};
 
 use nalgebra as na;
 use nalgebra::Point3;
@@ -102,6 +102,7 @@ pub fn edge_sharing<'a, I: IntoIterator<Item = &'a OrientedEdge>>(
     oriented_edges: I,
 ) -> EdgeSharingMap {
     let mut edge_sharing_map = fxhash::FxHashMap::default();
+
     for edge in oriented_edges {
         let unoriented_edge = UnorientedEdge(*edge);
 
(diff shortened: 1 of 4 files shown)


src/mesh/analysis.rs
```diff
@@ -227,19 +256,20 @@ pub fn is_mesh_watertight(edge_sharing: &EdgeSharingMap) -> bool {
...
 ///
+/// This function is slow and is therefore enabled only for tests.
+///
 /// Two mesh geometries are similar when they are visually similar (see the
 /// definition of `are_visually_similar`), and they have the same number of
 /// vertices and normals. Therefore they are going to be treated the same by all
 /// functions of this software and all their transformations result in similar
 /// mesh geometries.
-#[allow(dead_code)]
+#[cfg(test)]
```
janper, line 301: should we mark all similar cases like this instead of dead code?

src/mesh/analysis.rs
```diff
@@ -196,28 +197,57 @@ pub fn border_vertex_indices(edge_sharing: &EdgeSharingMap) -> HashSet<u32> {
...
+            let back = current_chain[current_chain.len() - 1];
+            let loop_closed = back.chains_to(front);
+
+            if loop_closed || !found {
+                break loop_closed;
+            }
+        };
+
+        // There can theoretically exist chains of edges that do not form border
+        // edge loops, e.g. if the mesh is non-manifold. We do not output those.
```
janper, line 246: We should treat such cases, though.

src/mesh/analysis.rs
```diff
@@ -196,28 +197,57 @@ pub fn border_vertex_indices(edge_sharing: &EdgeSharingMap) -> HashSet<u32> {
...
 ///
 /// The mesh may contain holes or islands, therefore it may have an unknown
-/// number of edge loops. If two edge loops meet at a single vertex, the result
-/// may be unpredictable and erratic.
-pub fn border_edge_loops(edge_sharing: &EdgeSharingMap) -> Vec<Vec<UnorientedEdge>> {
-    let mut border_edges: Vec<_> = border_edges(edge_sharing).map(UnorientedEdge).collect();
-
-    let mut edge_loops: Vec<Vec<UnorientedEdge>> = Vec::new();
+/// number of edge loops. If only two edge loops meet at a single vertex
+/// (e.g. if the mesh is non-manifold), this border edge chain is not included
```
janper: Two edge loops meeting at a single point don't make the mesh non manifold! Imagine removing (from an otherwise valid, manifold, orientable, watertight mesh) two faces sharing just a single vertex.

janper (changes requested):
Seems ok, I rather have just some questions inline.

src/mesh/analysis.rs
yanchith: No - this is to prevent us from using these functions in non-test builds. Notice I also added a comment that these functions can only be used in tests. They are now only compiled when running tests.

src/mesh/analysis.rs
yanchith: Could you explain what you mean?

The algorithm treats it. It should be outputting edge *loops* (based on its name), but skips over edge *chains* (or however they should be named). We can alternatively return `Option` or `Result` if the algorithm encounters a non-manifold mesh, where the chains can happen.

src/mesh/analysis.rs
yanchith: Ok, I'll remove that part of the comment.

yanchith: ~Hmm, after drawing it out, I believe the case you describe will produce a single border edge loop from both of the removed faces. Should we stabilize (in tests) and document that behavior?~

Nope I take it back. It is order dependent.

src/mesh/analysis.rs
janper: we've got more such cases though. will we mark them test-only just when they are slow?

src/mesh/analysis.rs
janper: is this what you call a "chain"? because if so, we need to consider the border edge loops as two - that's what I mean by "treating such cases" in my previous comment. I know it's difficult, but apparently such cases are not rare.

src/mesh/analysis.rs
yanchith: Yes. We have a lot of dead code.

> will we mark them test-only just when they are slow?

We should mark them test only if they are only for tests. But yeah, we should try to not have slow code.

src/mesh/analysis.rs
yanchith: > is this what you call a "chain"? 

Not sure what you refer to. I call a "chain" a collection of edges that can be chained, but do not form a closed loop.

Ok, so what is the expected behavior? Also, since the original task was about fixing performance, maybe let's define the behavior in trello and create another task for it?

janper: yes, let's make a fix task.

to answer: in my described scenario, two correct edge loops should be detected. honestly, not sure how.

yanchith: We can jump on call sometimes to discuss.

src/mesh/analysis.rs
yanchith: Or at least I remember us discussing, these functions will only ever be used for tests some time in the past. If that's not true, I can remove the annotation.

src/mesh/analysis.rs
janper: okay, ping me when you are online. let's talk about this before closing this PR

yanchith:
Discussed on slack. All concerns are resolved. We created a task to improve edge case detection behavior of `border_edge_loops`. https://trello.com/c/FNCTyltB/327-improve-borderedgeloops-function-to-inform-of-encountered-edge-cases

PR #166, opened 2020-03-22
Filter out empty models when importing obj file

ondrowan:
I have added filtering of empty models to importer side. Is it okay like this or should I move this functionality to some separate function?

### DIFF
--- a/src/importer.rs
+++ b/src/importer.rs
@@ -139,6 +139,8 @@ impl<C: ObjCache> Importer<C> {
     /// Otherwise, file is read, checksum calculated and cache is checked whether
     /// given file contents were already saved. If not, obj file is parsed and
     /// cached.
+    ///
+    /// Empty models are filtered out of the result.
     pub fn import_obj(&mut self, path: &str) -> ImporterResult {
         let mut file = fs::File::open(path)?;
         let file_metadata = file.metadata().expect("Failed to load obj file metadata");
@@ -159,7 +161,18 @@ impl<C: ObjCache> Importer<C> {
                 let models = match self.cache.get_by_checksum(checksum) {
                     Some(models) => models.clone(),
                     None => {
-                        let (tobj_models, _) = obj_buf_into_tobj(&mut file_contents.as_slice())?;
+                        let (mut tobj_models, _) =
+                            obj_buf_into_tobj(&mut file_contents.as_slice())?;
+                        let mut index = 0;
+
+                        while index != tobj_models.len() {
+                            if tobj_models[index].mesh.positions.is_empty() {
+                                tobj_models.remove(index);
+                            } else {
+                                index += 1;
+                            }
+                        }
+
                         tobj_to_internal(tobj_models)
                     }
                 };
--- a/src/importer.rs
+++ b/src/importer.rs
@@ -13,12 +13,9 @@ use tobj;
 
 use crate::mesh::{Mesh, NormalStrategy, TriangleFace};
 
-// FIXME: ParsingError always comes from tobj. It could contain its LoadError,
-// for more granular error messages, but since it doesn't implement PartialEq,
-// we cannot use it for now.
 #[derive(Debug, PartialEq)]
 pub enum InvalidStructureError {
-    ParsingError,
+    ParsingError(tobj::LoadError),
     DuplicateIndices,
     BlankModel,
 }
@@ -26,7 +23,9 @@ pub enum InvalidStructureError {
 impl fmt::Display for InvalidStructureError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
-            InvalidStructureError::ParsingError => write!(f, "Parsing error."),
+            InvalidStructureError::ParsingError(tobj_err) => {
+                write!(f, "Parsing error: {}", tobj_err)
+            }
             InvalidStructureError::DuplicateIndices => {
                 write!(f, "Duplicate indices were found in one of the models.")
             }
@@ -69,8 +68,8 @@ impl From<io::Error> for ImporterError {
 }
 
 impl From<tobj::LoadError> for ImporterError {
-    fn from(_err: tobj::LoadError) -> Self {
-        ImporterError::InvalidStructure(InvalidStructureError::ParsingError)
+    fn from(err: tobj::LoadError) -> Self {
+        ImporterError::InvalidStructure(InvalidStructureError::ParsingError(err))
     }
 }
 
--- a/src/importer.rs
+++ b/src/importer.rs
@@ -161,18 +161,7 @@ impl<C: ObjCache> Importer<C> {
                 let models = match self.cache.get_by_checksum(checksum) {
                     Some(models) => models.clone(),
                     None => {
-                        let (mut tobj_models, _) =
-                            obj_buf_into_tobj(&mut file_contents.as_slice())?;
-                        let mut index = 0;
-
-                        while index != tobj_models.len() {
-                            if tobj_models[index].mesh.positions.is_empty() {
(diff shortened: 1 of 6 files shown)


src/importer.rs
```diff
@@ -159,7 +161,18 @@ impl<C: ObjCache> Importer<C> {
                 let models = match self.cache.get_by_checksum(checksum) {
                     Some(models) => models.clone(),
                     None => {
-                        let (tobj_models, _) = obj_buf_into_tobj(&mut file_contents.as_slice())?;
+                        let (mut tobj_models, _) =
+                            obj_buf_into_tobj(&mut file_contents.as_slice())?;
+                        let mut index = 0;
+
+                        while index != tobj_models.len() {
+                            if tobj_models[index].mesh.positions.is_empty() {
```
janper: unless you are removing orphans (which you may, I just haven't noticed), its not the `mesh.positions` but rather `mesh.indices` that may make the mesh empty.

janper (changes requested):
Sorry for not noticing sooner but  I think you don't validate the data while importing? We assert the validity of the data when generating a mesh and it seems importing an invalid obj file crashes the program.

A simple validation would be sufficient - check if all the indices are within vertex/normal bounds and check if the faces don't contain two or more same indices. If any of this is true, don't import the file and  report an error instead.

src/importer.rs
yanchith: True. How do you propose we best solve this? We could change the mesh constructors to return `Option`?

src/importer.rs
```diff
@@ -159,7 +161,18 @@ impl<C: ObjCache> Importer<C> {
...
+                                tobj_models.remove(index);
+                            } else {
+                                index += 1;
+                            }
+                        }
```
yanchith: Could we move the functionality you added to `tobj_to_internal`? Empty OBJs are completely valid. It is just our software doesn't accept them :joy:

src/importer.rs
janper: no need for an `Option` here, i think. just check if the `indices` contain anything, instead of `positions`.

i think it's ok when the mesh constructor panicks. we should feed the mesh constructors valid data. 
therefore i§d suggest validating the imported data first. should be fairly simple:

- the faces should reference an existing vertex (and normal, if present) index
- the faces should contain three non-repetetive indices
- the mesh has to have at least one valid face

we don't need to remove orphans during import, but we need to decide what should we do if any of the above mentioned tests fails. i suggest not trying to autofix but rather fail the import instead.

ondrowan: I agree with failing the import. We shouldn't let the user think everything went well while some models are missing.

ondrowan: > the faces should reference an existing vertex (and normal, if present) index

There's unfortunatelly a bug in `tobj` that crashes conversion if there's index out of bound. I filed a bug report for it.

janper: do we have a workaround meanwhile?

ondrowan: Not anything that would be easy to implement on our side. One thing we can do later, if it isn't fixed upstream, is to fork (or just copy, it's 1 file) the library and fix it on our side.

janper: that would be much appreciated! we can't wait until somebody sometime fixes a bug

ondrowan: Let's give that maintainer at least couple of days before we fix this ourselves. I don't have capacity for it right now.  Would that be possible?

ondrowan:
I have revamped how this patch works. If there's an empty model or some vertex uses duplicate indices, it errors out explicitly. The case with index out of bounds will be fixed later, once it's fixed in `tobj`, as discusses in one of the comment threads.

I hope I understood correctly how this should behave. I'm also not sure about error messages / naming, so help me out with that.

src/importer.rs
```diff
@@ -13,19 +13,41 @@ use tobj;
 
 use crate::mesh::{Mesh, NormalStrategy, TriangleFace};
 
+// FIXME: ParsingError always comes from tobj. It could contain its LoadError,
+// for more granular error messages, but since it doesn't implement PartialEq,
+// we cannot use it for now.
+#[derive(Debug, PartialEq)]
+pub enum InvalidStructure {
```
yanchith: Isn't this pretty much an error? Should probably be named `InvalidStructureError` then, and implement `error::Error`.

src/importer.rs
```diff
@@ -190,10 +215,16 @@ pub fn obj_buf_into_tobj(file_contents: &mut &[u8]) -> tobj::LoadResult {
 /// Converts `tobj::Model` vector into vector of internal `Model` representations.
 /// It expects valid `tobj::Model` representation, eg. number of positions
 /// divisible by 3.
-pub fn tobj_to_internal(tobj_models: Vec<tobj::Model>) -> Vec<Model> {
+pub fn tobj_to_internal(tobj_models: Vec<tobj::Model>) -> Result<Vec<Model>, ImporterError> {
```
yanchith: Could just return `InvalidStructureError`. If you add a `From` impl for `ImporterError` that takes `InvalidStructureError`, the `?` operator will pick it up and use it. That means you don't have to wrap every error in this function.

src/importer.rs
```diff
@@ -290,7 +329,8 @@ mod tests {
             vec![1.0, 0.0, 0.0, 1.0, 0.0, 0.0, 1.0, 0.0, 0.0],
         );
         let tobj_models = vec![tobj_model.clone()];
-        let models = tobj_to_internal(tobj_models);
+        let models = tobj_to_internal(tobj_models)
```
yanchith, line 337: I'd just unwrap in tests, but feel free to leave those expects there.

src/importer.rs
```diff
@@ -13,19 +13,41 @@ use tobj;
 
 use crate::mesh::{Mesh, NormalStrategy, TriangleFace};
 
+// FIXME: ParsingError always comes from tobj. It could contain its LoadError,
+// for more granular error messages, but since it doesn't implement PartialEq,
+// we cannot use it for now.
```
ondrowan: I have already created PR to `tobj` to derive `PartialEq` for `LoadError`.

yanchith (commented):
Looks good to me. Got some rust error handling improvement suggestions.

src/importer.rs
ondrowan: I didn't want to type a lot, but here I am, typing a lot. Fixed it.

src/importer.rs
```diff
@@ -17,37 +17,41 @@ use crate::mesh::{Mesh, NormalStrategy, TriangleFace};
...
 }
 
+impl error::Error for InvalidStructureError {}
+
 #[derive(Debug, PartialEq)]
 pub enum ImporterError {
     FileNotFound,
     PermissionDenied,
-    InvalidStructure(InvalidStructure),
+    InvalidStructureError(InvalidStructureError),
```
yanchith: Just a note that in enums, it usually looks like this:

```rust
pub enum SomeError {
    Reason1,
    Reason2(Reason2Error), // Note that this is not: Reason2Error(Reason2Error),
}
```
https://blog.burntsushi.net/rust-error-handling/#composing-custom-error-types

yanchith:
@ondrowan looks like the fix made it https://github.com/Twinklebear/tobj/commit/b798027c11d0fa9c7bb1af26a386968a38143d36 and is released on crates (https://crates.io/crates/tobj/1.0.0). Can we please finish this before Monday (30th)? We need to provide a build to SCD by then.

src/importer.rs
ondrowan: Oops, it seems this is something RA did and I didn't notice it.

yanchith (approved):
I am happy. @janper ?

janper (approved):

PR #167, opened 2020-03-29
Fix interpreter caching bug

yanchith:
This is the most fascinating bug of all, if you know what I mean :)

When we interleave editing the program with executions (what always happens with autorun, but could be achieved manually as well), it is possible to create a variable value in the interpreter that replaces another variable value without the interpreter noticing. 

The parameters of the function call that creates that value can potentially reference completely different variable values later on, but since those referenced values were not invalidated, the value of the variable created by referencing them also wasn't.

The fix is to attach a "timestamp" (`created_epoch` field) to each stored variable value and use that to determine if a variable value could have legally be constructed by referencing another variable value. The other value, while its declaration being earlier in the program, could have been replaced by subsequent program edits and executions.

As a FIXME in the patch suggests, another possible fix could have been to never keep stale values in the interpreter environment and conservatively remove values from the environment whenever the variable declaration statements creating them are edited. This would cause us to loose some caching opportunities, though.

### DIFF
--- a/src/interpreter/mod.rs
+++ b/src/interpreter/mod.rs
@@ -311,11 +311,18 @@ pub struct InterpretValue {
     pub unused_values: Vec<(VarIdent, Value)>,
 }
 
+/// A variable value stored in the environment.
+///
+/// Contains metadata helpful to determine when the value should be invalidated
+/// or forgotten.
 #[derive(Debug, Clone)]
-struct VarInfo {
+struct VarValue {
     /// The parameters and function call this variable was created
     /// with. Used to verify validity of this variable.
     created_call: ast::CallExpr,
+
+    /// The epoch this variable was created.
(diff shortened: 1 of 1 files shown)


janper (approved):
I understand the solution as much as I could ever understand it. It seems to be a valid fix and it doesn't cost that much - only that sometimes the values are invalidated even when they don't need to be?
We can live with it.

yanchith:
> only that sometimes the values are invalidated even when they don't need to be?

You are correct. We could be more precise, but it would also be more costly to be more precise, both in terms of runtime spent on invalidation checking, and algorithm complexity. I didn't feel comfortable to choose in favor of more complexity, but we can reopen this later.

janper:
Great! Happy I understood and correctly approved the change.

PR #168, opened 2020-03-29
Restore previous numbering of variable names in the pipeline window

yanchith:
Previously, we used the statement index to generate part of the displayed
variable names in the combo boxes. In the UI however, we did rely on the fact
that the var declaration statement index was the same as the internal raw value
of the variable indentifier.

This patch removes that incorrect assumption and once again uses the statement
index to generate the displayed human readable name.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -805,16 +805,20 @@ pub fn init_and_run(options: Options) -> ! {
                             .map(|(value_path, (_, mesh))| {
                                 if value_path.1 == 0 {
                                     // Do not suffix zero mesh-array index
-                                    let name = match session.var_name_for_ident(value_path.0) {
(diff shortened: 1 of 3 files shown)


janper (approved):
So this treats the cases when due to the different epochs, the UI names can be confusing, right?

It seems to look the same as before on the front-end, so I believe it's all good.

yanchith:
> So this treats the cases when due to the different epochs, the UI names can be confusing, right?

Yes, the screenshot attachment of the trello issue describes the problem: https://trello.com/c/OnkRkNhP/328-variable-names-do-not-correspond-to-the-operation-names-that-created-them

janper:
Great! Happy I understood and correctly approved the change.

PR #169, opened 2020-03-29
Add voxel transform operation

janper:
Adds a voxel transform operation, that is useful in two scenarios:

- as a supplemental function, it implements scalar field resampling, which is nice to have when we need to change some of its properties but keep the contents
- it will be a crucial part of an autofit function that aligns two geometries with minimal deviation

One more Scalar Field related PR will come after this <- it will be just a maintenance and cleanup commit.


src/interpreter_funcs/mod.rs
```diff
@@ -138,7 +141,7 @@ pub fn create_function_table() -> BTreeMap<FuncIdent, Box<dyn Func>> {
     );
     funcs.insert(FUNC_ID_LOOP_SUBDIVISION, Box::new(FuncLoopSubdivision));
 
-    // Voxel-based funcs: 8xxx
+    // Voxel-based funcs
```
yanchith, line 144: Let's keep the comment as it was.

src/interpreter_funcs/voxel_transform.rs
```diff
@@ -0,0 +1,312 @@
...
+                \n\
+                High values produce coarser results, low values may increase precision but produce \
+                heavier geometry that significantly affect performance. Too high values produce \
+                single large voxel, too low values may generate holes in the resulting geometry.",
+                refinement: ParamRefinement::Float3(Float3ParamRefinement {
+                    min_value: Some(0.005),
+                    max_value: None,
+                    default_value_x: Some(0.25),
+                    default_value_y: Some(0.25),
+                    default_value_z: Some(0.25),
```
yanchith: Hmm, except for noise (which has 0.1), every other voxel operation has a default 1.0. Want to keep it this way, or re-evaluate? We can resize the primitives to have default dimensions 10, e.g.

src/interpreter_funcs/voxel_transform.rs
```diff
@@ -0,0 +1,312 @@
...
+        let error_if_large = args[8].unwrap_boolean();
+        let analyze_bbox = args[9].unwrap_boolean();
+        let analyze_mesh = args[10].unwrap_boolean();
+
+        if voxel_dimensions.iter().any(|dim| dim <= &0.0) {
+            return {
+                let error = FuncError::new(FuncVoxelTransformError::VoxelDimensionZeroOrLess);
+                log(LogMessage::error(format!("Error: {}", error)));
+                Err(error)
+            };
```
yanchith: Let's remove the inner block. Doesn't seem to be adding any value.

src/interpreter_funcs/voxel_transform.rs
```diff
@@ -0,0 +1,312 @@
...
+                    log(LogMessage::error(format!("Error: {}", error)));
+                    Err(error)
+                }
+            }
+        } else {
+            {
+                let error = FuncError::new(FuncVoxelTransformError::TransformFailed);
+                log(LogMessage::error(format!("Error: {}", error)));
+                Err(error)
+            }
```
yanchith, line 321: Let's remove the inner block.

src/interpreter_funcs/voxel_transform.rs
```diff
@@ -0,0 +1,312 @@
...
+
+        let scaling = Vector3::from(scale);
+
+        if let Some(transformed_sf) = ScalarField::from_scalar_field_transformed(
+            &voxel_cloud,
+            &(0.0..=0.0),
+            &voxel_dimensions,
+            &translate,
+            &rotation,
+            &scaling,
```
yanchith: Either translate/rotate/scale or translation/rotation/scaling? I think we mostly use the former.

src/mesh/voxel_cloud.rs
```diff
@@ -214,6 +213,211 @@ impl ScalarField {
...
+                        &target_scalar_field.block_dimensions,
+                        &target_scalar_field.voxel_dimensions,
+                    );
+
+                    // Set the new voxel value according to a sampled value from
+                    // the source scalar field.
+                    let absolute_coordinate_source = cartesian_to_absolute_voxel_coordinate(
+                        &cartesian_coordinate_target,
+                        &source_scalar_field.voxel_dimensions,
+                    );
```
yanchith: Let's not do it now, because there are unresolved design questions, but we could use this pair of functions:

- `ScalarField::value_at_cartesian_coordinate`
- `ScalarField::set_value_at_cartesian_coordinate`

This would bring us one step closer to `ScalarField` having an orthogonal API and removing most of it's methods to functions.

The main unresolved design question is: How should `set_value_at_cartesian_coordinate` behave for out of bounds accesses. It is relatively ok to assert with integer coordinates when the caller should be able to easily check the math, but this is a bit more tricky, because a rounding error could suddenly cause a panic.

src/mesh/voxel_cloud.rs
```diff
@@ -214,6 +213,211 @@ impl ScalarField {
...
+                // determine the minimal needed bounding box due to the
+                // differences in the voxel sizes. Leaving the scalar field
+                // bigger should be also considered.
+                target_scalar_field.shrink_to_fit(volume_value_range);
+
+                target_scalar_field
+            })
+    }
+
+    /// Creates a new scalar field from another scalar field transformed
```
yanchith, line 278: Similarly to the above doc comment, let's also mention that voxel dimensions can change.

src/mesh/voxel_cloud.rs
```diff
@@ -214,6 +213,211 @@ impl ScalarField {
...
+    /// # Panics
+    ///
+    /// Panics if the voxel dimension is below or equal to zero.
+    pub fn from_scalar_field_transformed<U>(
+        source_scalar_field: &ScalarField,
+        volume_value_range: &U,
+        voxel_dimensions: &Vector3<f32>,
+        cartesian_translation: &Vector3<f32>,
+        rotation: &Rotation3<f32>,
+        scale_factor: &Vector3<f32>,
```
yanchith: translate/rotate/scale?

src/mesh/voxel_cloud.rs
```diff
@@ -214,6 +213,211 @@ impl ScalarField {
...
+                    voxels: source_scalar_field.voxels.to_vec(),
+                });
+            } else {
+                // or resample to new voxel dimensions and return.
+                return ScalarField::from_scalar_field(
+                    source_scalar_field,
+                    volume_value_range,
+                    voxel_dimensions,
+                );
+            }
```
yanchith: No need for this if/else. `ScalarField::from_scalar_field` already handles this edge case.

src/mesh/voxel_cloud.rs
```diff
@@ -214,6 +213,211 @@ impl ScalarField {
...
+        }
+
+        assert!(
+            voxel_dimensions.x > 0.0 && voxel_dimensions.y > 0.0 && voxel_dimensions.z > 0.0,
+            "One mor more voxel dimensions is below or equal to zero"
+        );
```
yanchith, line 303: Let's move the assert up. Similarly to how it is positioned in the other function.

src/mesh/voxel_cloud.rs
```diff
@@ -450,6 +654,42 @@ impl ScalarField {
         tools::weld(&joined_voxel_mesh, (min_voxel_dimension as f32) / 4.0)
     }
 
+    /// Returns the bounding box of the mesh produced by `ScalarField::to_mesh`
```
yanchith, line 691: What is the relationship of the bounding box of mesh produced by `ScalarField::to_marching_cubes` compared to this one? Should it be documented?

src/mesh/voxel_cloud.rs
```diff
@@ -450,6 +654,42 @@ impl ScalarField {
...
+    /// for the current scalar field.
+    ///
+    /// The Bounding box will be defined in cartesian units.
+    ///
```
yanchith: Remove this line

src/mesh/voxel_cloud.rs
```diff
@@ -450,6 +654,42 @@ impl ScalarField {
...
+                        (volume_start.z as f32 + volume_dimensions.z as f32 + 0.5)
+                            * voxel_dimensions.z,
+                    ),
+                )
+            },
+        )
+    }
+
+    /// Compute discrete distance field.
+    ///
```
yanchith: This comment doesn't make sense to me in its current location.

src/mesh/voxel_cloud.rs
```diff
@@ -214,6 +213,211 @@ impl ScalarField {
...
+
+        assert!(
+            voxel_dimensions.x > 0.0 && voxel_dimensions.y > 0.0 && voxel_dimensions.z > 0.0,
+            "One mor more voxel dimensions is below or equal to zero"
+        );
+
+        // Make a bounding box of the source scalar field's mesh volume. This
+        // will be the volume to be scanned for any voxels.
+        if let Some(source_sf_bounding_box) =
+            source_scalar_field.bounding_box_mesh_volume_cartesian_space(volume_value_range)
```
yanchith: Why mesh volume and not some other? Not obvious to me. Is it optimization to sample less space?

The thing is, this causes the generated mesh to be offset diagonally sideways. Try creating a box, and using voxel transform move on it. Once the translation becomes nonzero (and doesn't land in the happy edge case), the mesh will snap elsewhere. This probably shouldn't be happening, and it doesn't happen if you use `bounding_box_cartesian_space`.

Is it possible, that because we consider voxel-space coordinates to be centers of voxels, we somehow added/subtracted 0.5 twice or in the wrong order? The error shouldn't accumulate. This doesn't add up and I think it needs some research and explanation.

Note that the error might not be here, but it looks like it is somewhere between here and the first voxelization of the original mesh.

src/mesh/voxel_cloud.rs
```diff
@@ -214,6 +213,211 @@ impl ScalarField {
...
+    ///
+    /// Panics if the voxel dimension is below or equal to zero.
+    pub fn from_scalar_field_transformed<U>(
+        source_scalar_field: &ScalarField,
+        volume_value_range: &U,
+        voxel_dimensions: &Vector3<f32>,
+        cartesian_translation: &Vector3<f32>,
+        rotation: &Rotation3<f32>,
+        scale_factor: &Vector3<f32>,
+    ) -> Option<Self>
```
yanchith, line 296: Let's document when this returns `None`

src/mesh/voxel_cloud.rs
```diff
@@ -214,6 +213,211 @@ impl ScalarField {
...
+    /// scalar field.
+    ///
+    /// # Panics
+    ///
+    /// Panics if any of the voxel dimensions is below or equal to zero.
+    pub fn from_scalar_field<U>(
+        source_scalar_field: &ScalarField,
+        volume_value_range: &U,
+        voxel_dimensions: &Vector3<f32>,
+    ) -> Option<Self>
```
yanchith, line 229: Let's document when this returns `None`

yanchith (changes requested):
This is just the first part of the review. I didn't review the tests yet, I will get to that tomorrow. It is highly suspicious that they add 15k lines worth of snapshots though, so perhaps you can also add an explanation of that? The thing is snapshots of that size are impossible to review (did you review them?), beating the purpose of snapshots in the first place.

Apart from that I do have a couple of minor suggestions and I believe I found some strange behavior. I am not sure the reason for that behavior was added in this PR, or earlier (and we just didn't catch it). See inline.

src/interpreter_funcs/mod.rs
janper: Why? This would be the only place where we say that, also it's irrelevant in this context. In the block of code above it makes more sense.

src/interpreter_funcs/voxel_transform.rs
janper: I remember we decided to keep the voxel scale 1.0 everywhere. I'll change it here as well. 
I'll also change the noise to 1.0 and scale up the block it occupies. This way we encourage the user to generate geometry of a similar scale.

src/mesh/voxel_cloud.rs
janper: How would you iterate through the target scalar field? 
I can imagine iterating just like now and assigning `*voxel` the value from `value_at_cartesian_coordinate`.

src/mesh/voxel_cloud.rs
janper: the bounding box should be identical

src/mesh/voxel_cloud.rs
janper: I'll look into it. Th reason for taking the mesh volume bounding box is the possible difference in voxel dimensions. 
If this causes some offset, the error must be somewhere in the math rather than in the usage of mesh volume.

src/interpreter_funcs/voxel_noise.rs
```diff
@@ -78,9 +78,9 @@ impl Func for FuncVoxelNoise {
                 refinement: ParamRefinement::Float3(Float3ParamRefinement {
                     min_value: None,
                     max_value: None,
-                    default_value_x: Some(1.0),
-                    default_value_y: Some(1.0),
-                    default_value_z: Some(1.0),
+                    default_value_x: Some(10.0),
+                    default_value_y: Some(10.0),
+                    default_value_z: Some(20.0),
```
yanchith, line 83: ~Are these block start and block end changes on purpose?~

Nevermind, I see. Shouldn't the Z values be the same as X and Y, though?

src/mesh/voxel_cloud.rs
```diff
@@ -688,7 +683,9 @@ impl ScalarField {
         )
     }
 
-    /// Compute discrete distance field.
+    /// Computes a relatively smooth triangulated welded mesh from the current
```
yanchith, line 724: Now the comment is duplicated

src/mesh/voxel_cloud.rs
yanchith: Yes, something like that. True that most if not all of the current algorithms don't need `set_value_at_cartesian_coordinate`, if you have some version of `iter_mut`.

We can provide our own `values_mut` or `voxels_mut` (we should probably use just one of those terms) that is like `enumerate`, but gives you the absolute coordinate instead - similarly to what the `image` crate does with pixels.

src/interpreter_funcs/mod.rs
yanchith: What are you referring to? I am referring to the change from:
```
// Voxel-based funcs: 8xxx
```
to
```
// Voxel-based funcs
```

src/mesh/voxel_cloud.rs
yanchith: You are right - we still should be thinking that (0,0,0) in cartesian coordinates is the center of the voxel. Now to just find the meth error...

src/interpreter_funcs/mod.rs
janper: please, have a better look at lines 63 to 94 and compare them to lines 111 to 158

src/interpreter_funcs/voxel_noise.rs
janper: because the shadow is ugly when the geometry intersects the ground plane. therefore I put it above

src/mesh/voxel_cloud.rs
janper: This merely moves the same logic into separate functions.

I think we were already discussing the iterator over the voxel values. I prefer skipping that improvement for now.

src/interpreter_funcs/mod.rs
yanchith: Ugh, I see

src/mesh/voxel_cloud.rs
yanchith: Yes. That's why I said "Let's not do it now, ..." :slightly_smiling_face:

src/interpreter_funcs/voxel_noise.rs
yanchith: If you want to, we can be inconsistent here and keep it like this, but a lot of other things are not snapped to ground by default (and therefore ugly). What do we want to do about that?

src/mesh/voxel_cloud.rs
```diff
@@ -3021,6 +3258,172 @@ mod tests {
...
+            &scalar_field,
+            &(0.0..=0.0),
+            &Vector3::new(0.25, 0.25, 0.25),
+            &Vector3::new(0.1, 0.1, 0.1),
+            &Rotation3::from_euler_angles(0.0, 0.0, 0.0),
+            &Vector3::new(1.0, 1.0, 1.0),
+        )
+        .unwrap();
+
+        insta::assert_json_snapshot!(
```
yanchith: How did you hand-validate this snapshot? I can't think of a way. Same for the snapshots in tests below.

yanchith (commented):
Tests are now reviewed.

Is there trick to validating the scalar field snapshots? Did you hand-validate them? Because if not, we don't know they encode the correct specification. Unless I am missing something, I think it better to have no tests, than having tests that potentially specify incorrect behavior.

src/interpreter_funcs/voxel_noise.rs
janper: The other generators create primitives and it makes sense to create unitized versions of them. I can't imagine a unit version of a noise. I want the user to get an instantly gratifying result. Sure, we can wrap it around zero, but once again, what would it be good for in this case? Moreover, it is 10 times larger than any other geometry generated inside of our software (there is no good solution to this - either the noise is big or the voxels are too small, or the result looks like a single voxel).

That brings me to another issue - we know the models that will be most probably imported int our software and all of them are thousands of units big. Shouldn't we set our defaults to generate roughly 1000 units large results and operate correctly on such inputs?

src/mesh/voxel_cloud.rs
janper: I used the same values in the UI and eyeballed the result.

src/interpreter_funcs/voxel_noise.rs
yanchith: > Shouldn't we set our defaults to generate roughly 1000 units large results and operate correctly on such inputs?

That's what I was aiming at.

But let's do it in another task and another PR to prevent this one from getting intractable.
https://trello.com/c/VtG2C1Le/333-adjust-default-values-in-operations-for-more-instantly-gratifying-results

yanchith: Also, let's keep this one as is. We can potentially adjust it in the other task, if needed.

src/mesh/voxel_cloud.rs
yanchith: Okay. In this case, let's remove the snapshot tests from this PR. They don't add any layer of safety.

For the future: If you think of a way of testing our kind of data, let me know. I am out of ideas :slightly_frowning_face: , perhaps except of having a PR checklist of things to test manually (which we probably will need to do).

src/mesh/voxel_cloud.rs
janper: Now when I gave ti second, third and fourth thought, I'm not sure anymore if the meshing volume actually matters. In the current iteration, I'm trimming the bounding box to the voxel centers. I'm actually not sure if there is a correct answer.

src/mesh/voxel_cloud.rs
```diff
@@ -214,6 +213,191 @@ impl ScalarField {
...
+                        one_dimensional_target,
+                        &target_scalar_field.block_start,
+                        &target_scalar_field.block_dimensions,
+                        &target_scalar_field.voxel_dimensions,
+                    );
+
+                    // Transform each new voxel inverse to the user
+                    // transformation.
+                    let transformed_voxel_center_cartesian = reversed_user_transformation
+                        .transform_point(&(cartesian_coordinate_target - vector_to_origin));
```
janper: I suspect this line to cause the offset

src/mesh/voxel_cloud.rs
```diff
@@ -688,10 +682,9 @@ impl ScalarField {
         )
     }
 
-    /// Compute discrete distance field.
-    ///
     /// Computes a relatively smooth triangulated welded mesh from the current
-    /// state of the scalar field.
+    /// state of the scalar field. The mesh will be made of triangular faces
+    /// positioned in discrete 40 degree steps.
```
yanchith: 40? Not 45?

src/mesh/voxel_cloud.rs
```diff
@@ -466,6 +446,21 @@ impl ScalarField {
...
+    /// Returns None if voxel is empty or out of bounds
+    pub fn value_at_cartesian_coordinate(&self, cartesian_coordinate: &Point3<f32>) -> Option<f32> {
+        match cartesian_to_one_dimensional_coordinate(
+            cartesian_coordinate,
+            &self.block_start,
+            &self.block_dimensions,
+            &self.voxel_dimensions,
+        ) {
+            Some(index) => self.voxels[index],
+            _ => None,
```
yanchith: Just a matter of style, but for enums that you know won't be extended, I'd not use `_`. Also, in this case, I believe `map` is more readable.

src/mesh/voxel_cloud.rs
yanchith: Not sure I follow

src/mesh/voxel_cloud.rs
yanchith: > Now when I gave ti second, third and fourth thought, I'm not sure anymore if the meshing volume actually matters.

Could you explain?

> In the current iteration, I'm trimming the bounding box to the voxel centers.

Won't this be less symmetrical? Most geometry will be centered around origin. Don't we want this to give good results (not shifting the geometry) for these cases? 

> I'm actually not sure if there is a correct answer.

It is a bit of an arbitrary decision, but having voxels in voxel space be centers of voxels in catesian space (instead of front-bottom-left corners) seems like a good decision.

src/mesh/voxel_cloud.rs
yanchith: Btw not sure if you intended to fix it, but the shift still happens.

janper: shift happens

janper: can you please share your testing parameters and a screenshot of a faulty result?
i believed i fixed the issue, so i think i should test more thoroughly

src/mesh/voxel_cloud.rs
janper: it's not the same though. `map(|x| x)` wraps `x` into an `Option`. in our case it already is an `Option`, so using `map` results in `Option<Option<_>>`.

yanchith: `Option::and_then`?

yanchith:
Sorry for not responding here. I am at the moment quite overwhelmed by other stuff. Should be able to respond in early May, hopefully.

src/mesh/voxel_cloud.rs
yanchith: Okay, will do so (when I get to this PR in May).

janper:
Ready for another review

src/mesh/voxel_cloud.rs
```diff
@@ -404,17 +388,13 @@ impl ScalarField {
...
-
-                // FIXME: @Optimization In some cases it might be not easy to
-                // determine the minimal needed bounding box due to the
-                // differences in the voxel sizes. Leaving the scalar field
-                // bigger should be also considered.
+                // FIXME: Consider not shrinking to fit because in most cases it
+                // won't be much larger than it has to be.
                 target_scalar_field.shrink_to_fit(volume_value_range);
 
                 return Some(target_scalar_field);
```
janper: I think sometimes the target scalar field may be larger than necessary, so `shrink_to_fit` may be appropriate. On the other hand, it may be even slower to shrink to fit and it may be smarter to keep the scalar field unnecessarily big. That applies to all functions shrinking the result scalar field. Maybe the shrinking could only happen in the operations and also only when there is too much obsolete space.

janper: I mean the shrink could be called explicitly rather than inside functions, that may or may not result in an unnecessarily large scalar field. Because maybe the caller wants to get the entire scalar field rather than its trimmed version.

src/bounding_box.rs
```diff
@@ -91,6 +91,7 @@ impl<T: Bounded + Scalar + Num + PartialOrd> BoundingBox<T> {
     pub fn from_points<I>(points: I) -> Option<Self>
     where
         I: IntoIterator<Item = Point3<T>>,
+        T: std::fmt::Display,
```
yanchith: You probably forgot to remove this after debugging

src/bounding_box.rs
```diff
@@ -139,6 +143,7 @@ impl<T: Bounded + Scalar + Num + PartialOrd> BoundingBox<T> {
     pub fn union<I>(bounding_boxes: I) -> Option<Self>
     where
         I: IntoIterator<Item = BoundingBox<T>>,
+        T: std::fmt::Display,
```
yanchith: You probably forgot to remove this after debugging

src/mesh/voxel_cloud.rs
```diff
@@ -214,6 +213,237 @@ impl ScalarField {
...
+
+        // If the target scalar field should be identical
+        if approx::relative_eq!(*voxel_dimensions, source_scalar_field.voxel_dimensions) {
+            // return identical copy of self.
+            return Some(ScalarField {
+                block_start: source_scalar_field.block_start,
+                block_dimensions: source_scalar_field.block_dimensions,
+                voxel_dimensions: source_scalar_field.voxel_dimensions,
+                voxels: source_scalar_field.voxels.to_vec(),
+            });
```
yanchith: `source_scalar_field.clone()`?

src/interpreter_funcs/voxel_transform.rs
```diff
@@ -0,0 +1,327 @@
...
+            },
+            ParamInfo {
+                name: "Detailed Mesh Analysis",
+                description: "Reports detailed analytic information on the created mesh.\n\
+                              The analysis may be slow, therefore it is by default off.",
+                refinement: ParamRefinement::Boolean(BooleanParamRefinement {
+                    default_value: false,
+                }),
+                optional: false,
+            },
```
yanchith: This was updated in master

yanchith (changes requested):
Almost there! I found a couple more small issues.

The issue about the offset once the transform is not identity seems to have been fixed 🎉 

TBH I didn't have much energy to go over the `from_scalar_field_transformed` function again in great detail, so it is what it is for now.

Also please update against master before the last review.

src/mesh/voxel_cloud.rs
yanchith: > Maybe the shrinking could only happen in the operations and also only when there is too much obsolete space.

Good idea! I am creating an issue in our board.

yanchith: https://trello.com/c/2XSIdfuH/343-only-shrink-voxel-clouds-scalar-fields-to-fit-explicitly-in-operations

janper:
My job is done here

yanchith (approved):

PR #170, opened 2020-03-30
Prevent imgui_winit_support changing cursor icon if outside window

yanchith:
For some reason, this problem only manifested on Windows. Maybe other OSes don't
allow apps changing the cursor in certain priority conditions?

Tested on:

- [x] Windows (fixes the problem)
- [x] macOS (worked fine before and after)
- [x] X11 (worked fine before and after)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -138,6 +138,7 @@ pub fn init_and_run(options: Options) -> ! {
 
     let window = winit::window::WindowBuilder::new()
         .with_title(BASE_WINDOW_TITLE)
+        .with_inner_size(winit::dpi::LogicalSize::new(1280, 720))
         .with_maximized(true)
         .with_window_icon(Some(icon))
         .build(&event_loop)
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -138,6 +138,7 @@ pub fn init_and_run(options: Options) -> ! {
 
     let window = winit::window::WindowBuilder::new()
         .with_title(BASE_WINDOW_TITLE)
+        .with_inner_size(winit::dpi::LogicalSize::new(1280, 720))
```
janper, line 141: why do we need this?

janper (approved):

src/lib.rs
yanchith: This is very interesting and I learned about this just recently (when contributing a fix to winit about `WindowBuilder::with_maximized`). On Windows, and perhaps other platforms as well, "maximized" is only a state. When that state is exited, the window returns to the size it previously had. By specifying a size, we can ensure that size is something usable.

yanchith: So specifying both `with_maximized` and `with_inner_size` actually has meaning.

PR #171, opened 2020-03-31
Edge loop cases

janper:
- Detect edge loops for non-orientable and non-deterministic meshes
- Join operation analytics, set their default to off


src/mesh/analysis.rs
```diff
@@ -1157,12 +1309,15 @@ mod tests {
...
         ];
 
-        let computed_loops = border_edge_loops(&edge_sharing_map);
-
-        assert_eq!(computed_loops.len(), 1);
-        assert_eq!(computed_loops[0].len(), correct_loop.len());
-        for edge in correct_loop {
-            assert!(computed_loops[0].iter().any(|e| *e == edge));
+        if let BorderEdgeLoopsResult::Found(computed_loops) = border_edge_loops(&edge_sharing_map) {
+            dbg!("{}", computed_loops.clone());
```
yanchith: Remove dbg

src/mesh/analysis.rs
```diff
@@ -1213,31 +1368,100 @@ mod tests {
...
+        );
+
+        let oriented_edges: Vec<OrientedEdge> = mesh.oriented_edges_iter().collect();
+        let edge_sharing_map = edge_sharing(&oriented_edges);
+
+        if let BorderEdgeLoopsResult::FoundWithNondeterminism(edge_loops) =
+            border_edge_loops(&edge_sharing_map)
+        {
+            assert!(
+                edge_loops.len() > 1,
```
yanchith: Are you sure? Isn't this one-or-more?

src/mesh/analysis.rs
```diff
@@ -1213,31 +1368,100 @@ mod tests {
...
-                }
-            }
+        let oriented_edges: Vec<OrientedEdge> = mesh.oriented_edges_iter().collect();
+        let edge_sharing_map = edge_sharing(&oriented_edges);
+
+        if let BorderEdgeLoopsResult::FoundWithNondeterminism(edge_loops) =
+            border_edge_loops(&edge_sharing_map)
+        {
+            assert!(
+                edge_loops.len() > 1,
```
yanchith: Can we be sure? Isn't this one-or-more?

src/mesh/analysis.rs
```diff
@@ -1213,31 +1368,100 @@ mod tests {
...
+                    }
+                }
+            } else {
+                assert_eq!(computed_loops[1].len(), correct_loops[0].len());
+                for (i, correct_loop) in correct_loops.iter().enumerate() {
+                    for edge in correct_loop {
+                        assert!(computed_loops[(i + 1) % 2].iter().any(|e| e == edge));
+                    }
+                }
+            }
```
yanchith: Isn't this a bit too generic? We know there are exactly two loops. Should probably extract them to variables and do comparisons on those instead of obfuscating with `[(i + 1) % 2]`, especially if the expression just means "the other of the two loops".

src/mesh/analysis.rs
```diff
@@ -193,14 +193,57 @@ pub fn border_vertex_indices(edge_sharing: &EdgeSharingMap) -> HashSet<u32> {
...
+    if border_edges.is_empty() {
+        return BorderEdgeLoopsResult::Watertight;
+    }
+
+    // Check if the border edges are branching (the loops are touching at a
+    // single vertex). If they are, the result is non-deterministic.
+
+    // FIXME: @Optimization: use a Map structure or at least trim the list of
+    // vertices also from below. Find the maximum index that occurs in the
+    // border edge loop.
```
yanchith, line 228: Separate FIXME from explanatory comment.

yanchith: Also, not sure by what you mean by the FIXME. This does a single iteration, which is not too bad, considering what the rest of the algorithm does.

src/mesh/analysis.rs
```diff
@@ -193,14 +193,57 @@ pub fn border_vertex_indices(edge_sharing: &EdgeSharingMap) -> HashSet<u32> {
...
+        vertex_occurrence_counts[cast_usize(edge.vertices.1)] += 1;
+    }
+
+    // For a deterministic result, each vertex should occur in the list of
+    // boundary edges exactly twice - in two edges, which meet at the vertex.
+    // Higher number means that more edge loops meet at one vertex and therefore
+    // the edge loop list is non-deterministic and possible wrong. If the vertex
+    // appears only once, the mesh may be erroneous. There are holes in the list
+    // of vertex occurrence, therefore some items may be zero.
+    // FIXME: Panic if the occurrence is 1
```
yanchith: I'd change this FIXME to `FIXME: Examine situation when occurrence is 1 and determine what to do`. For now, we can just assume non-determinism in this case too, although IIRC that situation didn't seem possible from our explorations.

src/mesh/analysis.rs
```diff
@@ -222,9 +265,27 @@ pub fn border_edge_loops(edge_sharing: &EdgeSharingMap) -> Vec<Vec<OrientedEdge>
...
+                    found = true;
+
+                    break;
+                }
+
                 let back = current_chain[current_chain.len() - 1];
                 if back.chains_to(*other_edge) {
-                    current_chain.push_back(border_edges.remove(i));
+                    current_chain.push_back(*other_edge);
+                    found = true;
```
yanchith: For symmetry and therefore readability, `found` should be last, as is in the other blocks.

src/mesh/analysis.rs
```diff
@@ -222,9 +265,27 @@ pub fn border_edge_loops(edge_sharing: &EdgeSharingMap) -> Vec<Vec<OrientedEdge>
...
                 }
 
+                let other_edge_reverted = other_edge.to_reverted();
+                if other_edge_reverted.chains_to(front) {
+                    current_chain.push_front(other_edge_reverted);
+                    border_edges.remove(i);
+                    found = true;
+
+                    break;
+                }
```
yanchith: This is a bit weird, because the reverted edge was not present in the original mesh, yet we output it in the result. Should this be the default behavior or should perhaps this function have a parameter determining whether we want this? Should we even be fixing it, and if so, should we be doing it here?

src/mesh/analysis.rs
```diff
@@ -242,14 +303,18 @@ pub fn border_edge_loops(edge_sharing: &EdgeSharingMap) -> Vec<Vec<OrientedEdge>
...
         };
 
-        // There can theoretically exist chains of edges that do not form border
-        // edge loops, e.g. if the mesh is non-manifold. We do not output those.
-        if loop_closed {
-            edge_loops.push(Vec::from(current_chain));
+        edge_loops.push(Vec::from(current_chain));
+
+        if !loop_closed {
+            non_deterministic = true;
```
yanchith: Not sure I agree with this change. If the loop wasn't closed, it is not a loop. Should we be including it? And if we do want to include it, isn't this another edge-case or another case of non-determinism that is distinct from the one we detect by counting vertex index occurrences? Shouldn't we tell that, for example with another result variant that returns a vector of closed loops and a vector of unclosed loops? Something like:

```rust
pub enum BorderEdgeLoopsResult {
    Found {
        detected_nondeterminism: bool,
        loops: Vec<Vec<OrientedEdge>>,
    }
    FoundWithIncompleteLoops {
        detected_nondeterminism: bool,
        loops_closed: Vec<Vec<OrientedEdge>>,
        loops_unclosed: Vev<Vec<OrientedEdge>>,
    }
    Watertight,
}
```

yanchith (commented):
Got some questions for you :)

src/mesh/analysis.rs
janper: the mesh may have millions of vertices, out of which last three are part of border edges. this then allocates a vector of million integers. I want to prevent that. the easiest thing is to find the first and the last border vertex index and only allocate such chunk of the list, but that is a very dull optimization

src/mesh/analysis.rs
janper: the oriented edge (halfedge) is an internal technicality of mesh math. for a user, there is only an edge. and so is an edge loop. so if an edge is refused to become a part of a visually obvious edge loop, then it seems like an error.

from the other perspective - what good is it to be so strict about the edge orientation?

src/mesh/analysis.rs
janper: is it even possible to get non-closed loops? what was the case with two-segment long edge loops? can we replicate it?

yanchith: I didn't really find out. The geometry was huge and the original bug could have theoretically come from elsewhere. We could checkout the original infinite looping function and find out, if we really want to. I didn't try reproducing on smaller data, but in theory, we could hand-craft such data (although it probably can not come from a mesh). Do you want to do it or should I? Either way, we should know what the assumptions are. If such data is impossible, we should be able to prove that. If it is possible, we need to handle the possibility.

Also. If the function always wants to assume the data it processes comes from a real mesh (and therefore certain invariants are upheld - e.g. that all edge loops must be at least three segments long, because they come from a face or a hole that used to be a face), we should adjust the API so that it is impossible to pass incorrect data here. One way to do that is to only allow creating `EdgeSharingMap`from a `Mesh`, and not from any iterator of oriented edges. Doing this will make it ok to assume/assert internally that all border edge loops will have at least 3 segments.

src/mesh/analysis.rs
yanchith: > what good is it to be so strict about the edge orientation?

IDK, I am strict about everything by defautt :joy: , but I don't know mesh math.

> if an edge is refused to become a part of a visually obvious edge loop, then it seems like an error.

Ok, understood. half-edges are only our internal thing. Maybe let's add this to the doc comment then? Explain why it internally operates on oriented edges, but detects loops as though the edges were not oriented?

src/mesh/analysis.rs
janper: you are right. what should I test then? maybe the number of segments in all border loops?

src/mesh/analysis.rs
janper: i can't think of a scenario where a vertex would be contained only in a single edge. the same applies to a border edge. i think we can assume if this number is 1, there must be some error in our code.

src/mesh/analysis.rs
janper: it's an internal thing of every mesh software. but true, i'll document it.

src/mesh/analysis.rs
janper: so, what do you prefer? i'd be also ok with panicking if the loop is not closed. it means the input data wasn't valid.

src/mesh/analysis.rs
yanchith: If the data comes from a mesh. You can currently hand-craft edge sharing maps like this. This is what I meant by the comment, that we should only be able to construct edge sharing from a mesh. Once we have that, We can start asserting things.

This can be a two-stage process:

1) We can start asserting all edge sharing data comes from a mesh and is not handcrafted AND add a fixme to make it so

2) Actually implement the fixme

src/mesh/analysis.rs
yanchith: Nah, let's not go that far. Just change the comparison sign to `>=` so that this doesn't suddenly start failing when a hash function changes.

src/mesh/analysis.rs
yanchith: If we truly believe this cannot happen to our mesh, then let's assert.

However, we can only panic if we are asserting certain invariants hold. In this case, the invariant that a border edge loop must be closed. This is true only if the data comes from a real mesh, but currently it doesn't have to, because the API allows you to hand-craft edge sharing maps.

We need to change the API and before we can safely assert (not necessarily in this order - see https://github.com/subdgtl/HURBAN-Selector/pull/171/files#r404086223)

src/mesh/analysis.rs
yanchith: Now that I'm thinking about it, I don't know if `edge_loops.len() >= 1` has much meaning. But for sure let's not assert anything concrete when the output is not deterministic and depends on the hash algorithm and envinronment.

janper: I've removed the test completely

src/mesh/analysis.rs
janper: I've removed this one too

src/mesh/analysis.rs
janper: This is a more general problem - we are almost nowhere sure the data comes from any valid source. Shouldn't we make a new task to change the API so that this is safe?

janper: let's assert. this shouldn't happen. if it does, we messed up somewhere else.

janper:
ready for another review

yanchith (dismissed):
Nice! I made some minor cleanup. Going to merge once CI passes.

yanchith:
Meh, don't merge yet. I forgot to push my minor fixes and now github has an outage. I will merge this as soon as I managed update the branch.

yanchith (approved):

PR #176, opened 2020-05-18
Implement voxel-based metaballs

janper:
hurban

Please review after #175 
Reference: http://www.geisswerks.com/ryan/BLOBS/blobs.html


src/mesh/voxel_cloud.rs
```diff
@@ -14,6 +14,27 @@ use crate::plane::Plane;
...
+#[derive(Debug)]
+pub enum TreatEmpty {
+    AsNone,
+    AsValue(f32),
+}
+
+/// Selects falloff function for the falloff field computation. The parameter
+/// specifies a distance multiplier.
+#[allow(dead_code)]
+#[derive(Debug)]
```
yanchith: `derive(Debug, Clone, Copy, PartialEq)`

src/mesh/voxel_cloud.rs
```diff
@@ -14,6 +14,27 @@ use crate::plane::Plane;
...
+pub enum TreatEmpty {
+    AsNone,
+    AsValue(f32),
+}
+
+/// Selects falloff function for the falloff field computation. The parameter
+/// specifies a distance multiplier.
+#[allow(dead_code)]
```
yanchith: Let's add this to the unused variants, not the whole enum.

src/mesh/voxel_cloud.rs
```diff
@@ -995,6 +1015,45 @@ impl ScalarField {
...
     }
 
+    /// Add values of the other scalar fields to the respective voxels of the
+    /// current scalar field. The current scalar field will not be resized, so
+    /// only the overlapping areas will be will be added. If the voxel in the
+    /// current or the other scalar field is None (does not contain any voxel or
+    /// does not exist), it is treated according to the selected
+    /// `empty_voxels_treatment` - it is either replaced with a constant value
+    /// or considered None and then also the resulting voxel will become None.
+    #[allow(dead_code)]
```
yanchith: Unnecessary `#[allow(dead_code)]`

src/mesh/voxel_cloud.rs
```diff
@@ -14,6 +14,27 @@ use crate::plane::Plane;
 
 use super::{primitive, tools, Face, Mesh, NormalStrategy};
 
+/// Describes how should be empty (None) voxel be treated. Either as None or as
+/// a constant value. This is used for arithmetic functions, such as addition.
+#[allow(dead_code)]
```
yanchith: Let's add this to the unused variants, not the whole enum.

src/mesh/voxel_cloud.rs
```diff
@@ -14,6 +14,27 @@ use crate::plane::Plane;
...
+#[derive(Debug)]
+pub enum TreatEmpty {
+    AsNone,
+    AsValue(f32),
+}
+
+/// Selects falloff function for the falloff field computation. The parameter
```
yanchith: "for the distance field computation"? I searched, but I see nothing named falloff field.

src/mesh/voxel_cloud.rs
```diff
@@ -14,6 +14,27 @@ use crate::plane::Plane;
 
 use super::{primitive, tools, Face, Mesh, NormalStrategy};
 
+/// Describes how should be empty (None) voxel be treated. Either as None or as
+/// a constant value. This is used for arithmetic functions, such as addition.
+#[allow(dead_code)]
+#[derive(Debug)]
+pub enum TreatEmpty {
```
yanchith: `derive(Debug, Clone, Copy, PartialEq)`

src/mesh/voxel_cloud.rs
```diff
@@ -995,6 +1015,45 @@ impl ScalarField {
         }
     }
 
+    /// Add values of the other scalar fields to the respective voxels of the
+    /// current scalar field. The current scalar field will not be resized, so
+    /// only the overlapping areas will be will be added. If the voxel in the
+    /// current or the other scalar field is None (does not contain any voxel or
+    /// does not exist), it is treated according to the selected
+    /// `empty_voxels_treatment` - it is either replaced with a constant value
+    /// or considered None and then also the resulting voxel will become None.
```
yanchith: This description should be on the `TreatEmpty` type and the documentation here should just say something like: "Empty voxels of both scalar fields are treated according to the selected `empty_voxels_treatment` parameter."

Also I think we will want different `TreatEmpty` for self and other in the future.

src/mesh/voxel_cloud.rs
```diff
@@ -995,6 +1015,45 @@ impl ScalarField {
...
+                &other.voxel_dimensions,
+            );
+            let voxel_other = other.value_at_absolute_voxel_coordinate(&absolute_coordinate_other);
+
+            match empty_voxels_treatment {
+                TreatEmpty::AsNone => {
+                    *voxel = if let (Some(value_self), Some(value_other)) = (&voxel, voxel_other) {
+                        Some(value_self + value_other)
+                    } else {
+                        None
```
yanchith: Whoah! I think this resets `voxel` to `None` even if by itself it is not `None`. Shouldn't this be `voxel`? Or even better, don't assign to it in the else branch.

src/mesh/voxel_cloud.rs
```diff
@@ -1200,13 +1262,98 @@ impl ScalarField {
...
+            let is_outside = discovered_as_outer_and_empty[one_dimensional];
+
+            // Process the current voxel.
+
+            self.voxels[one_dimensional] = match falloff_function {
+                FalloffFunction::Linear(mul) => {
+                    // distance values: 0 to infinity
+                    // suggested multiplier: 1
+                    // output values: 0 to infinity
+                    // value on volume: 0
```
yanchith: This information is great! But it would be even greater if it wasn't an internal comment inside a function, but an actual doc comment on `FalloffFunction::Linear`. Same goes for all other variants of `FalloffFunction`.

src/mesh/voxel_cloud.rs
```diff
@@ -14,6 +14,27 @@ use crate::plane::Plane;
...
+
+/// Selects falloff function for the falloff field computation. The parameter
+/// specifies a distance multiplier.
+#[allow(dead_code)]
+#[derive(Debug)]
+pub enum FalloffFunction {
+    Linear(f32),
+    Square(f32),
+    InverseLinear(f32),
+    InverseSquare(f32),
```
yanchith: I'd maybe call the inverse functions reciprocal. (reciprocal is multiplicative inverse, but there also is an additive inverse)

src/mesh/voxel_cloud.rs
```diff
@@ -1200,13 +1262,98 @@ impl ScalarField {
...
-            self.voxels[one_dimensional] = if discovered_as_outer_and_empty[one_dimensional] {
-                Some(distance)
-            } else {
-                Some(-distance)
+            let is_outside = discovered_as_outer_and_empty[one_dimensional];
+
+            // Process the current voxel.
+
+            self.voxels[one_dimensional] = match falloff_function {
+                FalloffFunction::Linear(mul) => {
```
yanchith: The following should be used sparingly, but I think this is a good occasion for it:

Instead of having the logic here (and potentially duplicating it soon in the future), we could have `FalloffFunction::apply(&self, distance: f32) -> f32` and just call it here. You can add the apply associated function in a regular impl block for `FalloffFunction`.

src/mesh/voxel_cloud.rs
```diff
@@ -1200,13 +1262,98 @@ impl ScalarField {
...
+                }
+                FalloffFunction::Square(mul) => {
+                    // distance values: 0 to infinity
+                    // suggested multiplier: 1
+                    // output values: 0 to infinity
+                    // value on volume: 0
+                    if distance == 0.0 {
+                        Some(0.0)
+                    } else {
+                        let value = distance * mul * distance * mul;
```
yanchith: Just a philosophical question, but should here and in `InverseSquare` be `mul` squared also?

src/interpreter_funcs/voxel_metaballs.rs
```diff
@@ -0,0 +1,303 @@
...
+                refinement: ParamRefinement::Float(FloatParamRefinement {
+                    default_value: Some(0.5),
+                    min_value: Some(0.0001),
+                    max_value: Some(1.0),
+                }),
+                optional: false,
+            },
+            ParamInfo {
+                name: "Volume range",
+                description: "Max value 2 = fill volumes",
```
yanchith: Are you sure about this comment?

src/interpreter_funcs/voxel_metaballs.rs
```diff
@@ -0,0 +1,303 @@
...
+            ParamInfo {
+                name: "Detailed Mesh Analysis",
+                description: "Reports detailed analytic information on the created mesh.\n\
+                              The analysis may be slow, therefore it is by default off.",
+                refinement: ParamRefinement::Boolean(BooleanParamRefinement {
+                    default_value: false,
+                }),
+                optional: false,
+            },
+        ]
```
yanchith: The analyses have been merged into one in the master branch meanwhile.

yanchith (changes requested):
Very nice! I have almost no concerns regarding functionality (besides one small bug I discovered in a dead branch of the code), but do have a couple of technical suggestions.

src/mesh/voxel_cloud.rs
janper: It is a convention https://www.dpmag.com/how-to/tip-of-the-week/how-light-works-the-inverse-square-law/

src/mesh/voxel_cloud.rs
janper: dunno. i think it's easier to imagine the distance multiplied and then something magic happens with the new value

src/interpreter_funcs/voxel_metaballs.rs
janper: this is incredibly tricky, not only regarding the comment. the values can be unexpected and actually i don't know how to set them properly. the issue roots in the falloff functions, for voxels inside volumes. actually, i don't know what should be the correct mathematical solution.
more in person

src/interpreter_funcs/voxel_metaballs.rs
janper: i have to update it here too. this was based on an older version of master.

src/mesh/voxel_cloud.rs
janper: maybe... but i can't imagine a situation when id' like to treat self and other differently. i prefer just leaving a FIXME note

src/mesh/voxel_cloud.rs
janper: okay, here i can imagine what i couldn't imagine in the previous comment. this behavior is similar to the interpolation function

src/mesh/voxel_cloud.rs
janper: so if the falloff functions happen to contain the logic, shouldn't this be their param?

src/mesh/voxel_cloud.rs
janper: good, i like it. what about doing the same for `interpolate_to` and `add_values` and potentially other arithmetic operations on scalar fields?

janper: side note: we need to reveal the scalar fields in the UI. they are becoming beautifully powerful!

src/mesh/voxel_cloud.rs
yanchith: Seems like independent parameters to me. Also I can't find where the falloff functions contain this logic. Regardless of whether the falloff functions are implemented in an impl block or inline, the value from the other voxel cloud is extracted beforehand.

src/mesh/voxel_cloud.rs
yanchith: Ok, I don't feel strongly either way. Both are conventions in different fields it seems.

src/mesh/voxel_cloud.rs
yanchith: > what about doing the same for interpolate_to and add_values and potentially other arithmetic operations on scalar fields?

`add_values` already has it (doesn't it?). We can add it to other scalar field methods, but since that isn't directly affecting metaballs, let's do it in a separate PR and for now just create an issue for that (or a fixme).

src/mesh/voxel_cloud.rs
yanchith: So which do you prefer? Because for polynomials the parameter usually isn't exponentiated, but here it is.

janper: I prefer the way it is

src/mesh/voxel_cloud.rs
janper: rubberducking:
```
treat_self = AsNone, treat_other = AsNone {
1+1 = 2
None + 1 = None
1 + None = None
}

treat_self = AsValue(1), treat_other = AsNone {
1+1 = 2
None + 1 = 2
1 + None = None
}

treat_self = AsNone, treat_other = AsValue(1) {
1+1 = 2
None + 1 = None
1 + None = 2
}

treat_self = AsValue(1), treat_other = AsValue(1) {
1+1 = 2
None + 1 = 2
1 + None = 2
}
```

janper:
My job is done here

janper:
no idea why does the formatter fail. it doesn't do anything on my local machine

janper:
Also the last small improvement awaits when this PR will have been merged. After that, I am ready for the major cleanup of the Voxel cloud (naming and function order) and then we are done with features.

yanchith:
> no idea why does the formatter fail. it doesn't do anything on my local machine

Yes, this started happening after the toolchain update. Just ignore the windows CI for now and I will fix it eventually.

yanchith:
@janper Note that on Windows CI fails because of the formatter issue, but on other platforms clippy complains about valid problems.

src/mesh/voxel_cloud.rs
```diff
@@ -1292,10 +1389,16 @@ impl ScalarField {
...
-    #[allow(dead_code)]
-    pub fn add_values(&mut self, other: &ScalarField, empty_voxels_treatment: TreatEmpty) {
+    /// `empty_voxels_treatment` for both, the current and the other scalar
+    /// field - it is either replaced with a constant value or considered None
+    /// and then also the resulting voxel will become None.
+    // FIXME: Join with interpolate_to
+    pub fn add_values(
+        &mut self,
+        other: &ScalarField,
+        empty_voxels_treatment_current: TreatEmpty,
```
yanchith: To have consistent parameter naming with other functions int this file, rename "current" "self".

src/mesh/voxel_cloud.rs
```diff
@@ -15,25 +15,121 @@ use super::{primitive, tools, Face, Mesh, NormalStrategy};
...
 }
 
-/// Selects falloff function for the falloff field computation. The parameter
+/// Selects falloff function for the distance field computation. The parameter
 /// specifies a distance multiplier.
-#[allow(dead_code)]
-#[derive(Debug)]
+#[derive(Debug, Clone, Copy, PartialEq)]
 pub enum FalloffFunction {
+    /// distance values: 0 to infinity \
```
yanchith, line 20: Since this is now a doc comment, these could use some formatting.

src/mesh/voxel_cloud.rs
```diff
@@ -15,25 +15,121 @@ use super::{primitive, tools, Face, Mesh, NormalStrategy};
...
+    /// output values: 0 to 1 \
+    /// value on volume: 1
+    #[allow(dead_code)]
     Perlin(f32),
 }
 
+impl FalloffFunction {
+    pub fn apply(&self, distance: f32, is_outside: bool) -> Option<f32> {
+        match self {
+            FalloffFunction::Linear(mul) => {
```
yanchith, line 55: Bad luck for you, but now I looked at these functions more closely.

- Linear and Square don't need to special case zero, the multiplication should handle that.
- InverseLinear and InverseSquare should return `None` for zero distances. Those functions are not defined for zero (see attached image) Actually, it is pretty weird that `apply` returns `Option` but every case actually returns `Some`. Maybe you wanted to do this and forgot? Alternatively we could go for infinity, but that is not very practical. Maybe `None` will actually work quite well, if we select a good default value for undefined parts of the parameter domain.. Let's discuss in person tomorrow? 


desmos-graph

src/mesh/voxel_cloud.rs
```diff
@@ -15,25 +15,121 @@ use super::{primitive, tools, Face, Mesh, NormalStrategy};
...
+            }
+            FalloffFunction::Square(mul) => {
+                if distance == 0.0 {
+                    Some(0.0)
+                } else {
+                    let value = distance * mul * distance * mul;
+                    if is_outside {
+                        Some(value)
+                    } else {
+                        Some(-value)
```
yanchith, line 77: This is not how squaring behaves. We can have Cube, if we want negative numbers.

src/mesh/voxel_cloud.rs
```diff
@@ -15,25 +15,121 @@ use super::{primitive, tools, Face, Mesh, NormalStrategy};
...
+            }
+            FalloffFunction::InverseLinear(mul) => {
+                if distance == 0.0 {
+                    Some(1.0)
+                } else {
+                    let value = 1.0 / (distance * mul);
+                    if is_outside {
+                        Some(value)
+                    } else {
+                        Some(1.0 + value)
```
yanchith, line 90: How did you get this value for InverseLinear and InverseSquare? Those functions look nothing like that. (See image attached above)

src/mesh/voxel_cloud.rs
```diff
@@ -15,25 +15,121 @@ use super::{primitive, tools, Face, Mesh, NormalStrategy};
...
+                        * distance
+                        * mul
+                        * distance
+                        * mul
+                        * (distance * mul * (distance * mul * 6.0 - 15.0) + 10.0);
+                    if is_outside {
+                        Some(1.0 / value)
+                    } else {
+                        Some(1.0 + value)
+                    }
```
yanchith: Could you explain the reasoning here for 1/value vs 1+value?

src/mesh/voxel_cloud.rs
```diff
@@ -1267,6 +1384,74 @@ impl ScalarField {
...
+
+            *voxel = match empty_voxels_treatment_current {
+                TreatEmpty::AsNone => {
+                    if let Some(value_self) = &voxel {
+                        match empty_voxels_treatment_other {
+                            TreatEmpty::AsNone => {
+                                if let Some(value_other) = voxel_other {
+                                    Some(value_self + value_other)
+                                } else {
+                                    None
```
yanchith: Is this how we want it to behave? `Some(value_self)` also looks like a legit alternative.

src/mesh/voxel_cloud.rs
```diff
@@ -1267,6 +1384,74 @@ impl ScalarField {
...
+                                }
+                            }
+                            TreatEmpty::AsValue(substitute_value_other) => {
+                                let value_other = voxel_other.unwrap_or(substitute_value_other);
+                                Some(value_self + value_other)
+                            }
+                        }
+                    } else {
+                        None
```
yanchith: Shouldn't this return `value_other`? (potentially `Some(value_other)` or `None`)

src/mesh/voxel_cloud.rs
```diff
@@ -1267,6 +1384,74 @@ impl ScalarField {
...
+                    }
+                }
+                TreatEmpty::AsValue(substitute_value_self) => {
+                    let value_self = voxel.unwrap_or(substitute_value_self);
+                    match empty_voxels_treatment_other {
+                        TreatEmpty::AsNone => {
+                            if let Some(value_other) = voxel_other {
+                                Some(value_self + value_other)
+                            } else {
+                                None
```
yanchith: Shouldn't this return `Some(value_self)`?

yanchith (changes requested):
Upon closer examination I discovered a few more issues. It's nothing terrible but they are design issues, not just technicalities. Let's talk about them in person tomorrow to speed things up.

src/mesh/voxel_cloud.rs
yanchith: Also, maybe instead of passing in `distance` and `is_outside`, we could collapse that to a single `x` that is not necessarily a positive number?

src/mesh/voxel_cloud.rs
yanchith: Just now noticed this. I would personally treat `None` as zero for `AsNone`, so:

None + None = 0
1 + None = 1
None + 1 = 1

If we were multiplying, we would treat `None` as one. I think it makes sense to preserve identity.

src/mesh/voxel_cloud.rs
janper: this is exactly what i mentioned before - https://github.com/subdgtl/HURBAN-Selector/pull/176#discussion_r427812431
there is a very difficult decision to be made and i couldn't find a better solution than this

src/mesh/voxel_cloud.rs
janper: this is a mistake. should be `1.0 / value` and `1.0 + 1.0 / value`

src/mesh/voxel_cloud.rs
janper: Not really, at least not when then treatment is set to `AsNone`. I can imagine adding something like `Ignore` here, that would do what you request

janper: I run into the same issue with Grasshopper today. With each new version of Grasshopper (which doesn't happen too often), there are some new versions of components. However, if you open a file with old components, Grasshopper uses the old versions of the components so that you get the same result as before. 

Now the old `Multiply component` returned `Null` for `Null * Number` and a `Vector` for `Vector * Number`. The new `Multiply component` returns a `Number` for  `Null * Number` (`Null * 1.0 = 1.0`). Both, old and new Grasshopper return a `Vector` for `Null + Vector` but fail for `Number + Vector`. So a chain `Null * Number + Vector` used to return `Vector` and now it fails. 

The moral of this story is: there is no convention ;)

src/mesh/voxel_cloud.rs
janper: like what? in my editor it formats nicely
image

src/mesh/voxel_cloud.rs
janper: yes, we need to discuss this in person. i know the values are not mathematically correct, but they are useful. however, i'm sure we can figure a more consistent behavior

src/mesh/voxel_cloud.rs
yanchith: Ugh, ok, let's leave it as is.

yanchith (approved):

PR #177, opened 2020-05-24
Update to wgpu 0.5 and rust toolchain 1.42.1, fix many things

yanchith:
This updates to wgpu 0.5 and fixes a few long-standing issues along the
way. The new wgpu version required a newer compiler, so I updated the toolchain
too. That caused rustfmt and clippy to complain in a few trivial cases so those
are now fixed as part of this patch.

Important fixes:

- It is now possible to create screenshots on Linux (fixed by updating wgpu).
- Resizing the window no longer crashes on Linux (fixed by updating wgpu).
- Shadows are now more predictable and use a tighter bounding box.
- Fixed an error in the correction matrix creation (the row/column order was
  incorrect) that caused some rendering inconsistencies for shadow
  rendering and was just plain incorrect.

Small improvements:

- Changes in wgpu allowed for removing callback code from screenshot drawing in
  lib.rs, simplifying the main loop slightly.
- Documentation in lib.rs is updated to reflect winit and wpgu behavior
  regarding presenting the swapchain image and OS redraw requests.
- Shadow casting light now doesn't need to be skewed, the renderer automatically
  computes the up vector such that it doesn't produce NaNs with the light vector
  in the view matrix.
- Fixed minor oversampling issue in the shadow map sampling code.

TODO:

- [x] Test on Windows
- [X] Test on Linux
- [x] Test on macOS


yanchith:
The rust toolchain is actually 1.43.1, not 1.42.1 :)

Also I am currently looking into the windows rustfmt fail. It only happens on CI, not my windows machine.

yanchith:
Ok, tested on all platforms now. The windows CI fails on formatting, but formatting passes on local windows machines. I'd not gate this PR on that and will fix that separately.

janper (approved):

PR #181, opened 2020-05-27
Change adapter selection to explicitly go over a list of backends

yanchith:
The default backend list for Windows now contains the D3D12 backend, but the
Vulkan backend still has priority, if present.


yanchith:
Checked on linux and macOS. @janper Could you please check windows? (especially your venerable dell XPS 13 :slightly_smiling_face: )

janper (approved):

PR #183, opened 2020-05-27
General operations for one and two scalar fields, explicit inner volumes

janper:
In this PR

- Mathematical operations on values of a single scalar field are now a general function with an enum param, which defines the specific function to perform
- Mathematical operations on values of two scalar fields are now a general function with an enum param, which defines the specific function to perform
- Detection of voxels inside watertight volumes is now explicit

**UI problem: for Metaballs, the volume value range is `infinity..=infinity` due to the nature of the inverse square falloff function. Even though this is correct, it causes a UI problem when setting the volume range params. Suggestions?**

_Note: This is the last feature PR from me. There will be one more which cleans up naming code of the scalar field related fields._

### DIFF
--- a/src/interpreter_funcs/voxel_boolean_intersection.rs
+++ b/src/interpreter_funcs/voxel_boolean_intersection.rs
@@ -11,7 +11,9 @@ use crate::interpreter::{
     BooleanParamRefinement, Float3ParamRefinement, Func, FuncError, FuncFlags, FuncInfo,
     LogMessage, ParamInfo, ParamRefinement, Ty, UintParamRefinement, Value,
 };
-use crate::mesh::voxel_cloud::{self, FalloffFunction, ScalarField};
+use crate::mesh::voxel_cloud::{
+    self, FalloffFunction, OneVoxelFunction, ScalarField, TwoVoxelFunction,
+};
 
 const VOXEL_COUNT_THRESHOLD: u32 = 100_000;
 
@@ -105,6 +107,7 @@ impl Func for FuncBooleanIntersection {
                 }),
                 optional: false,
             },
+            // FIXME: Consider changing to volume range for consistency.
             ParamInfo {
                 name: "Grow",
                 description: "The voxelization algorithm puts voxels on the surface of \
@@ -219,14 +222,28 @@ impl Func for FuncBooleanIntersection {
         let mut voxel_cloud1 = ScalarField::from_mesh(mesh1, &voxel_dimensions, 0.0, growth_u32);
         let mut voxel_cloud2 = ScalarField::from_mesh(mesh2, &voxel_dimensions, 0.0, growth_u32);
 
-        voxel_cloud1.compute_distance_field(&(0.0..=0.0), FalloffFunction::Linear(1.0));
-        voxel_cloud2.compute_distance_field(&(0.0..=0.0), FalloffFunction::Linear(1.0));
+        let volume_value_range = 0.0..=0.0;
 
-        let meshing_range = if fill {
-            (Bound::Unbounded, Bound::Included(growth_f32))
-        } else {
-            (Bound::Included(-growth_f32), Bound::Included(growth_f32))
-        };
+        voxel_cloud1.compute_distance_field(&volume_value_range, FalloffFunction::Linear(1.0));
+        voxel_cloud2.compute_distance_field(&volume_value_range, FalloffFunction::Linear(1.0));
+
+        if fill {
+            let mut voxel_cloud1_inside =
+                voxel_cloud1.extract_voxels_inside_volumes(&volume_value_range);
+
+            let mut voxel_cloud2_inside =
+                voxel_cloud2.extract_voxels_inside_volumes(&volume_value_range);
+
+            voxel_cloud1_inside.apply_one_voxel_function(OneVoxelFunction::SetConstant(0.0));
+            voxel_cloud2_inside.apply_one_voxel_function(OneVoxelFunction::SetConstant(0.0));
+
+            voxel_cloud1
+                .apply_two_voxel_function(&voxel_cloud1_inside, TwoVoxelFunction::ReplaceValues);
+            voxel_cloud2
+                .apply_two_voxel_function(&voxel_cloud2_inside, TwoVoxelFunction::ReplaceValues);
+        }
+
+        let meshing_range = (Bound::Included(-growth_f32), Bound::Included(growth_f32));
 
         voxel_cloud1.boolean_intersection(&meshing_range, &voxel_cloud2, &meshing_range);
 
--- a/src/interpreter_funcs/voxel_boolean_intersection.rs
+++ b/src/interpreter_funcs/voxel_boolean_intersection.rs
@@ -12,7 +12,7 @@ use crate::interpreter::{
     LogMessage, ParamInfo, ParamRefinement, Ty, UintParamRefinement, Value,
 };
 use crate::mesh::voxel_cloud::{
-    self, FalloffFunction, OneVoxelFunction, ScalarField, TwoVoxelFunction,
+    self, BinaryVoxelFunction, FalloffFunction, ScalarField, UnaryVoxelFunction,
 };
 
 const VOXEL_COUNT_THRESHOLD: u32 = 100_000;
@@ -224,8 +224,8 @@ impl Func for FuncBooleanIntersection {
 
         let volume_value_range = 0.0..=0.0;
 
-        voxel_cloud1.compute_distance_field(&volume_value_range, FalloffFunction::Linear(1.0));
-        voxel_cloud2.compute_distance_field(&volume_value_range, FalloffFunction::Linear(1.0));
+        voxel_cloud1.compute_distance_field(&volume_value_range, 1.0, FalloffFunction::Linear);
+        voxel_cloud2.compute_distance_field(&volume_value_range, 1.0, FalloffFunction::Linear);
 
         if fill {
             let mut voxel_cloud1_inside =
@@ -234,18 +234,26 @@ impl Func for FuncBooleanIntersection {
             let mut voxel_cloud2_inside =
                 voxel_cloud2.extract_voxels_inside_volumes(&volume_value_range);
 
-            voxel_cloud1_inside.apply_one_voxel_function(OneVoxelFunction::SetConstant(0.0));
-            voxel_cloud2_inside.apply_one_voxel_function(OneVoxelFunction::SetConstant(0.0));
+            voxel_cloud1_inside.apply_unary_voxel_function(UnaryVoxelFunction::SetConstant(0.0));
+            voxel_cloud2_inside.apply_unary_voxel_function(UnaryVoxelFunction::SetConstant(0.0));
 
-            voxel_cloud1
-                .apply_two_voxel_function(&voxel_cloud1_inside, TwoVoxelFunction::ReplaceValues);
-            voxel_cloud2
-                .apply_two_voxel_function(&voxel_cloud2_inside, TwoVoxelFunction::ReplaceValues);
+            voxel_cloud1.apply_binary_voxel_function(
+                &voxel_cloud1_inside,
+                BinaryVoxelFunction::ReplaceIfValue,
+            );
+            voxel_cloud2.apply_binary_voxel_function(
+                &voxel_cloud2_inside,
+                BinaryVoxelFunction::ReplaceIfValue,
+            );
         }
 
         let meshing_range = (Bound::Included(-growth_f32), Bound::Included(growth_f32));
 
-        voxel_cloud1.boolean_intersection(&meshing_range, &voxel_cloud2, &meshing_range);
+        voxel_cloud1.apply_binary_voxel_function(
+            &voxel_cloud2,
+            BinaryVoxelFunction::BooleanAnd(meshing_range, meshing_range),
+        );
+        // voxel_cloud1.boolean_intersection(&meshing_range, &voxel_cloud2, &meshing_range);
 
         if !voxel_cloud1.contains_voxels_within_range(&meshing_range) {
             let error = FuncError::new(FuncBooleanIntersectionError::EmptyScalarField);
--- a/src/interpreter_funcs/voxel_boolean_union.rs
+++ b/src/interpreter_funcs/voxel_boolean_union.rs
@@ -12,7 +12,9 @@ use crate::interpreter::{
     BooleanParamRefinement, Float3ParamRefinement, Func, FuncError, FuncFlags, FuncInfo,
     LogMessage, ParamInfo, ParamRefinement, Ty, UintParamRefinement, Value,
 };
-use crate::mesh::voxel_cloud::{self, FalloffFunction, ScalarField};
+use crate::mesh::voxel_cloud::{
+    self, FalloffFunction, OneVoxelFunction, ScalarField, TwoVoxelFunction,
+};
 
 const VOXEL_COUNT_THRESHOLD: u32 = 100_000;
 
@@ -109,6 +111,7 @@ impl Func for FuncBooleanUnion {
                 }),
                 optional: false,
             },
+            // FIXME: Consider changing to volume range for consistency.
             ParamInfo {
                 name: "Grow",
                 description: "The voxelization algorithm puts voxels on the surface of \
@@ -221,14 +224,28 @@ impl Func for FuncBooleanUnion {
         let mut voxel_cloud1 = ScalarField::from_mesh(mesh1, &voxel_dimensions, 0.0, growth_u32);
         let mut voxel_cloud2 = ScalarField::from_mesh(mesh2, &voxel_dimensions, 0.0, growth_u32);
 
-        voxel_cloud1.compute_distance_field(&(0.0..=0.0), FalloffFunction::Linear(1.0));
-        voxel_cloud2.compute_distance_field(&(0.0..=0.0), FalloffFunction::Linear(1.0));
+        let volume_value_range = 0.0..=0.0;
 
-        let meshing_range = if fill {
-            (Bound::Unbounded, Bound::Included(growth_f32))
-        } else {
-            (Bound::Included(-growth_f32), Bound::Included(growth_f32))
-        };
+        voxel_cloud1.compute_distance_field(&volume_value_range, FalloffFunction::Linear(1.0));
+        voxel_cloud2.compute_distance_field(&volume_value_range, FalloffFunction::Linear(1.0));
+
+        if fill {
+            let mut voxel_cloud1_inside =
+                voxel_cloud1.extract_voxels_inside_volumes(&volume_value_range);
+
+            let mut voxel_cloud2_inside =
+                voxel_cloud2.extract_voxels_inside_volumes(&volume_value_range);
+
+            voxel_cloud1_inside.apply_one_voxel_function(OneVoxelFunction::SetConstant(0.0));
+            voxel_cloud2_inside.apply_one_voxel_function(OneVoxelFunction::SetConstant(0.0));
+
+            voxel_cloud1
+                .apply_two_voxel_function(&voxel_cloud1_inside, TwoVoxelFunction::ReplaceValues);
+            voxel_cloud2
+                .apply_two_voxel_function(&voxel_cloud2_inside, TwoVoxelFunction::ReplaceValues);
+        }
+
+        let meshing_range = (Bound::Included(-growth_f32), Bound::Included(growth_f32));
 
         voxel_cloud1.boolean_union(&meshing_range, &voxel_cloud2, &meshing_range);
 
--- a/src/interpreter_funcs/voxel_boolean_union.rs
+++ b/src/interpreter_funcs/voxel_boolean_union.rs
@@ -13,7 +13,7 @@ use crate::interpreter::{
     LogMessage, ParamInfo, ParamRefinement, Ty, UintParamRefinement, Value,
 };
 use crate::mesh::voxel_cloud::{
-    self, FalloffFunction, OneVoxelFunction, ScalarField, TwoVoxelFunction,
+    self, BinaryVoxelFunction, FalloffFunction, ScalarField, UnaryVoxelFunction,
 };
 
 const VOXEL_COUNT_THRESHOLD: u32 = 100_000;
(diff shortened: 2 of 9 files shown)


src/mesh/voxel_cloud.rs
```diff
@@ -30,93 +30,158 @@ pub enum FalloffFunction {
...
-                    Some(0.0)
-                } else {
-                    let value = d * d;
-                    if is_outside {
-                        Some(value)
-                    } else {
-                        Some(-value)
-                    }
-                }
+                Some(d * d)
```
yanchith: We talked in person about not applying the function to the parameter (`mul`), but it still is applied. Do we want to change that? If we do, it changes implementations of almost all of these fallof functions

src/mesh/voxel_cloud.rs
```diff
@@ -30,93 +30,158 @@ pub enum FalloffFunction {
...
-                let d = distance * mul;
-                if d == 0.0 {
-                    Some(1.0)
+        }
+    }
+}
+
+/// A function with one voxel as an input to be applied with `apply_one_voxel_function`.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum OneVoxelFunction {
```
yanchith: Naming nit: these things are sometimes called unary and binary. To me it sounds more precise than one-voxel and two-voxel.

src/mesh/voxel_cloud.rs
```diff
@@ -30,93 +30,158 @@ pub enum FalloffFunction {
...
+    #[allow(dead_code)]
+    AddConstant(f32),
+    /// Multiply any voxel value that is currently not None with the specified
+    /// constant.
+    #[allow(dead_code)]
+    MultiplyWithConstant(f32),
+    /// Raise any voxel value that is currently not None to the specified
+    /// power.
+    #[allow(dead_code)]
+    PowF(f32),
```
yanchith, line 68: Just pow?

src/mesh/voxel_cloud.rs
```diff
@@ -30,93 +30,158 @@ pub enum FalloffFunction {
...
+    /// the result is None.
+    ///
+    /// `interpolation_factor` is a value between 0.0 and 1.0 defining the ratio
+    /// between the current value to the other value. For factor outside this
+    /// range, the values will be extrapolated.
+    LinearInterpolation(f32),
+    /// Used to add values of two voxels. If any voxel's value is None, the
+    /// result is an identity of the other voxel. If both voxels are None, the
+    /// result is also None.
+    Addition,
```
yanchith: Naming between OneVoxelFunction and TwoVoxelFunction variants are not very consistent. E.g. OneVoxelFunction has AddConstant, but TwoVoxelFunction has Addition instead of Add.

src/mesh/voxel_cloud.rs
```diff
@@ -1494,14 +1518,90 @@ impl ScalarField {
...
+
+            // Process the current voxel.
+            self.voxels[one_dimensional] = falloff_function.apply(distance);
+        }
+    }
+
+    /// Clone the current scalar field and remove all outer and volume voxels,
+    /// so that only the voxels inside the closed volumes remain. Their original
+    /// values remain intact.
+    pub fn extract_voxels_inside_volumes<U>(&self, volume_value_range: &U) -> ScalarField
```
yanchith, line 1450: Without reading the code here, can't this be emulated with the more general functionality this PR provides, and a bit of functionality it doesn't yet provide, but we talked about?

That is, first clone and then a call to something like `fn flood_fill_from_borders(fill_value: Option<f32>, stop_value: Option<f32>)`? It is possible that `stop_value` might need to be a range (but that's difficult with Option) or a list of values, or alternatively, we can hack around it and have both `stop_range`and `stop_if_none: bool`. OR we can get fancy a have it be a predicate (function taking a voxel and producing bool), similarly to what you did elsewhere :)

yanchith (commented):
Overall I like the direction this is taking :heart: 

This is not a proper review yet. I have a few high level comments and questions. I just skimmed through the code for now, because I expect changes to happen. I will review more in-depth once the shape has stabilized more.

Two side notes for future consideration, when we will want to expose scalar fields to the UI:

- Couldn't we re-implement the existing boolean operations with TwoVoxelFunctions and resizing, especially since you wanted to remove auto-shrinking? Actually the only piece of functionality that cannot easily be re-implemented on top these new primitives is the distance field computation, because it is a graph algorithm.

- Also, as I now see the added functions, I am not too happy about the using Option in the storage, and would consider prototyping a zero-initialized (or filled with a constant value) alternative in the future. I believe all of current functionality could be preserved by adding a few more functions, e.g. SetConstantIfGreater(value, comparison_value), etc.  Option could still be used to indicate sampling outside the bounding box.

janper:
> Couldn't we re-implement the existing boolean operations with TwoVoxelFunctions and resizing, especially since you wanted to remove auto-shrinking?

perhaps yes, i'll try

janper:
> Also, as I now see the added functions, I am not too happy about the using Option in the storage, and would consider prototyping a zero-initialized (or filled with a constant value) alternative in the future. Option could still be used to indicate sampling outside the bounding box.

I'd be very cautious about this. How do you pick the _empty_ value? What if you want to read the scalar field in various ways without changing its content?

Hmmm... Maybe it wouldn't be a problem, but I've got a feeling it's not the right thing to do.

janper:
> I believe all of current functionality could be preserved by adding a few more functions, e.g. SetConstantIfGreater(value, comparison_value), etc.

What about a callback function?

src/mesh/voxel_cloud.rs
janper: ~Is that what we were talking about? I understood it this way - only changing the style so that the math functions look cleaner? If we went your way, we can remove the multiplier whatsoever because we already have a function that multiplies all the values.~

Do you actually mean applying the multiplier in the caller of these functions?

src/mesh/voxel_cloud.rs

src/mesh/voxel_cloud.rs
janper: `1/n`

src/mesh/voxel_cloud.rs
janper: Which one do you prefer?

src/mesh/voxel_cloud.rs
janper: why though?

janper: also this not only removes the outer areas but also the volumes, so there is no numerical boundary stop value per se.

src/mesh/voxel_cloud.rs
yanchith: No, I just meant applying it only once, instead of it being squared, etc. That was the only thing I was really talking about. So e.g. for square: instead of `mul * distance * mul * distance`, there would be just `mul * distance * distance`. Analogically for others.

src/mesh/voxel_cloud.rs
yanchith: What about it? I was only talking about the naming, not that the exponent must be an integer. In maths, It's just called pow. In programming languages it is called powf, if the exponent can be a real number.

src/mesh/voxel_cloud.rs
yanchith: Add/Subtract/Multiply/Divide, etc.

yanchith:
> perhaps yes, i'll try

Note that I only mentioned this as something to do in the future. This will be quite a bit of work, so we don't have to do it right now, in this PR

> Hmmm... Maybe it wouldn't be a problem, but I've got a feeling it's not the right thing to do.

Yeah, I'd just like to see the comparison of the two approaches, but again, that is something to maybe try in the future, not now.

> What about a callback function?

I'd avoid it if possible. There will be no way to provide it from the UI, so we might as well just enumerate every possible option explicitly instead of using callbacks.

src/mesh/voxel_cloud.rs
janper: nah, that's not good. we need to multiply the input. 
i've already implemented it, you'll see with the next push

src/mesh/voxel_cloud.rs
yanchith: You may be right! I'll have a look again to make sure I am not misleading you here.

janper:
ready for another review

src/mesh/voxel_cloud.rs
janper: forgotten about this one. will do in the next batch of changes

src/mesh/voxel_cloud.rs
```diff
@@ -18,103 +18,263 @@ use super::{primitive, tools, Face, Mesh, NormalStrategy};
...
+    /// `(Bound<f32>, Bound<f32>)` for the second voxel. If the other voxel is
+    /// True, None is returned. Otherwise the first voxel (even if None) is
+    /// returned.
+    // FIXME: Consider making a wrapper function for symmetry
+    BooleanDifference((Bound<f32>, Bound<f32>)),
+}
+
+impl BinaryVoxelFunction {
+    // FIXME: make tests
+    /// # Panics
```
yanchith, line 157: Add newline after title

src/mesh/voxel_cloud.rs
```diff
@@ -18,103 +18,263 @@ use super::{primitive, tools, Face, Mesh, NormalStrategy};
...
-                    let value = 1.0 / (d);
-                    if is_outside {
-                        Some(value)
+            BinaryVoxelFunction::Add => match (voxel_self, voxel_other) {
+                (None, None) => None,
+                _ => Some(voxel_self.unwrap_or(0.0) + voxel_other.unwrap_or(0.0)),
+            },
+            BinaryVoxelFunction::Subtract => match (voxel_self, voxel_other) {
+                (None, None) => None,
+                (None, Some(_)) => voxel_other,
```
yanchith, line 183: This is kinda an edge case, but shouldn't `Subtract` use additive identity for the voxel that is `None`? This case shouldn't be necessary at all, and the case below can use `unwrap_or(0.0)` for both operands, so for `(None, Some(value))` you would get `-value`.

src/mesh/voxel_cloud.rs
```diff
@@ -18,103 +18,263 @@ use super::{primitive, tools, Face, Mesh, NormalStrategy};
...
+impl BinaryVoxelFunction {
+    // FIXME: make tests
+    /// # Panics
+    /// Panics when interpolating infinite values
+    pub fn apply(self, voxel_self: Option<f32>, voxel_other: Option<f32>) -> Option<f32> {
+        match self {
+            BinaryVoxelFunction::InterpolateLinear(interpolation_factor) => {
+                match (voxel_self, voxel_other) {
+                    (Some(value_self), Some(value_other)) => {
+                        let self_to_other_range = (value_self as f64)..(value_other as f64);
```
yanchith, line 164: f64::from

src/mesh/voxel_cloud.rs
```diff
@@ -18,103 +18,263 @@ use super::{primitive, tools, Face, Mesh, NormalStrategy};
...
+                        let value = math::remap(
+                            interpolation_factor as f64,
```
yanchith, line 166: f64::from

src/mesh/voxel_cloud.rs
```diff
@@ -18,103 +18,263 @@ use super::{primitive, tools, Face, Mesh, NormalStrategy};
...
+                (None, Some(_)) => voxel_other,
+                _ => Some(voxel_self.expect("Self voxel is None") - voxel_other.unwrap_or(0.0)),
+            },
+            BinaryVoxelFunction::Multiply => match (voxel_self, voxel_other) {
+                (None, None) => None,
+                _ => Some(voxel_self.unwrap_or(1.0) * voxel_other.unwrap_or(1.0)),
+            },
+            BinaryVoxelFunction::Divide => match (voxel_self, voxel_other) {
+                (None, None) => None,
+                (None, Some(_)) => voxel_other,
```
yanchith, line 192: This is kinda an edge case, but shouldn't `Divide` use multiplicative identity for the voxel that is `None`? This case shouldn't be necessary at all, and the case below can use `unwrap_or(1.0)` for both operands, so for `(None, Some(value))` you would get `1.0 / value`.

src/mesh/voxel_cloud.rs
```diff
@@ -18,103 +18,263 @@ use super::{primitive, tools, Face, Mesh, NormalStrategy};
...
+                _ => Some(voxel_self.expect("Self voxel is None") / voxel_other.unwrap_or(1.0)),
+            },
+            BinaryVoxelFunction::ReplaceIfValue => match (voxel_self, voxel_other) {
+                (_, Some(_)) => voxel_other,
+                _ => voxel_self,
+            },
+            BinaryVoxelFunction::BooleanAnd(
+                (volume_value_range_self_start, volume_value_range_self_end),
```
yanchith, line 200: Instead of destructuring completely here and rebuilding a few lines below, you can just destructure to `volume_value_range_self` directly. Same for volume_value_range_other`.

src/mesh/voxel_cloud.rs
```diff
@@ -18,103 +18,263 @@ use super::{primitive, tools, Face, Mesh, NormalStrategy};
...
-                let d = distance * mul;
-                if d == 0.0 {
-                    Some(1.0)
-                } else {
-                    let value = 1.0 / (d * d);
-                    if is_outside {
-                        Some(value)
+            BinaryVoxelFunction::BooleanOr(
+                (volume_value_range_self_start, volume_value_range_self_end),
+                (volume_value_range_other_start, volume_value_range_other_end),
```
yanchith, line 221: Same as above, you don't need to destructure to the individual extremes, just destructure to the range.

src/mesh/voxel_cloud.rs
```diff
@@ -18,103 +18,263 @@ use super::{primitive, tools, Face, Mesh, NormalStrategy};
...
-                    Some(1.0)
-                } else {
-                    let value = d * d * d * (d * (d * 6.0 - 15.0) + 10.0);
-                    if is_outside {
-                        Some(1.0 / value)
-                    } else {
-                        Some(1.0 + 1.0 / value)
+            BinaryVoxelFunction::BooleanDifference((
+                volume_value_range_other_start,
+                volume_value_range_other_end,
```
yanchith, line 265: Same as above, just destructure directly to the range.

src/mesh/voxel_cloud.rs
```diff
@@ -18,103 +18,263 @@ use super::{primitive, tools, Face, Mesh, NormalStrategy};
...
-                        Some(value)
+            BinaryVoxelFunction::BooleanOr(
+                (volume_value_range_self_start, volume_value_range_self_end),
+                (volume_value_range_other_start, volume_value_range_other_end),
+            ) => {
+                let volume_value_range_self =
+                    (volume_value_range_self_start, volume_value_range_self_end);
+                let volume_value_range_other =
+                    (volume_value_range_other_start, volume_value_range_other_end);
+                if let (_, Some(value_other)) = (voxel_self, voxel_other) {
```
yanchith, line 227: Not sure why `voxel_self` is in this tuple if it is just discarded in pattern matching. Shouldn't it be `(None, Some(value_other))`? Would also make more sense to only interpolate values that were not present in self.

src/mesh/voxel_cloud.rs
```diff
@@ -1095,296 +1258,24 @@ impl ScalarField {
...
-                        }
-                    }
-                }
-            }
-        } else {
-            // Wipe the current scalar field if none of the scalar fields
-            // contained any volume voxels.
-            self.wipe();
+    /// Use the current scalar field's voxels as parameters into the selected
+    /// one voxel function and replace the current scalar field's voxel with the
```
yanchith, line 1262: one -> unary

src/mesh/voxel_cloud.rs
```diff
@@ -1095,296 +1258,24 @@ impl ScalarField {
...
     }
 
-    /// Computes boolean difference of the current scalar field minus the other
-    /// scalar field. The current scalar field will be modified so that voxels,
-    /// that are within volume value range in both scalar fields will be set to
-    /// None in the current scalar field, while the rest remains intact. The two
-    /// scalar fields do not have to contain voxels of the same size.
-    pub fn boolean_difference<U>(
+    /// Use the respective current and other scalar fields' voxels as parameters
+    /// into the selected two voxel function and replace the current scalar
```
yanchith, line 1271: two -> binary

src/mesh/voxel_cloud.rs
```diff
@@ -1494,14 +1383,90 @@ impl ScalarField {
...
-        // Contains indices into the voxel map
-        let mut queue_to_find_outer: VecDeque<usize> = VecDeque::new();
         // Contains indices to the voxel map and their distance value
         let mut queue_to_compute_distance: VecDeque<(usize, f32)> = VecDeque::new();
         // Match the voxel map length
-        let mut discovered_as_outer_and_empty = vec![false; self.voxels.len()];
         let mut discovered_for_distance_field = vec![false; self.voxels.len()];
 
+        // Scan for void voxels at the boundaries of the scalar field
+        // and at the same time for volume voxels anywhere.
```
yanchith, line 1392: Looks like outdated comment. This only scans for volume voxels now.

src/mesh/voxel_cloud.rs
```diff
@@ -1494,14 +1383,90 @@ impl ScalarField {
...
+            // Check each neighbor
+            for neighbor_offset in &neighbor_offsets {
+                let neighbor_absolute_coordinate = absolute_coordinate + neighbor_offset;
+                // If the neighbor does exist
+                if let Some(one_dimensional_neighbor) = absolute_voxel_to_one_dimensional_coordinate(
+                    &neighbor_absolute_coordinate,
+                    &self.block_start,
+                    &self.block_dimensions,
+                ) {
+                    // and hasn't been discovered yet and is void,
```
yanchith, line 1425: Probably outdated comment. This cares about void voxels OR voxels outside the value range.

src/mesh/voxel_cloud.rs
```diff
@@ -1494,14 +1383,90 @@ impl ScalarField {
...
+                    if !discovered_for_distance_field[one_dimensional_neighbor]
+                        && self
+                            .value_at_absolute_voxel_coordinate(&neighbor_absolute_coordinate)
+                            .map(|value| !volume_value_range.contains(&value))
+                            .unwrap_or(true)
```
yanchith, line 1430: Can't the second part of the condition be an assert? We already discovered all nonvoid voxels inside the volume range in the first scan.

src/mesh/voxel_cloud.rs
```diff
@@ -3220,108 +3156,6 @@ mod tests {
         assert!(analysis::are_similar(&voxel_mesh, &voxel_mesh_synced));
     }
 
-    #[test]
-    fn test_scalar_field_boolean_intersection_all_volume() {
```
yanchith, line 3224: Would be great if we could bring these tests back. Of course the correct scalar fields will look a bit differently, because the boolean functions don't resize anything.

src/interpreter_funcs/voxel_boolean_intersection.rs
```diff
@@ -219,16 +222,38 @@ impl Func for FuncBooleanIntersection {
...
+        }
+
+        let meshing_range = (Bound::Included(-growth_f32), Bound::Included(growth_f32));
 
-        voxel_cloud1.boolean_intersection(&meshing_range, &voxel_cloud2, &meshing_range);
+        voxel_cloud1.apply_binary_voxel_function(
+            &voxel_cloud2,
+            BinaryVoxelFunction::BooleanAnd(meshing_range, meshing_range),
+        );
+        // voxel_cloud1.boolean_intersection(&meshing_range, &voxel_cloud2, &meshing_range);
```
yanchith, line 256: Forgotten commented code

src/interpreter_funcs/voxelize.rs
```diff
@@ -203,7 +206,19 @@ impl Func for FuncVoxelize {
...
+            let mut scalar_field_inside =
+                scalar_field.extract_voxels_inside_volumes(&volume_value_range);
+            scalar_field_inside.apply_unary_voxel_function(UnaryVoxelFunction::SetConstant(0.0));
+            scalar_field.apply_binary_voxel_function(
+                &scalar_field_inside,
+                BinaryVoxelFunction::ReplaceIfValue,
+            );
+        }
 
         let meshing_range = if fill {
```
yanchith, line 223: This `if` shouldn't be necessary anymore as far as I understand and the meshing range can just be `(Bound::Included(-growth_f32), Bound::Included(growth_f32))`

src/interpreter_funcs/voxel_boolean_union.rs
```diff
@@ -218,30 +221,77 @@ impl Func for FuncBooleanUnion {
...
+
+        // FIXME: Visually it seems like this is chopping off some parts of the
+        // volume. This is another place where the chopping may occur.
+        if let Some(bounding_box) = BoundingBox::union(
+            [bounding_box_voxel_cloud1, bounding_box_voxel_cloud2]
+                .iter()
+                .copied(),
+        ) {
+            voxel_cloud1.resize_to_bounding_box_cartesian_space(&bounding_box);
+            voxel_cloud2.resize_to_bounding_box_cartesian_space(&bounding_box);
```
yanchith, line 277: If I understand correctly, we don't need to resize the second scalar fields. Values outside will just be void.

src/interpreter_funcs/voxel_boolean_union.rs
```diff
@@ -218,30 +221,77 @@ impl Func for FuncBooleanUnion {
             return Err(error);
         }
 
-        let mut voxel_cloud1 = ScalarField::from_mesh(mesh1, &voxel_dimensions, 0.0, growth_u32);
-        let mut voxel_cloud2 = ScalarField::from_mesh(mesh2, &voxel_dimensions, 0.0, growth_u32);
+        // FIXME: The growth + 1 here is only because it seems like some
+        // chopping happens. This is one place where the chopping may occur.
+        let mut voxel_cloud1 =
```
yanchith, line 226: I think the chopping you mention can be caused by using cartesian bounding boxes instead of voxel space bounding boxes. Try if it helps and maybe we don't have to hack around here.

src/interpreter_funcs/voxel_metaballs.rs
```diff
@@ -142,10 +144,10 @@ impl Func for FuncVoxelMetaballs {
...
                 description: "Materializes voxels with value within the given range.",
                 refinement: ParamRefinement::Float2(Float2ParamRefinement {
-                    min_value: Some(0.0),
-                    max_value: Some(100.0),
-                    default_value_x: Some(0.5),
-                    default_value_y: Some(2.0),
+                    min_value: None,
+                    max_value: None,
+                    default_value_x: Some(f32::INFINITY),
+                    default_value_y: Some(f32::INFINITY),
```
yanchith, line 150: This break the UI. If you try to change this, the value becomes NaN and can't be set back to infinity. Would make sense if this parameter wasn't exposed?

yanchith (changes requested):
First of all, the changes are great and the result seems much simpler. I do have a couple of technical suggestions and other misc things I noticed.

Also, for some reason this PR breaks voxel interpolated union. I tested on the simplest case: sphere+box on the same position, 0.1 voxel dimesnions, both fill and not. For every t except 0.0 and 1.0 it behaves like boolean difference. Could be due to volume value ranges (which would enjoy some unification with the meshing ranges across the voxel functions).

PR #184, opened 2020-06-24
Downgrade winit to v0.22.0

yanchith:
This is another meh interaction of winit and the sytem file dialog we encountered. I reported it in the winit repo https://github.com/rust-windowing/winit/issues/1611

The current newest version of winit (v0.22.2) has a problem if something (in
our case the file dialog) lets the Windows VM_PAINT messages accumulate in the
queue. The following assert fails in
winit/src/platform_impl/windows/event_loop.rs:1942

```rust
assert!(flush_paint_messages(
    None,
    &subclass_input.event_loop_runner
));
```
... presumably because while the file dialog was open, more VM_PAINT messages
were accumulated.

Since this is a platform layer change, I'd like to test this a bit on each platform before merging.

Tested on:

- [x] Windows
- [x] Linux (X11)
- [x] macOS

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,8 +27,8 @@ dependencies = [
 ]
 
 [[package]]
-name = "android_log-sys"
-version = "0.1.2"
+name = "android_glue"
+version = "0.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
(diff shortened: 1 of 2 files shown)


janper (approved):
Tested on Windows

